PR #3, opened 2016-03-02
Redux

skellock:
Here's Redux.  The example context is login/logout.  We'll add some more redux-based goodies soon!

### DIFF
--- a/App/Config/DebugSettings.js
+++ b/App/Config/DebugSettings.js
@@ -1,7 +1,8 @@
 const SETTINGS = {
   useFixtures: false,
   ezLogin: false,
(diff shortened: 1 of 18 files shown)


jamonholmgren:
:+1:

App/Config/DebugSettings.js
```diff
@@ -1,7 +1,8 @@
 const SETTINGS = {
   useFixtures: false,
   ezLogin: false,
-  yellowBox: __DEV__
+  yellowBox: __DEV__,
+  reduxLogging: true
```
GantMan: nice!

PR #4, opened 2016-03-06
Introducing Sagas for controlling async flow.

skellock:
Here are a few sagas to get us started.


jamonholmgren:
Update README? :)

PR #10, opened 2016-03-13
keyboard avoiding with fancy animation

GantMan:
### Pull-Request Checklist

- [x] Tests Pass

login_animation

### DIFF
--- a/App/Containers/LoginScreen.js
+++ b/App/Containers/LoginScreen.js
@@ -4,11 +4,15 @@ import React, {
   View,
   Text,
   TextInput,
-  TouchableOpacity
+  TouchableOpacity,
+  Image,
+  DeviceEventEmitter,
+  LayoutAnimation
 } from 'react-native'
 import { connect } from 'react-redux'
 import Styles from '../Styles/LoginScreenStyle'
 import Actions from '../Actions/Creators'
+import {Images, Metrics} from '../Themes'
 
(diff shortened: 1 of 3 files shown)


GantMan:
@skellock plz review - Added keyboard avoiding code and fancy animation :+1:

skellock:
Nice!

So I used https://github.com/Andr3wHur5t/react-native-keyboard-spacer recently and I thought the interface was gloriously simple.  Your implementation here actually does more than the other component, so maybe it can't be componentized. 

3 small nits:

1.  We should unsubscribe to those listeners when we unmount. 
2.  Curious, is there a willShow and willHide we could hook so the animations are a little tighter?
3.  I've never used LayoutAnimation before.  Neat!  Is it android friendly?

Nice touch with the shrinking logo.

GantMan:
1 - Unsubscribe!  Of course!  My bad, I'm adding it now.
2 - willShow and willHide are wayyyy tighter looking, but they do not work on Android (not sure why).  For simplicity of the code base (and cross platform) I'm just leaving it on what works.
3 - LayoutAnimation is amazing [even customizable](https://medium.com/@Jpoliachik/react-native-s-layoutanimation-is-awesome-4a4d317afd3e#.jx3u7hf3j), but *NO* it does not work on Android.   It doesn't cause a crash or anything, so depending on your definition of Android friendly, it might be.

PR #12, opened 2016-03-18
Adds an HttpClient for making JSON API calls.

skellock:
image

This is good start on the HTTP client.

I'd like to have multiple "api styles" perhaps which could target servers dishing up:  

* json
* formdata
* html

Let's chat more about this.

Also, the kitty is growing!!!

image


App/Reducers/WeatherReducer.js
```diff
@@ -0,0 +1,43 @@
...
+export const INITIAL_STATE = Immutable({
+  temperature: null,
+  fetching: null,
+  error: null,
+  city: null
+})
+
+// request temp
+const request = (state, action) =>
+  state.merge({
```
GantMan: I'll be honest, I'm not in love with seamless-immutable `.merge` because it gives the connotation that a merge is happening, and not a brand new object.   I just have to say this complaint to assuage my cognitive dissonance.

App/Sagas/WeatherSaga.js
```diff
@@ -0,0 +1,26 @@
...
+import R from 'ramda'
+import Types from '../Actions/Types'
+import Actions from '../Actions/Creators'
+
+export function * getWeather (city) {
+  const client = Client({ baseUrl: 'http://openweathermap.org/data/2.1' })
+  const response = yield call(client.get, 'find/name', { q: city })
+  const { ok, json } = response
+  if (ok) {
+    const kelvin = R.path(['list', 0, 'main', 'temp'], json)
```
GantMan: RAMDAAAAAAA!!!!!

PR #15, opened 2016-03-25
Adds example of setting a nav bar button.

markrickert:
Tests?

no

### DIFF
--- a/App/Containers/LoginScreen.js
+++ b/App/Containers/LoginScreen.js
@@ -45,6 +45,14 @@ class LoginScreen extends Component {
     // TODO: Revisit this if Android begins to support - https://github.com/facebook/react-native/issues/3468
     DeviceEventEmitter.addListener('keyboardDidShow', this.keyboardDidShow.bind(this))
     DeviceEventEmitter.addListener('keyboardDidHide', this.keyboardDidHide.bind(this))
+
+    // Configure the right nav button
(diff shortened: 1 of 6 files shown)


skellock:
Tests are ... uh ... complicated for React Native.   I'm sure it won't be in a few months, but right now, for any file that imports `react-native`, you must stub the universe.

(Because RN requires JSC and our desktops run Node.js.  I'm sure there's more to it, but that's as much as I care to know.)

Thanks for the PR's!

GantMan:
I'll bring in Mocha for tests and shallow render, I don't mind.   The only issue is we carry the cognitive load if we do.   In my opinion, it's better than no tests on components, and we can strip them out when we get there.

skellock:
We're mocha-ified already!  I don't think we have examples of chai & jsx assertions just yet.  Did you end up using Enzyme on your client's project?

If we have to change our non-test code to support the crazy mocking, we've gone too far though. (imho)

But yes, you're correct.  It's better than no tests.   :heart_eyes_cat:

GantMan:
We'd only have to change the code if we start using `.ios.js` and `.android.js`  OR we can update the compiler to look for those files if it fails to find regular `.js` files.   All good stuff.  In any case we're clear to use it for now.

Yes, I used Enzym on the client project.  Works great, but I'm a complete idiot when I'm returned virtual dom info.   Takes me a while to write a good test.

PR #19, opened 2016-03-28
Feature/animatable

GantMan:
### Pull-Request Checklist

- [x] Tests Pass

rn_animatable


markrickert:
oh, so you'll merge #19 but not #20, huh? Just because of your `.merge()` hangup? :sob:

skellock:
Haha.  I've seen `animatable` before.

You were creating immutable objects and feeding that into an Immutable library.    SUPER IMMUTABLE is the term, I believe.

GantMan:
No steve, it's still called IMMUTABLE, you can't change what you call it without throwing an error.

PR #20, opened 2016-03-28
Add redux persistence.

markrickert:
### Pull-Request Checklist

- [x] Tests Pass

Admittedly I have no clue how the `...` (splat) operator works in javascript. But it seems to work.

### DIFF
--- a/App/Reducers/LoginReducer.js
+++ b/App/Reducers/LoginReducer.js
@@ -9,24 +9,20 @@ export const INITIAL_STATE = Immutable({
 })
 
 // login attempts
-const attempt = (state, action) => {
-  return Immutable({ ...state, attempting: true })
-}
+const attempt = (state, action) =>
+  Immutable(state).merge({ attempting: true })
 
 // successful logins
-const success = (state, action) => {
-  return Immutable({ ...state, attempting: false, errorCode: null, username: action.username })
-}
+const success = (state, action) =>
+  Immutable(state).merge({ attempting: false, errorCode: null, username: action.username })
 
 // login failure
-const failure = (state, action) => {
-  return Immutable({ ...state, attempting: false, errorCode: action.errorCode })
-}
+const failure = (state, action) =>
+  Immutable(state).merge({ attempting: false, errorCode: action.errorCode })
 
 // logout
-const logout = (state, action) => {
-  return Immutable({ ...state, username: null })
-}
+const logout = (state, action) =>
+  Immutable(state).merge({ username: null })
 
 // map our types to our handlers
 const ACTION_HANDLERS = {
--- a/App/Reducers/LoginReducer.js
+++ b/App/Reducers/LoginReducer.js
@@ -10,19 +10,19 @@ export const INITIAL_STATE = Immutable({
 
 // login attempts
 const attempt = (state, action) =>
-  Immutable(state).merge({ attempting: true })
+  state.merge({ attempting: true })
 
 // successful logins
 const success = (state, action) =>
-  Immutable(state).merge({ attempting: false, errorCode: null, username: action.username })
+  state.merge({ attempting: false, errorCode: null, username: action.username })
 
 // login failure
 const failure = (state, action) =>
-  Immutable(state).merge({ attempting: false, errorCode: action.errorCode })
+  state.merge({ attempting: false, errorCode: action.errorCode })
 
 // logout
 const logout = (state, action) =>
-  Immutable(state).merge({ username: null })
+  state.merge({ username: null })
 
 // map our types to our handlers
 const ACTION_HANDLERS = {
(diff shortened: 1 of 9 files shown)


App/Reducers/LoginReducer.js
```diff
@@ -9,20 +9,24 @@ export const INITIAL_STATE = Immutable({
 })
 
 // login attempts
-const attempt = (state, action) =>
-  state.merge({ attempting: true })
+const attempt = (state, action) => {
+  return { ...state, ...{ attempting: true } }
```
GantMan: this would be written
`return { ...state, attempting: true}`

but that doesn't return an Immutable object (requirement of reducers by convention).

So it would become
`return Immutable({ ...state, attempting: true})`

Now, @skellock  is a fan of `merge` over spread operator.  Can you go into more detail on why this change was needed so we can make sure we're taking the best path.

GantMan:
@skellock this also means we need to add reducer tests that verify reducers are kept immutable.  As you can see all redux tests passed with Mark's changes.

App/Reducers/LoginReducer.js
markrickert: Cool. I've fixed the syntax. 

@skellock I have no clue why the new `state` object doesn't have the `merge` function on it, but when i added in `redux-saga` the previous `state.merge(x)` was throwing an error: `state.merge is not a function`.

markrickert: My guess is that the `autoRehydrate()` mtehod returns back a state object that's different than it used to without the `merge` mixin applied.

skellock:
Just a couple of notes here.

`Immutable` is just law-enforcement around an object.  Line 5 where we set the `INITIAL_STATE` shows the default value.  See how it's actually an `Immutable` object?  

This means, your reducers will *always* receive that as the first parameter.

`.merge` actually makes a new Immutable object.  Merge describes the intent, not the implementation.  That fact that we're using a `immutable` guarantees this.

You want:  `state.merge({fetching: true})`.

So the question is:  why didn't it work?

That's because of autoRehydrate.  Auto-rehydrate must persist it as JSON?  So when it comes back to life, it's a plain-old JSON object.

You might be able to fix this by calling `Immutable(autoRehydrate())`.  This way, rehydrate state is an Immutable object, what the reducers expect.

markrickert:
AHA! I'll try that and see if it works.

skellock:
Spread glues the props of one object into the props of another object.

```
const a = {a:1}
const b = {a:2, b:2}
const c = { ...a, ...b }  // {a: 2, b: 2}
const d = { ...b, ...a } // {a:1, b:2}
```

GantMan:
too late to stir the pot and ask if we _should_ be including a persist middleware in all base projects?

markrickert:
I tried to figure out how to do it on individual reducers but couldn't. That's why i made the `PERSIST` constant. At least we have a starting point here.

GantMan:
ahhh, I see.  Please move the constant to https://github.com/infinitered/react_native_base/tree/master/App/Config

skellock:
I can see the usefulness.   I also see there's a `blacklist` property where you can opt-out for certain branches of the state tree.  I like that.

App/Reducers/LoginReducer.js
```diff
@@ -9,24 +9,20 @@ export const INITIAL_STATE = Immutable({
 })
 
 // login attempts
-const attempt = (state, action) => {
-  return Immutable({ ...state, attempting: true })
-}
+const attempt = (state, action) =>
+  Immutable(state).merge({ attempting: true })
```
skellock: If we fix the issue with `autoHydrate()`, this should be able to say as:

`  state.merge({attempting: true})`

markrickert:
persist setting is moved to the `Config`.

I changed up the Immutable stuff... wrapping the `autoRehydrate()` call didn't work so i wound up with something like this:

```js
const request = (state, action) =>
  Immutable(state).merge({
    fetching: true,
    city: action.city,
    temperature: null
  })
```

skellock:
So I had a few kicks at `redux-persist` tonight.  It's not bad.  I had a few surprises.

Here's where I landed:

https://github.com/infinitered/react_native_base/compare/redux-persist...redux-persist-using-transform

The big change here was to move the `Immutable` conversion out of `CreateReducer.js` and into it's own `transform`.

It worked great once I figured out they are double converting your objects as they come out of storage.  There's a hacky way around that though, as you'll see in this diff.  Yuck.

markrickert:
We might want to post a link to our solution to the issue - https://github.com/rt2zz/redux-persist/issues/64

rt2zz:
@markrickert you are referring to the solution for using immutable objects as top level state? I definitely like the concept, I am working on redux-persist@3 and will strongly consider this for inclusion. Let me know how the rest of the implementation goes and if you run into any unexpected behaivor :+1:

PR #27, opened 2016-03-31
Add navigationStyle

leonskim:
Navigation title can be centered by styles but I think it would be nice if it's default on Android.

image

### DIFF
--- a/App/Navigation/NavigationBar.js
+++ b/App/Navigation/NavigationBar.js
@@ -10,6 +10,7 @@ export default {
   render () {
     return (
       <Navigator.NavigationBar
+        navigationStyles={Navigator.NavigationBar.StylesIOS}
         routeMapper={NavigationBarRouteMapper}
         style={styles.navigationBar}
       />


GantMan:
Guys, I don't know if I agree with this one.   Android's default is to left justify navigation title.

image

image

image

skellock:
I think that whole section is "special".  The android toolbar.  I saw that RN has a component for it, but have never used it.

Android UI also doesn't recommend you having a back arrow in the nav iOS screens often do.

This is one of those things that will grow as we do I guess.

Thanks for bringing the screenshots in as evidence.  It's helpful.   I'm still new to Android.

ryanlntn:
I suggest we consider using: https://github.com/react-native-fellowship/react-native-navbar

It's under active development and they've already thought through a lot of these issues: https://github.com/react-native-fellowship/react-native-navbar/issues/109

jamonholmgren:
@GantMan I believe Leon is suggesting just that, to left-justify on Android and center in iOS.

leonskim:
I thought about the hard back button on Android so if there's no left button on nav, the title is kinda in awkward position I think.

PR #38, opened 2016-04-05
CLI is ready, ignite should be working again with this in place

GantMan:
Thought it best to separate the generator from the CLI since they had to be 2 different names on NPM due to yeoman naming restrictions.

Works a bit cleaner this way anyhow.  Once the basics of the CLI are in place, we probably won't touch it much.  We'll probably just modify the generators.


ignite-cli/src/index.es
```diff
@@ -0,0 +1,37 @@
+#! /usr/bin/env node
+'use strict'
+import Program from 'commander'
+import colors from 'colors/safe'
+import pjson from './package.json'
+import Shell from 'shelljs'
```
jamonholmgren: The inconsistency here makes me 

crazy

GantMan:
lol, they each have their preferred way, we should just pick what we want.  Sounds like a good PR Jamon :)

PR #50, opened 2016-04-12
Move font sizes into Font from Metrics.

### LINKED ISSUE #49: Thoughts on changing font size location
I find myself doing this very often:

```css
  title: {
    fontFamily: Fonts.bold,
    fontSize: Metrics.fonts.medium,
    color: Colors.charcoal
  },
```

I think it would feel a lot "cleaner" to do this:

```css
  title: {
    fontFamily: Fonts.bold,
    fontSize: Fonts.size.medium,
    color: Colors.charcoal
  },
```

It would be a relatively easy change. When i thing of "font size" i don't think of the "metrics" object, i think of the "font" object.

### Make sure verify the following before submitting an issue.

- [x] Issue does not already exist
- [x] You have example code or some other reproduce experiment that we can verify

markrickert:
Closes #49.

Discuss.

### Pull-Request Checklist

- [x] Tests Pass (run: `npm run test && npm run test:client`)

### DIFF
--- a/ignite-base/App/Containers/AllComponentsScreen.js
+++ b/ignite-base/App/Containers/AllComponentsScreen.js
@@ -91,7 +91,19 @@ export default class AllComponentsScreen extends React.Component {
         {loggedIn ? this.renderLogoutButton() : this.renderLoginButton()}
         <Text style={styles.componentLabel}>{I18n.t('progressiveImageComponent')}</Text>
         <ProgressiveImage
-          style={styles.progressiveImage}
+          style={[styles.progressiveImage, styles.smallImage]}
+          defaultSource={Images.logo}
+          source='https://upload.wikimedia.org/wikipedia/commons/c/cc/ESC_large_ISS022_ISS022-E-11387-edit_01.JPG'
+          thumbnail='http://i.imgur.com/eVAFUhj.png'
+        />
+        <ProgressiveImage
+          style={[styles.progressiveImage, styles.mediumImage]}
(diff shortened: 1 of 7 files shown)


skellock:
Terrible idea.

skellock:
image

just kidding.  love it.

GantMan:
I like it, but let me play devil's advocate.

For an icon I would use `Metrics.icon.regular`, but for a font I'd use `Fonts.size.title`?   Then, would I be crazy to expect an image to be something like `Images.size.thumbnail`?

I like the idea of fonts containing font sizes for simple imports, but how do we best illustrate that font metrics belong outside of Metrics?

skellock:
I like that too.  `Images.size.massive`.  That feels right as well.

For other font metrics not in our examples, I'd expect for them to live inside Fonts as well.  `lineHeight` for example.

markrickert:
I'm basically saying that like things should be grouped together. Fonts have properties: `family`, `size`, etc. and it feels more "natural" to fo `Font.size.regular` instead of `Metrics.fonts.regular`. Typing `Metrics.fonts.regular` doesn't explicitly state that we're getting a `size` there... it infers a font object of size and face in my opinion.

GantMan:
@markrickert - can you add `Images.size.large` and perhaps use it for the progressive component, so that we entrench this pattern as we accept the PR?

markrickert:
I added an example of image sizes. Check it out and let e know what you think.

PR #52, opened 2016-04-13
Turns off a mystery PN setting which was breaking fresh apps.

skellock:
I actually don't know anything about this push notification component.

Seems like at startup, it's triggering a request to display a push notification.  This phantom notification is causing itself to trip up.

Setting this value to false seems to dodge that issue, but our app now runs.


GantMan:
@markrickert - can you investigate?  I'm gonna merge this in the meantime.

Issue #51, opened 2016-04-13
Old rnpm = Issues Installing


skellock:
When through the fresh install process tonight.

Looks like we'll need make sure `rnpm` is installed properly and is up-to-date.

The I18N library wasn't able to be linked properly.   

![image](https://cloud.githubusercontent.com/assets/68273/14481024/d1c07b58-00fc-11e6-8310-7b24d2cc6680.png)

After upgrading and re-linking, the I18N library was found.

GantMan:
Doesn't it notice and fix that? 
https://github.com/infinitered/ignite/blob/master/ignite-generator/src/shared/shared.js#L26-L31

GantMan:
I'm an idiot.  It passes rnpm check, just wrong version.  Gotcha.   What do you think we should do?  prompt user to upgrade?

GantMan:
it does warn, though:  https://github.com/infinitered/ignite/blob/master/ignite-generator/src/shared/shared.js#L34

skellock:
Well, if we don't upgrade, this project doesn't work.   I think we upgrade that warning to a prompt.  If they say no, they we should bail.

The warning may show, but i missed it.  It took nearly 10 minutes to install and create the first app.  I had long since tabbed away from the shell.

Maybe we could move the `rnpm` check earlier?  Maybe surround it with emoji?

Something to soothe the wandering minds of absentee shell script runners like me.

GantMan:
You're write, we should prompt the user.  Do you know what version we need to enforce?  Or do you think any npm outdated is worth the bail?

skellock:
Comes down to whatever minimum version is required to install any of our deps that need it.

The i18n lib needed either 1.5 or 1.6 to work.

Bail if we can't fulfill the minimum. Stern finger wagging if you're out of date but still good enough to run us.

PR #53, opened 2016-04-13
[Generator] - Critical tool restructure

GantMan:
This is a hopeful fix for the bug https://github.com/infinitered/ignite/issues/51
For now, I just tell the user "update your shit dude!" and have the script exit early.

I did a bit of speed restructuring as well.   Any `npm outdated` check is SLOWWWWWWW, so I only do them when it's mission critical.  This restructured our validation into 3 phases we can opt as the CLI grows `{ showWarnings, verifyTools, verifyExtensiveTools }`

The reason I got rid of the `shared` folder, is because any folder becomes a yoeman generator.  It's best to leave these in files at root of `src` that we can include. e.g. `validation.js`

Lastly, I removed the old `Generator` as it has become 100% defunct.


GantMan:
input from @markrickert  and @skellock 

I'll push a new npm version if/when this is merged.

skellock:
👍  from me.  I'd be lying if I told you i understand everything that's going on.  But i like the split!

Issue #37, opened 2016-04-04
move all semaphore checks to scripts


GantMan:
as of right now, if we modify code, we need to log in to semaphore and modify the checks/balances as well.  We should move all this to an npm script and call that.  This allows for local testing, and test modifications.

GantMan:
This also allows for better documentation on WHY we're doing certain checks.

PR #56, opened 2016-04-14
[ALL] Check script

GantMan:
This is my solution for https://github.com/infinitered/ignite/issues/37

Now when you modify the templates/structure you can modify the tester in the same PR, instead of modifying semaphore. 

Additionally, I've riddled it with comments so people can see WHY we're checking things in the current stage of the project life.   Besides failing, the script will output friendly error messages to help identify why the checks did not pass.

I've also take the liberty of enforcing standard compliance on the CLI and the Generator code bases, and even caught one linting error in the process.


GantMan:
### Failing
image

### Passing
image

Not the friendliest output, but functional for semaphore to run.

PR #77, opened 2016-04-24
[Base] - Reactotron add

GantMan:
#### Ignite Base Change

image

### DIFF
--- a/ignite-base/App/Config/ReactotronConfig.js
+++ b/ignite-base/App/Config/ReactotronConfig.js
@@ -1,0 +1,8 @@
+import Reactotron from 'reactotron'
+import {Platform} from 'react-native'
+
+Reactotron.connect({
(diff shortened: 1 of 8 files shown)


skellock:
There's a race condition the way that this is setup.

The problem is that calling `Reactotron.connect()` happens after your `store` is created.  It might not be totally obvious, but it's because of how es6 import statements resolve dependencies.

What I recommend doing here is adding a:

`App/Config/ReactotronConfig.js` and placing your `connect()` inside there.

Then, as the first line of code in each of your entry points, do a `import './App/Config/ReactotronConfig/js'`.

See: https://github.com/skellock/reactotron/blob/master/examples/ReactNativeExample/index.ios.js#L2

GantMan:
Damn!
damn

PR #81, opened 2016-04-28
fixed typo of Action Types Enum Literal

hknd:
- [Base] - Type was misspelled. It worked because it was registered in the action handlers via an `undefined` key and the unit test used the same literal (=undefined). I changed it to match the usage in the Reducer and Action Creator.

#### Ignite Base Change


GantMan:
### Nice catch!

How did I not see this before?
catch

Part of why we use constants is to stop typos.  JavaScript seems determined to fail silently at every turn.

@skellock - it's safe to say we NEVER want `undefined` as our Action type, so I'm thinking `createReducer` should throw if it finds any.

If you approve, I'll make the designated patch.

Thanks again @hknd !  This fix opened my eyes to potential epic issues that we can prevent.

skellock:
Oops!  Thank you.  Yes plz @GantMan

PR #97, opened 2016-05-16
[Base] - Adding Android to FastLane

mmazzarolo:
#### Ignite Base Change

--

Don't know if you're interested, let me know.  
P.S.:  Interesting threads to keep an eye on:  
[Increment build and version numbers for Android apps using Gradle](https://github.com/fastlane/fastlane/issues/878)    
[Gradle upgrades](https://github.com/fastlane/fastlane/pull/1082)

### DIFF
--- a/ignite-base/package.json
+++ b/ignite-base/package.json
@@ -9,8 +9,6 @@
     "newclear": "rm -rf $TMPDIR/react-* && watchman watch-del-all && rm -rf ios/build/ModuleCache/* && rm -rf node_modules/ && npm cache clean && npm i",
     "test": "mocha --opts Tests/mocha.opts Tests/ --recursive",
     "tron": "node_modules/.bin/reactotron",
-    "deploy-android": "fastlane android alpha", 
-    "deploy-ios": "node node_modules/react-native/local-cli/cli.js bundle --entry-file ./index.ios.js --platform ios --bundle-output ios/main.jsbundle bundle && fastlane ios beta", 
     "android:build": "cd android && ./gradlew assembleRelease",
     "android:install": "cd android && ./gradlew assembleRelease && ./gradlew installRelease",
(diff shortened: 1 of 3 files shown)


ignite-base/package.json
```diff
@@ -9,6 +9,8 @@
     "newclear": "rm -rf $TMPDIR/react-* && watchman watch-del-all && rm -rf ios/build/ModuleCache/* && rm -rf node_modules/ && npm cache clean && npm i",
     "test": "mocha --opts Tests/mocha.opts Tests/ --recursive",
     "tron": "node_modules/.bin/reactotron",
+    "deploy-android": "fastlane android alpha", 
+    "deploy-ios": "node node_modules/react-native/local-cli/cli.js bundle --entry-file ./index.ios.js --platform ios --bundle-output ios/main.jsbundle bundle && fastlane ios beta",
```
GantMan: I actually don't think you need to do the bundle command, it seems to do that for you in my experience.

GantMan:
@mmazzarolo that's awesome!   Couple things and then we'll merge this guy in.

## First: 
I don't think you need to run the bundle command anymore.  I haven't been doing it, and deploys still work.   Can you verify for `ios` that you have to run it?  Perhaps this is just a change depending on your version of RN?  @skellock  - when you deploy do you create a bundle?

## Second
Our CI is complaining, because any change to `package.json` needs to be reflected in `package.json.template`  (linked here)[https://github.com/infinitered/ignite/blob/master/ignite-base/package.json.template].   The template is what gets shipped when we ignite a new project, so our CI makes sure the two are always in sync.

👍  on the contribution.

ignite-base/package.json
mmazzarolo: You're probably right, I'm using that command since 0.17, I guess that it is not needed anymore 🌞

GantMan:
easy enough!

PR #127, opened 2016-05-28
[Generator] - Fixes the generators by removing the verify check

skellock:
Wouldn't mind your opinion on this @GantMan .

My vote is to remove the verification check for the generators that aren't 'app'.

Once your project has gotten started, I think it's safe to assume that `react-native` is installed.

There are other checks I'd like to perform though.  (like verifying the proper directory)

### DIFF
--- a/ignite-generator/component/index.js
+++ b/ignite-generator/component/index.js
@@ -6,8 +6,6 @@ var _createClass = function () { function defineProperties(target, props) { for 
 
 var _yeomanGenerator = require('yeoman-generator');
 
-var _validation = require('../validation');
-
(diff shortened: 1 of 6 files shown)


GantMan:
So the verify step was more of a placeholder than a need.   It's a set of basic checks and balances you can perform that should be "fast".   The old verify check had a comment saying such:  
https://github.com/infinitered/ignite/blob/0.2.4/ignite-generator/src/validation.js#L11-L18

I like the idea of checking the folder, and who knows what else we could want in the future. 

I'll merge this, where the stub is removed, but it would be good to add an issue requesting verification on folder as mentioned above.

PR #132, opened 2016-05-28
[CLI] Keeps the colours when running through the CLI.

skellock:
image

Apparently we want to spawn instead of exec.  Something about keeping the TTY and environment variables?  Oh Unix.  You so crazy.


GantMan:
## 👏  The colo(u)rs make it worth it!
effects

PR #176, opened 2016-06-21
- [Base] - Adding Rehydration guard

fvonhoven:
#### Ignite Base Change

Crashes and bugs can be caused when the app rehydrates the store after a reducer shape has changed.
The app has no way to know that the reducer shape has changed since the rehydrate feature will  resurrect the outdated reducers...until now...

* Here we move `reduxPersist` from `DebugSettings.js` to its own file (`ReduxPersist.js`) with its own set of properties.
* Additionally we will configure the store in `ReduxPersist.js` as well and make it a part of ReduxPersist since this store is what we are creating/persisting and it can be accessed by `RehydrationServices.js`
* Finally we will add a `reducerVersion` in `ReduxPersist.js` to track changes in reducer shape and notify `RehydrationServices` that a change has occurred. 

##### How it works
* When the app is loaded, it will check for an update with `RehydrationServices`
* If the version number in persisted storage is different that the one in `ReduxPersist`, then the app will purge the current store and rebuild a new one with the updated reducer(s) shape

#### Why do we need this?

Let's say you remove an element(s) from the reducer shape (we're going to remove `city` and `state` since we can get both from just the zip code)
```
export const INITIAL_STATE = Immutable({
  city: null,
  state: null,
  zipCode: null
})
```
goes to:
```
export const INITIAL_STATE = Immutable({
  zipCode: null
})
```
##### COOL 👍 

But now our app will rehydrate with the old reducer shape and expect `city` and `state` to exist!!!

#### RehydrationServices to the rescue!
Because we bumped our `reducerVersion`, RehydrationServices will know that we would rehydrate with an outdated reducer shape, so...

alt tag

### YAY!!!

@GantMan This is a preventative feature that could be more fully explained and elaborated upon in a blog post 😉

### DIFF
--- a/ignite-base/App/Config/DebugSettings.js
+++ b/ignite-base/App/Config/DebugSettings.js
@@ -3,7 +3,6 @@ const SETTINGS = {
   ezLogin: false,
   yellowBox: __DEV__,
   reduxLogging: true,
-  reduxPersist: true,
   includeExamples: __DEV__
 }
 
--- a/ignite-base/App/Config/ReduxPersist.js
+++ b/ignite-base/App/Config/ReduxPersist.js
@@ -1,0 +1,15 @@
+import immutablePersistenceTransform from '../Store/ImmutablePersistenceTransform'
+import { persistentStoreBlacklist } from '../Reducers/'
+import { AsyncStorage } from 'react-native'
+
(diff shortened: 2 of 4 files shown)


GantMan:
nice

PR #193, opened 2016-06-24
[Base] Changed I18n.js link in README.md

enapupe:
the current link seems to be pointing to some older repo structure


GantMan:
good catch

PR #198, opened 2016-06-25
[Generator] - added .editorconfig to newly ignited project

GantMan:
newly ignited projects will grab the `.editorconfig` file, and like it!


GantMan:
this PR testable by the grace of @skellock 's branch args.

PR #216, opened 2016-06-29
don't forget fastlane

GantMan:
- [x] `fireDrill.sh` passed

## Describe your PR
Though we added fastlane to the template, we never copied it over!

Do so now.


skellock:
My bad.

PR #220, opened 2016-07-01
Added more obvious Gitter link to README

jamonholmgren:
Would like to direct more people to Gitter to discuss Ignite.


GantMan:
good call 👍

Issue #215, opened 2016-06-28
Problem with the Drawer on latest update


sudharsan1988:
- [x] Issue does not already exist
- Identify exactly what aspect your issue is about
  - [x] ignite base application
  - [ ] ignite the cli
  - [ ] ignite generated code

I have added "Hamburger" icon to a screen which is already available on the Drawer. Here is the Problem, when I press the menu item, it takes me to the corresponding screen. If I press the same menu item again, it loads a empty screen along with the below warning,

Warning: flattenchildren(...): Encountered two children with the same key, 'scene_1'. Child keys must be unique; when two children share a key, only the first child will be used.

Steps to reproduce:

1. Add a "Hamburger" icon on "Device Info Screen" as right-button
2. Open the drawer and press the "Device Info Screen"
3. Once you are on the "Device Info Screen", again press the "Device Info Screen"

GantMan:
@fvonhoven  - ping me when you can about this issue.   We'll probably need to figure out an elegant solution.

sudharsan1988:
Hi, any luck with this? Just looking for a solution to sort this out :) By the way, I am trying to add a background image to the drawer. Is it possible with the current version of Ignite?

skellock:
The screen navigation problem will go away once we merge in `react-native-router-flux`.  They have a solution for this.

For the background image, you can add an `<Image>` to your `DrawerContents.js` file.

GantMan:
@sudharsan1988 - we're going to be upgrading to RNRF this week likely.   This will cause a bit of change, but also a bit of advancement in the capabilities of navigation.

GantMan:
@sudharsan1988 - please check on this with 1.3.1 👍

Issue #29, opened 2016-03-31
NavigatorExperimental Away Mission


skellock:
`<Navigator>` is sun-setting.

`<NavigatorExperimental>` is transitioning in (the name will change).

I'm hearing nothing but great things about it.  Redux like apparently.

Anyone up for spiking it through before its officially released?   This is my only hesitation in picking a 3rd party library @ryanlntn for Navigation routing & presentation.  

It's all about to change.

jamonholmgren:
The notification email was funny, because it removed the code tags:

<img width="426" alt="screen shot 2016-03-31 at 1 37 03 pm" src="https://cloud.githubusercontent.com/assets/1479215/14190071/ae4c4938-f745-11e5-935d-5a73111512fc.png">

skellock:
is sorry.

is fixed.

:smiley_cat:

skellock:
@GantMan is on this.

mmazzarolo:
Hey @GantMan, are you still working on the Navigation Experimental?  
I've been working on an app that uses almost the same structure of ignite but I always used `react-native-router-flux`... which is a mess since it's last update.  
I was looking for a substitute for the navigation and I'm still undecided between the standard navigation and the experimental one, any hint? :)  
Thank you!

GantMan:
Hey @mmazzarolo :)

I spent a whole 2 days fighting the new Navigation Experimental, and it just wasn't ready.  It hurt so bad!  LOL.   I saw he was updating it every day, so I decided to give it a bit of time and back off.   Now it might be time to come back.  I'll start carving some time to venture off again.   If anyone else wants to take a look, I'm happy to download what I've learned to them.

mmazzarolo:
@GantMan i toyed a bit with it this morning, it *seems* to be working fine in my project (still need to add a drawer and some more tricky routes).  
If you're interested I created [this Gist](https://gist.github.com/mmazzarolo/5a4f235b6b07fa14247c0fb081b1f13f) with my current setup... I'm following most of your coding standard but you should style it a bit (especially `NavigationHeader.js`).  
To navigate between scenes I'm calling `navigatePush('routeName')`, but adding a specific action for each route might be cleaner.

**Disclaimer:** I've stolen and then tweaked all that boilerplate from [here](https://github.com/jlyman/RN-NavigationExperimental-Redux-Example) 🐼

GantMan:
[...]

PR #235, opened 2016-07-06
Router Flux Navigation

kevinvangelder:
- [x] `fireDrill.sh` passed

## Describe your PR
Refactors entire ignite-base app to use react-native-router-flux!


kevinvangelder:
This should resolve #215

kevinvangelder:
Closes #29.

ignite-base/App/Themes/ApplicationStyles.js
```diff
@@ -34,13 +34,15 @@ const ApplicationStyles = {
...
+      fontFamily: Fonts.type.bold,
+      backgroundColor: Colors.transparent
     },
     subtitle: {
       color: Colors.snow,
       padding: 5,
       marginBottom: 5,
-      marginHorizontal: Metrics.smallMargin
+      marginHorizontal: Metrics.smallMargin,
+      backgroundColor: Colors.transparent
```
GantMan: why did you have to do this?   I'd like to get away from Text having default background colors that have to be overridden.

GantMan: Just found out if you set the `contentContainerStyle` background to transparent it fixes this.... no clue.   Let's do that instead, though, rather than affecting every text item.

ahmed1490:
one question:
does android back button works in this?

ruskid:
Looking forward for this release =)

nonotest:
Same here !

skellock:
Me too!   Actually, I think now that 0.29.1 is in the clear & stable, there's nothing holding us back from finalizing this branch in.

@ahmed1490 :  back button is independent from *everything*.  it's absolutely possible to use it with this though if your navigation stacks is simple push & pop though.  It only gets a little tricky when you introduce tabs and modals.  And the trickiness is only from deciding what you want to do, not actually implementing it.

nonotest:
What is missing for this now? do you guys need any help?

cheers,

ruskid:
@nonotest I am already using this commit. Works great.
Maybe they need to clean the things up and modify README, cause there is a new package.

GantMan:
I think we just fell behind a little this week.   I do have a comment on removing the background color changes, which I would rather have on the scrollviews.   We'll update this soon, as well as adjust Ignite to 29.1

GantMan:
@nonotest - please feel free to update this branch if you'd like, so it can be ready to merge.  Unless someone else beats you to it :)

Issue #233, opened 2016-07-06
Support Version Locking for React Native


skellock:
Turns out we can pass a command line option into `react-native init --version 0.27.0`

https://github.com/facebook/react-native/blob/master/react-native-cli/index.js#L53

We should consider this for our generator, so we don't get bit like we did today with upstream bugs on React Native 0.29.

GantMan:
we should lock to 0.28 today, this is a great find.

GantMan:
completed via #236

PR #236, opened 2016-07-07
Quick fix for issue #233. Lock the current react-native version to 0.28

tminard:
- [X] `fireDrill.sh` passed

## Describe your PR

- Locks the react-native version used to init the base template to 0.28.0. Since the base directory assumes 0.28, I felt hard-coding the current version was appropriate.

### DIFF
--- a/ignite-generator/src/app/index.es
+++ b/ignite-generator/src/app/index.es
@@ -202,7 +202,7 @@ export class AppGenerator extends Generators.Base {
     this.spinner.text = status
     const done = this.async()
     const command = 'react-native'
-    const commandOpts = ['init', this.name]
+    const commandOpts = ['init', this.name, '--version 0.28.0']
     this.spawnCommand(command, commandOpts, {stdio: 'ignore'})
       .on('close', () => {
         this.spinner.stop()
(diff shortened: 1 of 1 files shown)


tminard:
@GantMan @skellock Is this what you were thinking for 0.28?

skellock:
Yes!  

It might need to become 2 parameters because of the space between `--version` and `0.28.0` though.  I'd have to test that.

But yes.  That's exactly it.

Just to need to run a few tests to make sure all is well.

Thank you @tminard

GantMan:
This looks great!   Couple things.

I'd like that moved into two params, and the second param I'd like it to pull from a well placed variable.   
A const at the top of the file is fine.

We're going to eventually need to move this as a possible parameter, so we can test locally with different RN versions before we release to the public.  But I feel that might be out the scope of your PR.  Perhaps it's a secondary ticket.

Thanks @tminard

GantMan:
This is perfect.   Yeah, the next thing I'd like to do is add that parameter at some point.  It would be exclusively for us to test versions of RN before changing the lock version here in the repo.

We can do it via code for now :)

Issue #241, opened 2016-07-08
Failing `ignite new App` on external lib installation


martincik:
### Generating new app isn't working for me

The problem is with `Ignite generated code` specifically `ignite new MyApp` fails with:
```
Linking external libs./myApp/ios/myApp/Info.plist: file does not exist or is not readable or is not a regular file (Error Domain=NSCocoaErrorDomain Code=260 "The file “Info.plist” couldn’t be opened because there is no such file." UserInfo={NSFilePath=./myApp/ios/myApp/Info.plist, NSUnderlyingError=0x7fc55a400830 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}})
```

My setup:
```
npm 3.10.3
node v6.3.0
```

GantMan:
what version of ignite do you have installed?
`ignite --version`

Also, that error looks like init failed, and was swallowed by the installer.  Can you verify you can properly build a react native project `react-native init TestForGant` without error.

GantMan:
@martincik - any update on your setup?

Novex:
I hit this tonight as well, first time running ignite :(

`ignite --version` reports 1.3.0 and a normal `react-native init TestForGant` generates everything properly. [Here's the full console output](https://gist.github.com/Novex/504d9fd80dc75adcd1836445ba2226dd), as well as an `ls` of the resulting directory which confirms the ios/android directories didn't get created.

Let me know if there's anything else I can run to help narrow down the problem.

Setup:
```
npm 3.7.3
node v5.9.0
```

Novex:
Got to the bottom of it!

react-native-cli was an older 0.1.10 version, which was causing [reactNativeInit](https://github.com/infinitered/ignite/blob/master/ignite-generator/src/app/index.es#L179) to error out.

running `npm update -g react-native-cli` brought it up to 1.0.0 and fixed the problem, but I was curious why the check in `findReactNativeCli` didn't catch this. There are a few reasons - pull request incoming.

GantMan:
awesome!

GantMan:
thanks again @Novex  - you got a shout out in https://github.com/infinitered/ignite/releases

Closing this ticket.

johneo:
[...]

PR #243, opened 2016-07-10
Fix the react-native-cli version check

Novex:
- [x] `fireDrill.sh` passed

## Describe your PR
The react-native-cli version check would always fail for a few different reasons:
* `Shell.exec` returns the exit code in an object, not as the return value itself
* The grep regex needed to be escaped, otherwise javascript throws in control characters
* The silent option needed to be set so we don't get the version string dumped into the console

Also when some checks did fail, the error message would be displayed in a run-on line instead of its own.

This PR should resolve both these things and avoid more issues like #241

### DIFF
--- a/ignite-generator/src/app/index.es
+++ b/ignite-generator/src/app/index.es
@@ -133,14 +133,12 @@ export class AppGenerator extends Generators.Base {
     this.spinner.text = status
     this.spinner.start()
     if (!isCommandInstalled('react-native')) {
-      this.log(`${xmark} Missing react-native - 'npm install -g react-native-cli'`)
-      process.exit(1)
+      this._logAndExit(`${xmark} Missing react-native - 'npm install -g react-native-cli'`)
     }
 
     // verify 1.x or higher (we need react-native link)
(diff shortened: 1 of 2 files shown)


ignite-generator/src/app/index.es
```diff
@@ -133,14 +133,12 @@ export class AppGenerator extends Generators.Base {
...
-      this.log(`${xmark} Missing react-native - 'npm install -g react-native-cli'`)
-      process.exit(1)
+      this._logAndExit(`${xmark} Missing react-native - 'npm install -g react-native-cli'`)
     }
 
     // verify 1.x or higher (we need react-native link)
-    if (!Shell.exec("react-native -v | grep 'react-native-cli: [1-9]\d*\.\d\.\d'")) {
-      this.log(`${xmark} Must have at least version 1.x - 'npm install -g react-native-cli'`)
-      process.exit(1)
+    if (Shell.exec("react-native -v | grep 'react-native-cli: [1-9]\\d*\\.\\d\\.\\d'", {silent: true}).code > 0) {
```
GantMan: ahhh, I can never remember what these things return.   Great catch, and I love the `_logAndExit`!

If we open this code up again, I'm thinking we should check `!== 0` since I've heard lore of people returning `-1` but I know grep doesn't.

Great fixes!

Issue #192, opened 2016-06-24
Use npm git hooks as package instead of standalone


enapupe:
in `ignite base application`, it is better to use a npm package like https://www.npmjs.com/package/ghooks or https://www.npmjs.com/package/pre-commit to handle git hooks instead of an isolated solution (like `brew install git-hooks`), this way, cloning and installing the project guarantees the hooks are in place.

GantMan:
Good suggestion.  I'll give this a look tomorrow.

GantMan:
completed in #247

PR #247, opened 2016-07-14
Use ghooks for git hooks

ryanlntn:
- [✔] Everything works on iOS/Android
- [✔] `ignite-base` **ava** tests pass
- [✔] `fireDrill.sh` passed

## Describe your PR
Addresses https://github.com/infinitered/ignite/issues/192

Uses the ghooks npm package to implement pre-commit linting. Removes previous git hooks implementation.

@GantMan Please review.

### DIFF
--- a/ignite-base/package.json
+++ b/ignite-base/package.json
@@ -5,6 +5,7 @@
   "scripts": {
     "start": "node node_modules/react-native/local-cli/cli.js start",
     "lint": "standard --verbose | snazzy",
+    "lintdiff": "git diff --name-only --cached --relative | grep '\\.js$' | xargs standard | snazzy",
     "clean": "rm -rf $TMPDIR/react-* && watchman watch-del-all && npm cache clean",
     "newclear": "rm -rf $TMPDIR/react-* && watchman watch-del-all && rm -rf ios/build/ModuleCache/* && rm -rf node_modules/ && npm cache clean && npm i",
     "test": "ava",
@@ -80,7 +81,7 @@
   },
   "config": {
     "ghooks": {
-      "pre-commit": "npm run lint"
+      "pre-commit": "npm run lintdiff"
     }
   }
 }
--- a/ignite-base/package.json
+++ b/ignite-base/package.json
@@ -81,7 +81,7 @@
   },
   "config": {
     "ghooks": {
-      "pre-commit": "npm run lintdiff"
+      "pre-commit": "git diff --name-only --cached --relative | grep '\\.js$' | xargs standard | snazzy"
     }
   }
 }
--- a/ignite-base/package.json
+++ b/ignite-base/package.json
@@ -81,7 +81,7 @@
   },
   "config": {
     "ghooks": {
-      "pre-commit": "git diff --name-only --cached --relative | grep '\\.js$' | xargs standard | snazzy"
+      "pre-commit": "if [ -d 'ignite-base' ]; then cd ignite-base; fi; npm run lint"
     }
   }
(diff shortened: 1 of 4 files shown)


GantMan:
sorry for the delay, will review soon.

ignite-base/package.json
```diff
@@ -76,5 +77,10 @@
       "XMLHttpRequest",
       "FormData"
     ]
+  },
+  "config": {
+    "ghooks": {
+      "pre-commit": "npm run lint"
```
GantMan: should it lint everything or just the files that are staged?  The previous `lint.sh` would just lint staged files.   Open to thoughts.

GantMan:
2 things @ryanlntn - Firstly, what do you think about adjusting this to lint only staged files of the commit, instead of all files. 

Second, you should update the README that gets generated:
https://github.com/infinitered/ignite/blob/master/ignite-base/README.md.template

ryanlntn:
@GantMan Done and done. 😄

GantMan:
Ryan, I'm trying to test this branch, and I get the following:
image

As you can see, it's not working very well for me.   Can you verify?

GantMan:
@ryanlntn did you verify this fixes the issue?

ryanlntn:
@GantMan Yeah. The only issue you might have is if you don't have snazzy installed globally it will complain when working on the repo. This shouldn't be a problem for endusers.

GantMan:
@ryanlntn - more problems.
It's complaining even though the file is legit.   Lint works fine, but I'm thinking the problem might be that it can't understand that it is ECMA2015
image

My guess is normally it runs in the same folder with a proper `.babelrc` and now it's not happy.   Thoughts?

GantMan:
NOTE: We may have to go down this road: https://github.com/julien-f/babel-standard or something similar.

ryanlntn:
@GantMan Any reason we couldn't duplicate the .babelrc in the project root?

GantMan:
I don't see an issue with that, but let's get a sign off from @skellock - as he's a bit more fastidious with `.babelrc` files than I.  Also, I just tried copying the file, to verify that is the issue, and it didn't work.  So now I'm not so sure.

ryanlntn:
@GantMan Yeah it's kind of weird because I did a test run locally and was able to commit the change without that error showing up.

skellock:
I wouldn't place the .babelrc in the root if it were me.  The root isn't a node project.

It certainly looks like it cannot read the package.json file.

I'm wondering if it has to do with piping.  Perhaps when we pipe into xargs we lose the current working directory somehow?  

Maybe we have to pass args `--plugin babel-eslint` to that command?  (seems hax)

Are we able to pipe into `npm run lint`?  (seems lame)

ryanlntn:
Per ghooks README:

> Note: The hooks' working directory is relative to the git root (where you have your .git directory). This means that if your package.json is in a subdirectory of your git repo, you'll need to cd into the directory before running any npm scripts. E.g.:

I've updated to the following: `"pre-commit": "if [ -d 'ignite-base' ]; then cd ignite-base; fi; npm run lint"`

GantMan:
where's this PR at?  @skellock  have you verified the full process?   Should I?

skellock:
I was unable to get it to run.  But @ryanlntn since checked in some commits.  I'll have another look on Monday if it's still open.

skellock:
Ok.  Just tested.  Works great.  Green light from me.

GantMan:
Good for you, good for me.

PR #269, opened 2016-07-24
[CLI] Fixes the RN dependency version check.

skellock:
The `ignite doctor` was bombing when trying to load RN's package.json.  

Turns out, the `cwd` and the `require` root can be two different things.

My bad.


GantMan:
good to know!

PR #297, opened 2016-08-16
using nyc html reporter instead of text

hiaw:
- [x] `fireDrill.sh` passed

## Describe your PR

Using the html reporter for nyc coverage.

### DIFF
--- a/ignite-base/package.json
+++ b/ignite-base/package.json
@@ -10,7 +10,7 @@
     "newclear": "rm -rf $TMPDIR/react-* && watchman watch-del-all && rm -rf ios/build/ModuleCache/* && rm -rf node_modules/ && npm cache clean && npm i",
     "test": "ava",
     "test:watch": "ava --watch",
-    "coverage": "nyc ava",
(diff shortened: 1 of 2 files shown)


GantMan:
I like this idea.   Couple things.   1, we should have a nice option to run in terminal too.  Maybe `npm run cov` in classic *NIX shortcut, OR, perhaps it prints to the console AND does the HTML?  Thoughts @skellock ?

The second issue, is a bit more substantial.   If you don't have the folder created, this errors.  So first time users will see errors. 
image

Actually, it looks like nyc ava report --reporter=html doesn't generate latest output, just converts last output to HTML.  Thoughts?

hiaw:
Sorry I didn't check for first use. Now it should work first time. It will also give both text and html output. 

As for terminal, it should work `npm run coverage`

skellock:
Not at my computer, so cant verify, but I'm behind this if the html report is awesome!  Thanks!

GantMan:
@hiaw  - works great!

hiaw:
if you just type `npm run coverage` into the terminal then should work

PR #302, opened 2016-08-17
Translation templates

hiaw:
- [x] `fireDrill.sh` passed

## Describe your PR
Allow for more predefined languages for both platforms


skellock:
Wow!

hiaw:
actually don't accept the pull request yet. I think I did something wrong there. Currently every component that require I18n will have to reload the translation file. That's not the expected behaviour.

hiaw:
ok. should be fine now.

GantMan:
Looks good to me!   Awesome template!   @skellock @carlinisaacson @ryanlntn thoughts before merge?

skellock:
I like almost all of what I'm seeing.

1. Are the `_` prefixed files necessary?  Could we just call it `translations.js` and `en.json`?
1. I'd actually prefer to make these `.js` files.  It might open interesting reuse options.  For example, a better fallback for Canadian French would be French, not Canadian English.
1. `id` and `in` are both mapping to the same file... it looks intentional, but just wanted to point it out.
1. We currently use themes as a catch all for "hash-based" data.  Could we consider placing this in `Themes/index.js` as the access point?
1. I'm not a fan of a git dependency for including this... could we just put the template inside this repo?

^^ all this is nitpicking, i love this PR.

hiaw:
>  Are the _ prefixed files necessary? Could we just call it translations.js and en.json?
    
I think the _prefix is nice but I can change it if you want. I did it so that it's clear that those are the primary files one should work with and the other languages are optional.

>  I'd actually prefer to make these .js files. It might open interesting reuse options. 

I'm not sure how to make them .js files. Haven't tried.

> For example, a better fallback for Canadian French would be French, not Canadian English.

Yes, Canadian French will fallback to French if French is provided. Otherwise, it falls back to English

>  id and in are both mapping to the same file... it looks intentional, but just wanted to point it out.
 
Yeap, they both point to Indonesian language. Some Android quirk.

> We currently use themes as a catch all for "hash-based" data. Could we consider placing this in Themes/index.js as the access point?

I'm fine with that.

> I'm not a fan of a git dependency for including this... could we just put the template inside this repo?

Yes, that is the intention. I did remove the .git folder and the example folder. I just left the README.md so user can refer to it. Let me know if this is what you mean?

hiaw:
Another point with json file is that it's easier to work with translation company as it's a common format that they would work with

skellock:
Good point with the 3rd party translation. 👍 I've never worked with translators, so thank you for pointing that out.  Also, I didn't realize how the `fr` fallback worked either. Thx.

I think that line 8 of `translation.js` explains the importance of the base file.  I vote to remove the underscores.  But I hear what you're saying.  I could be convinced to keep 'em.

Maybe extend our Themes a bit to return back our access to it?

`I18n`?  or just the function `t`?  Or `translate`?   Or something else?

hiaw:
I've rename `_translation.js` => `I18n.js`, `_en.js` => `english.js`

After looking at the Themes. I've decided it's still best to be kept in the App folder. The app text is really part of the app and not part of themes. 

For usage, the user only need to `import I18n from 'react-native-I18n'` instead of from `'../I18n/I18n.js'` or `'../Themes/I18n/I18n.js'` or other derivatives depending on where the file is located. So `I18n.js` doesn't need to return anything.

GantMan:
@hiaw -  Great work on this PR, we'll test and report || merge in the next few days.

PR #303, opened 2016-08-18
Upgrades Apisauce from 0.3.0 to 0.4.0.

skellock:
Apisauce now supports transforms.  Both request & response.

```js
// install it like a monitor
api.addResponseTransform(response => {
  // check for conditions
  if (response.data.email === 'gant@infinite.red') {
    // and mutate the data before it arrives in your other code
    response.data.password = 'p@ntsl3ss_c0d1ng'
  }
})
```


GantMan:
That's awesome!  Was that designed because of needs for logout?

skellock:
Both you & @kevinvangelder were asking about it for a while. Today someone else did too: https://github.com/skellock/apisauce/issues/23

Another interesting use for this is a stop-gap patch for an API.  Let's say the API isn't returning a few fields you need.  You've asked the dev to do it, but they're busy.  Drop a transform in until they can patch it up.

PR #304, opened 2016-08-18
Add react-native-config support

ryanlntn:
- [x] `fireDrill.sh` passed

## Describe your PR
Adds [react-native-config](https://github.com/luggit/react-native-config) support to Ignite. Similar to dotenv.

### DIFF
--- a/ignite-base/.gitignore
+++ b/ignite-base/.gitignore
@@ -35,11 +35,17 @@ node_modules/
 npm-debug.log
 
 # Coverage
+#
 .nyc_output/
 coverage/
 
 # BUCK
+#
 buck-out/
 \.buckd/
 android/app/libs
 android/keystores/debug.keystore
+
+# Secrets
+#
+.env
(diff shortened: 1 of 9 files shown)


skellock:
Beautiful!

GantMan:
@ryanlntn if there is no .env file detected, does `react-native-config` fail gracefully?

Do we need to generate a .env file when we generate an app?

ryanlntn:
@GantMan I'm not sure I haven't tried yet. I do think it'd be a good idea to generate a `.env` file regardless. Maybe you could help me figure that out later today?

GantMan:
Sounds like a plan

jgkim:
@GantMan @ryanlntn It doesn't seem like that is the case. Cf. luggit/react-native-config#7

GantMan:
Doesn't or does?  It seems to be a filed issue.  Either way we're updating Ryan's code to place an example `.env` file.  We're beginner friendly 👍

jgkim:
@GantMan Oops, I meant if there is no .env file detected, `react-native-config` **doesn't** fail gracefully. ;)

GantMan:
hahaha, gotcha :)   All good, Ryan will have a fix for us soon.   I've requested that the example secret say "taco" somewhere.   We'll see what bubbles up.

PR #313, opened 2016-08-22
Gitignore

hiaw:
- [X] `fireDrill.sh` passed


GantMan:
@hiaw  - good call, but this is already done:   https://github.com/infinitered/ignite/blob/master/ignite-base/.gitignore#L39-L40

Your PR actually adds duplicates.

PR #316, opened 2016-08-25
standard --fix to fix code according to standardjs

hiaw:
- [X] `fireDrill.sh` passed

## Describe your PR
To format the code correctly according to standardjs, one can use the `standard --fix file` CLI. But the quirk there is that it doesn't like the `static propTypes = {` line within the class. So instead, using `ClassName.propTypes = {` will allow the CLI to work properly. 

I personally also thinks it makes the code cleaner to seperate out the propType from the class. 

ps: If you look at Generator.js in this PR, you can see `standard --fix` at work.

ps: I have `standard --fix` running each time I save js files on my editor, that's why I know it wasn't working.

### DIFF
--- a/ignite-base/App/Components/AlertMessageComponent.js
+++ b/ignite-base/App/Components/AlertMessageComponent.js
@@ -21,18 +21,6 @@ ExamplesRegistry.add('Alert Message', () =>
 )
 
 export default class AlertMessage extends React.Component {
-
-  static propTypes = {
-    style: React.PropTypes.object,
-    title: React.PropTypes.string.isRequired,
-    icon: React.PropTypes.string,
-    show: React.PropTypes.bool
-  }
-
-  static defaultProps = {
-    show: true
-  }
-
   render () {
     let messageComponent = null
     if (this.props.show) {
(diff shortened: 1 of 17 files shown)


GantMan:
LGTM 👍 

@skellock  andy feedback?

skellock:
This is from `standard` 8.  Which we currently have a global dependency on.  I recommend we start rolling this stuff into dev dependencies.

GantMan:
@skellock  what exactly do mean?  
For this PR?  or for our overall structure?

skellock:
Also, this is linting some generated code, which can we can avoid in `firedrill`.

But ya... Love it!  Can't wait to switch to the new hawtness.  Thanks!

skellock:
@GantMan This PR adds `standard --fix` to an npm script in base.  This will fail unless the dev's machine has standard 8 globally installed.  I recommend that in base... we bump standard up to v 8 as well.

Global dependencies are best avoided at the project level.  I don't mind so much at the Ignite Contributor level.

GantMan:
Maybe we should work together to make that namespaced CLI lib, and we can add that as a dependency?

Take a look at what I'm doing in commander, and tell me if you think it's a terrible idea:   https://github.com/GantMan/mobile-commander/blob/master/bin/lintme.sh

GantMan:
My goal was to enforce latest standard.

skellock:
@hiaw :  Could you bump us up to standard `^8.0.0` in `ignite-base/package.json` & template and we'll get this merged if @GantMan is cool with it!  👍 

@GantMan it's the `-g` that's a bit concerning.   Projects generally should be able to control their dev dependencies.  Including `standard`, `snazzy`, and `babel-*`.  I'd help you figure this out with `mobile-commander` or `ignite` too.  Whatever you need man!

hiaw:
sorted.

PR #324, opened 2016-08-27
simpler way to solving the undefined prop problem

hiaw:
- [X] `fireDrill.sh` passed

## Describe your PR
fixing my previous PR regarding drawer initial state


GantMan:
I like this so much better!  Technically you're supposed to treat props as pure, and we violated that.  I knew there had to be a better way, and you found it :)

PR #326, opened 2016-08-27
Updates Reactotron to 1.1.0 and adds some in-app samples.

skellock:
image

image

Heads up, there's a warning about peer dependency of React Native with `reactotron-react-native`.  I'm going to loosen up the requirement on my end from `^0.31.0` to `>=0.31.0`.  

But it's harmless.  You're welcome to hold off on this merge, but it might not be for a few days before I can get to that task.

https://github.com/reactotron/reactotron/issues/186

### DIFF
--- a/ignite-base/App/Config/ReactotronConfig.js
+++ b/ignite-base/App/Config/ReactotronConfig.js
@@ -1,14 +1,37 @@
-import Reactotron from 'reactotron'
-import {Platform} from 'react-native'
-
-Reactotron.connect({
-  enabled: __DEV__,
-  name: 'ignite App',
-  userAgent: Platform.OS
-})
+const Reactotron = require('reactotron-react-native').default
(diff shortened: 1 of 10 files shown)


GantMan:
Beautiful steve!

BTW, I'm looking into our API change.   Should I add the API key and just keep it weather for now?  Or do you think switching to Github is best?

skellock:
I really like Github.  I think fellow devs might too!

Here's a head start if you go that direction.  I did this on the Reactotron React Native demo too.

https://github.com/reactotron/reactotron/tree/master/packages/demo-react-native/App/Sagas

skellock:
btw:  they do have a limit without an API key.  it's 60 per hour.  It's more than enough because cached requests aren't included in that total.

skellock:
Plus if we query the React Native repo on github, there's always a chance to see the Ganticorn.

GantMan:
Just got around to testing this.  Works like a charm.

PR #330, opened 2016-08-29
Fixes problem with chicken-and-env for react-native-config.

skellock:
Finishes hooking up `react-native-config` to Android and temporarily switch to a branch which doesn't die when building projects missing an `.env` file.

https://github.com/luggit/react-native-config/pull/31


skellock:
Self Merged.  I did however, review it in a british accent.  So pretty much good to go.

GantMan:
as long as it wasn't tea time, you're good.

PR #331, opened 2016-08-29
Bumps Reactotron due to version warning inside that dependency.

skellock:
Switched Reactotron's React Native dependency from `~0.31.0` to `>=0.31.0` because it was throwing warnings.  Sorry about that.


skellock:
Back to back self merges.

image

PR #339, opened 2016-08-31
Rehydration services repair

### LINKED ISSUE #290: State is not persisted on app restart
Hi all!!!

I'm having some issue in my application (both android and iOS Simulator) where the store is never persisted after app restart (I need to persist some facebook login/token data).

I tried to generate a new app and ended up with the same.
I could reproduce the problem in StartupSaga.js where state.weather.temperature is always null. I'm not sure that I'm doing something wrong, maybe some permission?

- [X] Issue does not already exist
- Identify exactly what aspect your issue is about
  - [X] ignite base application
  - [X] ignite the cli
  - [X] ignite generated code

```
Computer
  Platform: darwin

Ignite
  Version: 1.5.0
  Path: /Users/fernandomorais/.nvm/versions/node/v6.3.1/bin/ignite

Node
  Version: v6.3.1
  Path: /Users/fernandomorais/.nvm/versions/node/v6.3.1/bin/node

NPM
  Version: 3.10.3
  Path: /Users/fernandomorais/.nvm/versions/node/v6.3.1/bin/npm

Yeoman
  Version: 1.8.4

React Native CLI
  Version: 1.0.0

App
  React Native Version: 0.31.0
```

fvonhoven:
- [x] `fireDrill.sh` passed

Refactored `RehydrationServices.js` so that store gets purged on `reducerVersion` change in `ReduxPersist.js`

#### BONUS 
Cool Reactotron logging
purge output
### Fixes #290 
alt tag

### DIFF
--- a/ignite-base/App/Services/RehydrationServices.js
+++ b/ignite-base/App/Services/RehydrationServices.js
@@ -1,27 +1,26 @@
 import ReduxPersist from '../Config/ReduxPersist'
 import { AsyncStorage } from 'react-native'
 import { persistStore } from 'redux-persist'
+import Reactotron from 'reactotron-react-native'
 
 const updateReducers = (store) => {
   const reducerVersion = ReduxPersist.reducerVersion
   const config = ReduxPersist.storeConfig
 
-  // Begin a fresh store
-  persistStore(store, config)
-
   // Check to ensure latest reducer version
   AsyncStorage.getItem('reducerVersion').then((localVersion) => {
     if (localVersion !== reducerVersion) {
-      // console.log('PURGING STORE', localVersion, 'vs.', reducerVersion)
-      // Purge store and refresh
-      persistStore(store, config, () => {
-        // Start a fresh store
-        persistStore(store, config)
-      }).purgeAll()
-      // Update reducer to current version number
+      Reactotron.log(`PURGING STORE ${localVersion} 'vs.' ${reducerVersion}`)
+      // Purge store
+      persistStore(store, config).purge()
       AsyncStorage.setItem('reducerVersion', reducerVersion)
+    } else {
+      persistStore(store, config)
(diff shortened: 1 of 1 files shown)


ignite-base/App/Services/RehydrationServices.js
```diff
@@ -1,16 +1,23 @@
...
   const config = ReduxPersist.storeConfig
 
+
   // Check to ensure latest reducer version
   AsyncStorage.getItem('reducerVersion').then((localVersion) => {
     if (localVersion !== reducerVersion) {
-      Reactotron.log(`PURGING STORE ${localVersion} 'vs.' ${reducerVersion}`)
+      console.tron.display({
+        name: 'PURGE',
+        value: `Reducer Version Change
```
skellock: Nothing wrong with what you've done here at all.   Props for doing this!

Another way would be to set `value` as an object.  That would have a nice table layout if you did.

ignite-base/App/Services/RehydrationServices.js
```diff
@@ -1,16 +1,23 @@
...
+        Old Version: ${localVersion}
+        New Version: ${reducerVersion}`,
+        preview: 'Purging Store',
```
skellock: Maybe even chuck the version numbers in here, so the user doesn't even need to expand the row?  Totally your call.  I wouldn't have even thought to do this.

ignite-base/App/Services/RehydrationServices.js
```diff
@@ -1,16 +1,23 @@
...
+        important: true
```
skellock: Damn right it is!

skellock:
Does this fix the async issue that people are reporting?

fvonhoven:
@skellock 
new purge output

Issue #362, opened 2016-09-10
Get "Process 'command '/usr/lib/jvm/java-8-oracle/bin/java'' finished with non-zero exit value 2" when building Ignite app 


make2514:
Hi guys,

I am very new to Ignite. I have followed the instruction in README.md file to initialize ignite app.
After executing this command: "react-native run-android", I got this failure notification from react-native:

Execution failed for task ':app:dexDebug'.
> com.android.ide.common.process.ProcessException: org.gradle.process.internal.ExecException: Process 'command '/usr/lib/jvm/java-8-oracle/bin/java'' finished with non-zero exit value 2

I have searched for a solution on stackoverflow and edit build.gradle as answered in this thread:
http://stackoverflow.com/questions/29460443/execution-failed-for-task-appdexdebug-com-android-ide-common-process-proces

The above failure went away, but there is this new failure notification:

Execution failed for task ':app:packageAllDebugClassesForMultiDex'.
> java.util.zip.ZipException: duplicate entry: org/webkit/android_jsc/R.class

Any insight would be appreciated! Thank you very much!

GantMan:
please `cd android` and run `./gradlew clean`  Please let me know if that fixes it.  I'm interested in this issue, as this sometimes happens to me.

make2514:
Thank you very much for your help, GantMan! The command `./gradlew clean ` indeed fixed the issue for me.
I have initialized a new Ignite app instance, follow your instruction and it worked.
Have a nice weekend!

GantMan:
@make2514  and @tnikolay - In version 1.7.1 which I will release soon, this will have been run for you.   Ensuring no one has to worry to do it themselves :)

TimFletcher:
I used 1.7.1 (having never installed Ignite before) and bumped into the #371 issue. I needed to run `./gradlew clean` manually to get `react-native run-android` to work.

GantMan:
that's so odd, since as of 1.7.1 it runs `./gradlew clean` for you after everything is installed

I'm gonna try to recreate this problem locally.

GantMan:
just did a fresh install of 1.7.1 and it didn't have this problem :(  Anyone got ideas?

xcsrz:
(bump)  With a fresh install of Android Studio, SDK packages and Ignite I ignited a new project and ran into this same issue right out of the box.

GantMan:
I'm frustrated by the need for `./gradlew clean`, I'll continue trying to track this down.

[...]

PR #361, opened 2016-09-09
Upgrade/rn0.33.0

GantMan:
## Please verify the following:
- [x] Everything works on iOS/Android
- [x] `ignite-base` **ava** tests pass
- [x] `fireDrill.sh` passed

## Describe your PR
All upgrades to bring Ignite to 0.33.0 support

PR #364, opened 2016-09-11
gradlew clean android now

GantMan:
- [x] `fireDrill.sh` passed

## Describe your PR
As of #361 we started getting a strange error on Android.  As reported here in #362 

This PR now automatically does some buttwiping for React Native Android so the dex files don't complain.  Not sure if this symptom should be be further investigated, but this fixes it for now.   Someone like @darinwilson  or @skellock might know why this was needed?

If not, the screw it, this PR does all the fixing necessary.


GantMan:
### proof it works now
image

PR #383, opened 2016-09-20
Create a test when a new component is generated

juddey:
- [X] Everything works on iOS/Android >> Generator Change
- [X] `fireDrill.sh` passed

## Describe your PR
After a bit of gitter chat this PR starts the ball rolling with the idea that we should scaffold out a test when we generate a component/saga/whatever. This PR adds the component test.

I've copied the `context.fs.copyTpl( ...` lines in `index.es` but since there are two similar functions in the same place I'm wondering if another function might be the best. 

🔚

### DIFF
--- a/ignite-generator/component/templates/component-test.js.template
+++ b/ignite-generator/component/templates/component-test.js.template
@@ -9,7 +9,7 @@ test('component exists', t => {
 })
 
 test('component structure', t => {
-  const wrapper = shallow(<<%= name %>  />)  //Add in Props if needed
+  const wrapper = shallow(<<%= name %>  />)  // Add in Props if needed
 
   t.is(wrapper.name(), 'View')                       // the right root component
   t.is(wrapper.children().length, 1)                 // has 1 child
@@ -18,7 +18,7 @@ test('component structure', t => {
 })
 
 //test('some other things here', t => {
-//   const wrapper = shallow(<YourComponentNameHere //SomeProps )
+//   const wrapper = shallow(<YourComponentNameHere // SomeProps > )
 //   // make some assertions.
 //
 //})
--- a/ignite-generator/component/templates/component-test.js.template
+++ b/ignite-generator/component/templates/component-test.js.template
@@ -3,22 +3,21 @@ import React from 'react'
 import { shallow } from 'enzyme'
 import <%= name %> from '../../App/Components/<%= name %>'
 
+const wrapper = shallow(<<%= name %> />)
+
 test('component exists', t => {
-  const wrapper = shallow(<<%= name %>  />)
   t.is(wrapper.length, 1) // exists
 })
 
 test('component structure', t => {
-  const wrapper = shallow(<<%= name %>  />)  // Add in Props if needed
-
-  t.is(wrapper.name(), 'View')                       // the right root component
-  t.is(wrapper.children().length, 1)                 // has 1 child
-  t.is(wrapper.children().first().name(), 'Text')    // that child is Text
+  t.is(wrapper.name(), 'View') // the right root component
+  t.is(wrapper.children().length, 1) // has 1 child
+  t.is(wrapper.children().first().name(), 'Text') // that child is Text
   t.true(wrapper.children().first().containsMatchingElement('<%= name %> Component')) // That the Component Text is included
 })
 
-//test('some other things here', t => {
-//   const wrapper = shallow(<YourComponentNameHere // SomeProps > )
-//   // make some assertions.
-//
-//})
+// test('some other things here', t => {
+// const wrapper = shallow(<YourComponentNameHere // SomeProps > )
+   // You can add in props as shown above, or use the constant wrapper declared
+   // at the top of the file.
+// })
--- a/ignite-generator/component/templates/component-test.js.template
+++ b/ignite-generator/component/templates/component-test.js.template
@@ -1,0 +1,24 @@
+import test from 'ava'
+import React from 'react'
+import { shallow } from 'enzyme'
+import <%= name %> from '../../App/Components/<%= name %>'
+
+test('component exists', t => {
+  const wrapper = shallow(<<%= name %>  />)
+  t.is(wrapper.length, 1) // exists
+})
+
+test('component structure', t => {
+  const wrapper = shallow(<<%= name %>  />)  //Add in Props if needed
+
+  t.is(wrapper.name(), 'View')                       // the right root component
+  t.is(wrapper.children().length, 1)                 // has 1 child
+  t.is(wrapper.children().first().name(), 'Text')    // that child is Text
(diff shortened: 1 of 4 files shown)


ignite-generator/component/templates/component-test.js.template
```diff
@@ -0,0 +1,24 @@
+import test from 'ava'
+import React from 'react'
+import { shallow } from 'enzyme'
+import <%= name %> from '../../App/Components/<%= name %>'
+
+test('component exists', t => {
+  const wrapper = shallow(<<%= name %>  />)
```
GantMan: one thing I've started doing is moving these `const`s to above all functions, so I don't have to repeat `wrapper` often.

ignite-generator/component/templates/component-test.js.template
```diff
@@ -0,0 +1,24 @@
...
+  t.is(wrapper.length, 1) // exists
+})
+
+test('component structure', t => {
+  const wrapper = shallow(<<%= name %>  />)  //Add in Props if needed
```
GantMan: Please add a space after `//` standard will lint this.

ignite-generator/component/templates/component-test.js.template
```diff
@@ -0,0 +1,24 @@
...
+
+  t.is(wrapper.name(), 'View')                       // the right root component
+  t.is(wrapper.children().length, 1)                 // has 1 child
+  t.is(wrapper.children().first().name(), 'Text')    // that child is Text
+  t.true(wrapper.children().first().containsMatchingElement('<%= name %> Component')) // That the Component Text is included
+})
+
+//test('some other things here', t => {
+//   const wrapper = shallow(<YourComponentNameHere //SomeProps )
```
GantMan: this one looks like it's missing the closing tag of `YourComponentNameHere`

GantMan:
@juddey  -  I'm thinking this is a GREAT candidate for our first generator to read the `.ignite` file!

### Consider this:
We now have a `.ignite` file.  It looks like so:
```js
{
  "igniteVersion": "1.8.0",
  "reactNativeVersion": "0.33.0",
  "options": {},
  "plugins": {}
}
```

What about, by default [the `.ignite` file](https://github.com/infinitered/ignite/blob/master/ignite-base/.ignite.template) looks like this:
```js
{
  "igniteVersion": "1.8.0",
  "reactNativeVersion": "0.33.0",
  "options": {
    "testing": "ava"
  },
  "plugins": {}
}
```

And the generator reads the `.ignite` file.  And if testing is `ava` it will generate the tests.  If it is NOT, then it won't create the test files.  This allows us to possibly support other types of tests in the future.  Now here at IR, we are happy with AVA, but as we move towards 2.0 we may allow others to support modules that work with jest or mocha, or some future thing. 

I think this is where generators need to go, and this is the first opinionated part.  They need to start using this file to make decisions on generating templates, this will be key for Ignite 2.0 on the roadmap.

Does that sound like something you would be interested in adding to this PR?  :fire:

ignite-generator/component/templates/component-test.js.template
juddey: OK, but how would that work with multiple props passed in (or not) to the component?
eg. I pass in one prop to the component, it renders text of `miss piggy`, I pass in another prop and it renders a picture of kermit 🐸 ... Wouldn't you need two tests to cover that scenario and two wrapper declarations (for the different props).

I might be missing the point... 🎯

ignite-generator/component/templates/component-test.js.template
juddey: Oh yeah!

juddey:
Hey @GantMan 

Re: Reading the ignite file. Sure thing, we can add it to this PR. I'll need some guidance about how best to implement it though, perhaps a starter on what files you're thinking might be changed and how you see the conditional logic being applied. Don't want to over (or under) complicate matters. 

Over to you.

ignite-generator/component/templates/component-test.js.template
GantMan: All good!  It's more picky than anything.  I usually make a vanilla wrapper up top.  This is for testing basics of the component.  Should I want to test differences I'll make a second const inside each test.   But in some cases 3 tests all use the same wrapper.   Might just be me, I'm flexible on this.

GantMan:
Hey @juddey - I felt my request might have left you a bit in the dark, so I PRed your PR! hah.

Let me know if that looks reasonable.   This is the dawn of generators reading the `.ignite` file!  Also means someone could add mocha/jest test generators... though I have zero plans to do so myself.

ignite-generator/component/templates/component-test.js.template
juddey: aah, yes, got you now. Its clear now that you've explained it. I'll annotate the template to make it clear that this is what is happening.

juddey:
Hey @GantMan 
> Let me know if that looks reasonable. 

This sounds like there should be something for me to look at - am I missing something?

GantMan:
@juddey  ->  https://github.com/juddey/ignite/pull/1

Told you I PRed your PR 😆

juddey:
oh gotcha now 😊

juddey:
Oh - sorry 😊 ! I didn't think to look in my ***own*** repo. Possibly because it's Friday (well here anyway)

Will check it out.

GantMan:
Yup!  I know that was tricky, but when you asked for clarification and I couldn't do it with words, I was like.... I'm gonna do something crazy, lol.

juddey:
You know what's real tricky? I'm in the future 🚀  😆

GantMan:
@fvonhoven - I know you hate tests, this change will give you control to write terrible code and no tests to catch them!  😆 

@juddey - thanks for the merge!  Let me know when you do that change to the `wrapper` et al. and then this will be good to go!

juddey:
Hey @GantMan 

I'm not sure whether the code you sent on the juddey PR was meant to work straight off, but  after doing some testing I found that the code wasn't quite functioning - I couldn't create a new Component Test even if I had the ignite file in place with the correct options.

I made the following changes:

1. Moved the `igniteConfig` constant inside the `copyOverComponent` function
2. Changed the file name from `.ignite` to `ignite.json` - I know node is **meant** to read js, json and unknown files in the same, but for some reason the script wasn't playing ⚽ .

and all was golden.

Also, I've added the constant explanation and tidied up a few things in the templates which caused the linter to error. 

Once these changes are merged, I'll add some documentation in a separate PR.

GantMan:
Ah, thanks for following up.  I thought I tested it.   I'm gonna merge this beast, because it's awesome! and then see if I can run a few tests :)

Fantastic PR sir!

juddey:
Sweet as! 🙌 

Good to see we have a model for doing this sort of stuff now. Means we can go ahead with some of those other generators too.

PR #388, opened 2016-09-23
fix html coverage report opening on Linux

brad:
- [X] `fireDrill.sh` passed

## Describe your PR
Currently the HTML coverage reports don't open on Linux because there is no `open` executable. This PR fixes that by falling back to use `xdg-open` if the `open` command fails.


GantMan:
thumbs

PR #400, opened 2016-10-03
Make sure you can't ignite a project named `React`

### LINKED ISSUE #399: Make sure you can't ignite a project named `React
As you can see, React will fail if you chose the name `React`
https://github.com/facebook/react-native/blob/f22e9353a7fd18eb88218000d0debecae3140d19/react-native-cli/index.js#L129-L137

We should endeavor to catch this error and alert the user in our CLI checks.

rajington:
## Make sure you can't ignite a project named `React`
Fixes #399


GantMan:
deftly done

PR #403, opened 2016-10-03
Master check against FB - For early warning checks

GantMan:
### Early testing script
If you'd like to test Ignite against FB Master branch
Run with 2nd param latest `./testRelease.sh fb_check latest`

This change goes all the way down through the CLI to the generator.


skellock:
There's some extra noise in here because of the merge (rebasing makes this cleaner).

I think I'm reading this right though, and it looks great and useful to me.

fvonhoven:
👍👍

PR #405, opened 2016-10-03
Remove ava global dependencies

guidiego:
- [ x ] Everything works on iOS/Android
- [ x ] `ignite-base` **ava** tests pass
- [ x ] `fireDrill.sh` passed

## Describe your PR
I have some problems with globals during the installation of the project, so I decided to remove globals dependencies from `package.json`, I start for `ava` but let me know if has something more that I can change :)


guidiego:
I was looking alternatives for `xdg-open`, something that I can install by npm :) I found that [opn-bg](https://www.npmjs.com/package/opn-bg). If you guys have some other ideias will be nice listen that :)

skellock:
Thanks for the `PR`.  `nyc` would also qualify for this treatment too.  Also, I'm onboard for `open-bg`.  Good find.

guidiego:
@skellock Yeah! I think the same, after do the onboarding give me a feed back :D

skellock:
So, hey, i tested this out.

Turns out I'm not sure any of this is needed.

These deps should be sourced from `node_modules/.bin`.   Both `nyc` and `ava` are devDeps.

I tested by `npm -g rm ava` and nyc.   

Are you running `npm` version 2 by any chance?

guidiego:
@skellock Yeah! I saw that in this moment, when I clone into my mac and run NYC/AVA without global installation. My fault, I use `c9.io` to write in open source projects, sometimes I forgot that instances old versions of node/npm. This will be only necessary if someone wants run the project with lowest node versions.

I think that we can only consider saw `xdg-open` :)

GantMan:
gonna close this one, in favor of whatever comes next.

PR #413, opened 2016-10-06
Fix Redux Persist Config whitelist.

### LINKED ISSUE #409: ReduxPersist config should not include empty whitelist
As several people on gitter mentioned, an empty whitelist seems to blacklist all reducers.
Specifying an empty whitelist effectively blacklists everything. This is something redux-persist should sort out, but maybe ignite should consider having only the whitelist or blacklist defined in ReduxPersist.js, since once you have a whitelist, adding to the blacklist is analogous to removing from the whitelist and vice versa. i.e it's redundant.

juddey:
- [ ] Everything works on iOS/Android(cli changes)
- [X] `fireDrill.sh` passed

## Describe your PR
Fixes #409

1/4 of a 👕  !


GantMan:
image

PR #465, opened 2016-10-29
Use shallow clone when downloading latest Ignite [Fixes #401]

### LINKED ISSUE #401: [CLI] Ignite can be sped up with a quick performance tweak
currently, Ignite CLI will pull down a git version of itself on running, and use the branches/tags/files etc.  

As Ignite grows the `git clone` will grow, and the time associated.   We can narrow this process with prime precision by pulling only a `head` and `branch` of the desired version.  Thus ensuring Ignite CLI runs will only await the files they are required to use.

Example of pulling a specific branch with no history:  https://strk.kbt.io/blog/2011/06/07/getting-just-the-tip-of-a-remote-git-branch/comment-page-1/

patricksmith:
- [ ] `ignite-base` **ava** tests pass
- [ ] `fireDrill.sh` passed

## Describe your PR

Addresses #401. Uses `--depth 1` and specifies requested branch when cloning Ignite.

The branch is printed in the step description:

```
$ ignite new test --branch 1.9.1
🔥 Setting test on FIRE! 🔥
... snip ...
✔︎ Running React Native setup version 0.35.0 (~ 2 minutes-ish)
✔︎ Downloading latest Ignite files from https://github.com/infinitered/ignite.git#1.9.1
```

One functional change is that previously if an invalid branch name was specified, the clone would succeed, the checkout would fail, and `ignite new` would continue using either master or the latest tag. Now, however, the clone will fail and `ignite new` will bail out:

```
$ ignite new test --branch foobar
🔥 Setting test on FIRE! 🔥
... snip ...
✔︎ Getting ready for guests
✔︎ Running React Native setup version 0.35.0 (~ 2 minutes-ish)
𝗫 Downloading latest Ignite files from https://github.com/infinitered/ignite.git#foobar
Failed to clone https://github.com/infinitered/ignite.git with branch foobar
```

### DIFF
--- a/ignite-generator/src/app/index.es
+++ b/ignite-generator/src/app/index.es
@@ -278,7 +278,11 @@ export class AppGenerator extends Generators.Base {
     this.spawnCommand(command, commandOpts, {stdio: 'ignore'})
       .on('close', (retCode) => {
         animation.finish(retCode)
-        done()
+        if (retCode === 0) {
+          done()
+        } else {
+          this._logAndExit(`Failed to clone ${repo} with branch ${branch}`)
+        }
       })
   }
 
--- a/ignite-generator/src/app/index.es
+++ b/ignite-generator/src/app/index.es
@@ -240,6 +240,23 @@ export class AppGenerator extends Generators.Base {
   }
 
   /**
+   * Get the git branch or tag that we should check out.
+   */
+  _getGitBranch () {
+    const tag = this.options['tag'] || lockedIgniteVersion
+    // read the user's choice from the source-branch command line option
+    const branch = this.options['branch']
+    // check if the user specified a branch
+    const isBranchUndefined = typeof branch === 'undefined' || branch === null || branch === 'master' || branch === ''
+    // return branch if its defined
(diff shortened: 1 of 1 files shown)


ignite-generator/src/app/index.es
```diff
@@ -251,61 +268,21 @@ export class AppGenerator extends Generators.Base {
...
     const repo = useCustomRepo ? requestedRepo : defaultRepo
+    const branch = this._getGitBranch()
     // start spinner and message
-    const animation = Utilities.startStep(`Downloading latest Ignite files from ${repo}`, this)
+    const animation = Utilities.startStep(`Downloading latest Ignite files from ${repo}#${branch}`, this)
 
     const done = this.async()
     const command = 'git'
-    const commandOpts = ['clone', repo, this.sourceRoot()]
+    const commandOpts = ['clone', '--depth', '1', '--branch', branch, repo, this.sourceRoot()]
```
GantMan: This is awesome, so this will take care of tags and branches from what I'm reading!

PR #367, opened 2016-09-13
Added early exit if generators are being run in an ignite directory.

juddey:
## Please verify the following:
- [X] Everything works on iOS/Android (update to cli)
- [X] `ignite-base` **ava** tests pass
- [X] `fireDrill.sh` passed

## Describe your PR
* Early exit when generators are not being run in an ignite directory
* Also added an early exit if the new command is being run in an ignite directory.

Verification Tests:

screen shot 2016-09-13 at 12 46 44 pm

Let me know if there's anything else :)

PR #466, opened 2016-10-30
Moar Generators!

juddey:
## image

Way back in #367 I got the idea I should refactor to reduce the code repetition we were doing when copying the template files. 💥  - that's included in 67cce4a. You can now define a config object in the class constructor which contains parameters for the files to be copied. It looks a little bit like:

```
this.fileConfig = {
  templatePath: 'component.js.template',
  destinationPath: 'Components',
  style: true,
  test: true }
} 
```

This config object will cause:
- the app template to be copied to `./App/Components/`
- the style template to be copied to `./App/Components/Styles`
- a test to be generated to:
  - `./Tests/Components/` if the `.ignite` file specifies `ava`
  -  `./__tests__/Components/` if the `.ignite` file specifies `jest` (so forward looking 🚀  )

Both style and test parameters are optional.

This tidies up the generator classes and makes them more readable too imho. Read on for more!
---
### Now, the main event.

image

Well, it might not be that main any more, because I've let the cat out of the bag in the preview really.

I've added generators for redux and sagas using the new function I prattled on about above. There is support for ava and jest tests in there also. 

All the tests pass when you generate the files in the first instance. (I cannot be held accountable for changes post file generation though 😃  ) 

I thought about, but decided against, adding the files to the relevant index.js - ime this is mucky to implement and might not be what the dev wants anyway, so ends up as a high risk low return exercise.
---
### Final words

Theres a bit more code in here than I've submitted previously so I do expect some review 📝 !
And....

image

### DIFF
--- a/ignite-generator/.gitignore
+++ b/ignite-generator/.gitignore
@@ -44,5 +44,6 @@ listview/index.js
 mapview/index.js
 imagesfile/index.js
 redux/index.js
+saga/index.js
 screen/index.js
 testing/index.js
--- a/ignite-generator/.gitignore
+++ b/ignite-generator/.gitignore
@@ -43,5 +43,6 @@ container/index.js
 listview/index.js
 mapview/index.js
 imagesfile/index.js
+redux/index.js
 screen/index.js
 testing/index.js
--- a/ignite-generator/src/saga/index.es
+++ b/ignite-generator/src/saga/index.es
@@ -1,0 +1,27 @@
+#! /usr/bin/env node
+'use strict'
+import Generators from 'yeoman-generator'
+import * as Utilities from '../utilities'
+
+class SagaGenerator extends Generators.Base {
+
+  constructor (args, options) {
+    super(args, options)
+    this.argument('name', { type: String, required: true })
+    this.fileConfig = {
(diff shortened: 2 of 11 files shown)


ignite-generator/src/saga/index.es
```diff
@@ -0,0 +1,27 @@
...
+      test: true }
+  }
+
+  generateApp () {
+    Utilities.createFiles(this, this.fileConfig)
+  }
+
+  end () {
+
+    console.log('Time to get cooking! 🍽 ')
```
GantMan: Hah, we might need to abstract this to utilities, we say it so often :D 

🍽

ignite-generator/.gitignore
```diff
@@ -43,5 +43,7 @@ container/index.js
 listview/index.js
 mapview/index.js
 imagesfile/index.js
+redux/index.js
+saga/index.js
```
GantMan: w00h00!

GantMan:
Hearing no issues, I'm going to merge this!  Great work @juddey

GantMan:
@juddey - can you please document in the readme?

PR #473, opened 2016-11-03
Updates Reactotron to 1.3.x and simplifies the configuration. 💃

skellock:
Highlights:

* brings us up to 1.3.x of Reactotron
* we can now use snapshots
* moves redux integration out of `CreateStore.js` and into `ReactotronConfig.js` where it belongs
* simplifies redux integration by calling `Reactotron.createStore()` in dev mode
* clears Reactotron on each app startup


GantMan:
time

PR #481, opened 2016-11-04
fix ThemeScreen name

ruddell:
- [x] `fireDrill.sh` passed

I couldn't get fireDrill.sh to run at first, confusing error: `Wrong version of Flow. The config specifies version ^0.33.0 but this is version 0.34.0`. I added flow-bin@0.33.0 in fireDrill.sh and it worked

## Describe your PR
Very small change, the name of the class in `ThemeScreen.js` class was 'UsageExamplesScreen'

Loving Ignite so far, thanks!


GantMan:
good catch!

Issue #485, opened 2016-11-08
Update Ignite to RN 0.37.0


GantMan:
https://github.com/facebook/react-native/releases/tag/v0.37.0

Issue #484, opened 2016-11-08
Linting crapping out right away


GantMan:
latest react native throws in jest tests that fail linting:

![image](https://cloud.githubusercontent.com/assets/997157/20088853/c3767860-a537-11e6-8d56-861f267f0b7c.png)

As for now, we can just delete this folder and enforce AVA.  Eventually this will get cleaned up if the user selects Jest over AVA

GantMan:
Also, add `standard` to package.json.

baskan:
+ also add "seamless-immutable" to package.json

PR #489, opened 2016-11-10
Upgrade to RN 0.37

ryanlntn:
- [x] Everything works on iOS
- [ ]  Everything works on Android
- [x] `fireDrill.sh` passed

## Describe your PR
Addresses https://github.com/infinitered/ignite/issues/485

I wasn't able to check Android because maps failed due to not finding the right version of play services. I think that might just be an issue with the SDK version I have installed though. 

Also it looks like a bunch of the ignores I added to .flowconfig were removed so flow kind of blows up when you run fireDrill.sh. It still passes though ¯\_(ツ)_/¯


GantMan:
@ryanlntn - Can you please update/verify Android?

Also, let's make fireDrill fail on flow fail.  It makes sense right?   Who removed your ignores?

ryanlntn:
Apparently I did lol. I'll get this finished up today.

ryanlntn:
Interesting. Flow is passing for me locally but getting loads of required module not found errors on Semaphore. Not sure why that would be...

Also this is super annoying: https://github.com/facebook/flow/issues/869

I really wish we could just ignore all errors with node_modules.

ryanlntn:
@GantMan Still not able to run this on Android. Looks like maps is broken again: `Could not find com.google.android.gms:play-services-base:9.4.0.`

GantMan:
@ryanlntn - you can shell into semaphore, which is cool.   That way you can have a socket5 webterm to the actual instance, and play around on the actual box.

Also, that's crazy about the 9.4.0 stuff.  But especially since I added code that undoes 9.4.0 locking.  Perhaps we can pair sometime and identify the issue so we will be down to document/fix for someone who hits this problem.

robbue:
We are depended on react-native-router-flux to update using the NavigationExperimental to leverage the new changes provided in RN 0.37?

GantMan:
@robbue - I'm not sure.  I haven't tried yet.  Quite likely, yes.  Though you can always wire that up yourself and not wait.

ryanlntn:
@GantMan I've resolved the issues with fireDrill.sh. Looks like node modules weren't actually getting installed before flow check on Semaphore. The only issue now is that I still get build failures on Android due to react-native-maps looking play services to 9.4.0. The code you have for overriding that doesn't appear to work.

GantMan:
did you also take care of this one?   https://github.com/infinitered/ignite/issues/484

If not, I will take a look.

GantMan:
@ryanlntn - Looks great bud!

I did some library upgrades too.  This PR is ready as soon as it passes CI

PR #510, opened 2016-11-27
Remove underscore prefixing for “private” methods in React components

### LINKED ISSUE #509: Usage of underscore for internal methods in React components
It seems many people are abandoning the use of prefixing "internal" methods of a React component with an underscore (eg: `_handleOnPress`). The [Airbnb React's style guide](https://github.com/airbnb/javascript/tree/master/react#methods) explicitly discourages this:

> Do not use underscore prefix for internal methods of a React component.
>> Why? Underscore prefixes are sometimes used as a convention in other languages to denote privacy. But, unlike those languages, there is no native support for privacy in JavaScript, everything is public. Regardless of your intentions, adding underscore prefixes to your properties does not actually make them private, and any property (underscore-prefixed or not) should be treated as being public. See issues [#1024](https://github.com/airbnb/javascript/issues/1024), and [#490](https://github.com/airbnb/javascript/issues/490) for a more in-depth discussion.

Right now in Ignite the use of underscore to prefix "private" or "internal" methods is not consistent. 

For example, [DrawerContent](https://github.com/infinitered/ignite/blob/master/ignite-base/App/Containers/DrawerContent.js#L26) does not use underscore:

```javascript
handlePressComponents = () => {
    this.toggleDrawer()
    NavigationActions.componentExamples()
  }
```

While [ListviewExample](https://github.com/infinitered/ignite/blob/master/ignite-base/App/Containers/ListviewExample.js#L74) does use underscore prefixing:

```javascript
  _renderRow (rowData) {
    return (
      <View style={styles.row}>
        <Text style={styles.boldLabel}>{rowData.title}</Text>
        <Text style={styles.label}>{rowData.description}</Text>
      </View>
    )
  }
```

Personally I'd prefer removing underscore, as the Airbnb style guide suggests. If you guys are ok with this I could submit a PR.

To avoid future problems for this, maybe Ignite should subscribe or recommend some style guide.

javiercr:
Fixes #509 

I only have removed underscore prefixing for React components and its generators (templates). Let me know if I missed any.

### DIFF
--- a/ignite-base/App/Containers/ListviewExample.js
+++ b/ignite-base/App/Containers/ListviewExample.js
@@ -65,13 +65,13 @@ class ListviewExample extends React.Component {
 
   /* ***********************************************************
   * STEP 3
-  * `_renderRow` function -How each cell/row should be rendered
+  * `renderRow` function -How each cell/row should be rendered
   * It's our best practice to place a single component here:
   *
   * e.g.
     return <MyCustomCell title={rowData.title} description={rowData.description} />
   *************************************************************/
-  _renderRow (rowData) {
+  renderRow (rowData) {
     return (
       <View style={styles.row}>
         <Text style={styles.boldLabel}>{rowData.title}</Text>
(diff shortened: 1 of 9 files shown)


javiercr:
Uhm, not sure what to do about the tests. It seems the build for master is failing too:
https://travis-ci.org/infinitered/ignite/builds/179213380

GantMan:
fixed master, can you merge master into your branch so CI can pass?

PR #511, opened 2016-11-27
Move Reactotron switch to DebugSettings.js

javiercr:
While Reactotron is very helpful, it often affects the performance of the app in DEV mode to a point where it gets frustrating (mainly on navigation animations).

These changes allows to quickly disable Reactotron in DEV mode by just editting `DebugSettings.js` in the same fashion that we can disable Redux logging.

### DIFF
--- a/ignite-base/App/Config/ReactotronConfig.js
+++ b/ignite-base/App/Config/ReactotronConfig.js
@@ -1,12 +1,14 @@
 import { StartupTypes } from '../Redux/StartupRedux'
+import Config from '../Config/DebugSettings'
 import Immutable from 'seamless-immutable'
 const Reactotron = require('reactotron-react-native').default
 const errorPlugin = require('reactotron-react-native').trackGlobalErrors
 const apisaucePlugin = require('reactotron-apisauce')
 const { reactotronRedux } = require('reactotron-redux')
 const sagaPlugin = require('reactotron-redux-saga')
+const USE_REACTOTRON = Config.reactotron
 
-if (__DEV__) {
+if (USE_REACTOTRON) {
   Reactotron
     .configure({
       // host: '10.0.3.2' // default is localhost (on android don't forget to `adb reverse tcp:9090 tcp:9090`)
--- a/ignite-base/App/Config/ReactotronConfig.js
+++ b/ignite-base/App/Config/ReactotronConfig.js
@@ -6,9 +6,8 @@ const errorPlugin = require('reactotron-react-native').trackGlobalErrors
 const apisaucePlugin = require('reactotron-apisauce')
 const { reactotronRedux } = require('reactotron-redux')
 const sagaPlugin = require('reactotron-redux-saga')
-const USE_REACTOTRON = Config.reactotron
 
-if (USE_REACTOTRON) {
+if (Config.useReactotron) {
   Reactotron
     .configure({
       // host: '10.0.3.2' // default is localhost (on android don't forget to `adb reverse tcp:9090 tcp:9090`)
--- a/ignite-base/App/Redux/CreateStore.js
+++ b/ignite-base/App/Redux/CreateStore.js
@@ -45,8 +45,9 @@ export default (rootReducer, rootSaga) => {
     enhancers.push(autoRehydrate())
   }
 
-  // in dev mode, we'll create the store through Reactotron
-  const createAppropriateStore = __DEV__ ? console.tron.createStore : createStore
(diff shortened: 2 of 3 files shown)


ignite-base/App/Redux/CreateStore.js
```diff
@@ -45,8 +45,9 @@ export default (rootReducer, rootSaga) => {
     enhancers.push(autoRehydrate())
   }
 
-  // in dev mode, we'll create the store through Reactotron
-  const createAppropriateStore = __DEV__ ? console.tron.createStore : createStore
+  // if Reactotron is enabled (default for __DEV__), we'll create the store through Reactotron
+  const USE_REACTOTRON = Config.reactotron
+  const createAppropriateStore = USE_REACTOTRON ? console.tron.createStore : createStore
```
GantMan: why make a const if it's only going to be used this once?  maybe just use `Config.reactotron` here?  I don't think it adds readability

GantMan:
I like the idea of moving reactotron switch to config.  @skellock thoughts?

ignite-base/App/Config/ReactotronConfig.js
```diff
@@ -1,12 +1,14 @@
...
 import Immutable from 'seamless-immutable'
 const Reactotron = require('reactotron-react-native').default
 const errorPlugin = require('reactotron-react-native').trackGlobalErrors
 const apisaucePlugin = require('reactotron-apisauce')
 const { reactotronRedux } = require('reactotron-redux')
 const sagaPlugin = require('reactotron-redux-saga')
+const USE_REACTOTRON = Config.reactotron
 
-if (__DEV__) {
+if (USE_REACTOTRON) {
```
GantMan: maybe naming it to `Config.reactotronActive` or `Config.useReactotron` would be good.  The const feels a bit much in this spot.

skellock:
Ya, that seems perfectly reasonable to me.

For a tool like this, i think opt-in would be a better approach.

What concerns me though, is why it is slow?  If you wouldn't mind @javiercr, i'd love to hear your experiences.  [Drop me a line](https://github.com/reactotron/reactotron/issues/new).

Did it just start in `reactotron-react-native@1.5.2`?  Adding in the circular reference detection & function identification tanked perf.

ignite-base/App/Config/ReactotronConfig.js
javiercr: Thanks for the feedback, just changed that :)

javiercr:
@skellock to be honest I'm still quite new to React Native and redux-saga, therefore the performance problems I'm seeing could be my fault for not doing things in the proper way.  

Being said that, with redux logging and Reactotron enabled, some of my navigation transitions (which trigger API calls using redux-saga) freeze up to 3 seconds. When I disable redux logging and Reactotron everything runs pretty much fine.

This is a bit off topic, but something that I find confusing regarding these performance problems is that redux-saga documentation states that a saga is like a *separate thread*, however AFAIK in React Native there is a single JS thread (am I wrong?) and console.log statements create big bottlenecks to this thread. From what I can see when I use redux-saga to do API calls in combination with redux logging / reactotron, that JS threat get stuck quite often, blocking the navigation animations.

skellock:
@javiercr: 

Yes, JS is single threaded, but evented.  Which means they can have multiple call stacks.

All the complexities of multi-threading with none of the benefits.

The suspect problem I think you're experiencing is not redux logging or even reactotron logging.  It's `remote debugging`.  Is that on?

If so, here's the short version:  Chrome throttles the resource availability of the process when the tab and Chrome parent process is not foreground.

Now I'm biased, so don't take my word for it, but... what I do is turn off *all* logging.  Then turn on Reactotron.  And exclusively use `console.tron.log()` or `display()` or `error()`.

If you are more of frame steppin'/breakpointin' kinda guy.  Have a look at [react-native-debugger](https://github.com/jhen0409/react-native-debugger).  It is basically the debugging experience of Chrome, with 100% less tears.

I still think this global switch in the PR is a good idea.

javiercr:
@skellock 
> JS. All the complexities of multi-threading with none of the benefits.
I think I'm gonna print a poster with that sentence lol.

Remote debugging does not effect performance that much in my case. With redux logging and reactotron disable, the app runs in an acceptable way in the iOS simulator while having Chrome inspector open with remote debugging.

I will definitely check react-native-debugger, thank you for the tip!.

And thanks for merging the PR. Keep making Ignite awesome! 🙌

PR #517, opened 2016-11-30
Improve TemperatureSagaTest

derekgreenberg:
- [x] `fireDrill.sh` passed


GantMan:
nice, thanks!

derekgreenberg:
Small improvement to the TemperatureSagaTest's 'success path' test - rather than hard-code the expected temp value that is passed to the success path test, get the calculated temperature from the convertFromKelvin transform for comparison in the t.deepEqual() step. We already have unit tests for the transform, so we should use it to generate the expected value that is passed to the reducer in this test. This ensures that changes to the FixtureApi will not break this test.

PR #522, opened 2016-12-02
Add OPEN_SCREEN action and saga

derekgreenberg:
- [x] `fireDrill.sh` passed

## Describe your PR

This adds an action ('OPEN_SCREEN') & saga to enable users to dispatch a command to open any screen. 

For Reactotron users, when generating a new app via Ignite, users can retain the useful screens generated by ignite-base, such as 'componentExamples', even if their new app bypasses any Ignite-generated screens on startup, simply by dispatching to Reactotron, like this:

```{type: 'OPEN_SCREEN', screen: 'componentExamples'}```

In addition, 'OPEN_SCREEN' is helpful while developing screens for an app that have not been set up with any nav links in the app. The developer can work on the new screen and test it via a direct dispatch from Reactotron.

The saga sets the react-native-router-flux option param to type: ActionConst.RESET by default, but users can easily override this by passing in different options. The unit test for the saga demonstrates this.

### DIFF
--- a/ignite-base/App/Sagas/OpenScreenSagas.js
+++ b/ignite-base/App/Sagas/OpenScreenSagas.js
@@ -1,0 +1,12 @@
+import { call } from 'redux-saga/effects'
+import { Actions as NavigationActions, ActionConst } from 'react-native-router-flux'
+import { merge } from 'ramda'
+
+// Process OPEN_SCREEN actions, intended for use with Reactotron to open a screen via a direct dispatch
+export function * openScreen (action) {
+  const {screen, options} = action
+  // Always reset the nav stack when opening a screen via Reactotron
+  // You can override the RESET type in the options passed to the OPEN_SCREEN dispatch
+  const mergedOptions = merge({type: ActionConst.RESET}, options)
+  yield call(NavigationActions[screen], mergedOptions)
+}
--- a/ignite-base/App/Sagas/OpenScreenSagas.js
+++ b/ignite-base/App/Sagas/OpenScreenSagas.js
@@ -2,10 +2,10 @@ import { call } from 'redux-saga/effects'
 import { Actions as NavigationActions, ActionConst } from 'react-native-router-flux'
 import { merge } from 'ramda'
 
-// Process OPEN_SCREEN actions, intended for use with Reactotron to open a screen via a direct dispatch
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: ignite-base/App/Sagas/OpenScreenSagas.js [hunk 1]
```diff
@@ -0,0 +1,12 @@
+import { call } from 'redux-saga/effects'
+import { Actions as NavigationActions, ActionConst } from 'react-native-router-flux'
+import { merge } from 'ramda'
+
+// Process OPEN_SCREEN actions
+export function * openScreen (action) {
+  const {screen, options} = action
```
skellock: Maybe a safety check?
```js
const { screen, options = {} } = action
```

ignite-base/App/Sagas/OpenScreenSagas.js
```diff
@@ -0,0 +1,12 @@
...
+  // Always reset the nav stack when opening a screen by default
+  // You can override the RESET type in the options passed to the OPEN_SCREEN dispatch
+  const mergedOptions = merge({type: ActionConst.RESET}, options)
```
skellock: Totally thinking out loud, but maybe straight up es6 for this?  I mean, personally, i love ramda, in this case, I think the language expresses this clearer.

I'm totally ok if you don't want to.

```js
const navigationOptions = { type: ActionConst.RESET, ...options }
```

ignite-base/Tests/Sagas/OpenScreenSagaTest.js
```diff
@@ -0,0 +1,21 @@
+import test from 'ava'
+import { call } from 'redux-saga/effects'
+import { openScreen } from '../../App/Sagas/OpenScreenSagas'
+import { Actions as NavigationActions, ActionConst } from 'react-native-router-flux'
+
+const stepper = (fn) => (mock) => fn.next(mock).value
+
+// Using 'myScreen' in the mocks of the arguments passed to the action for these tests is important.
```
skellock: 👍

derekgreenberg: I totally agree with @skellock's comments and have committed those changes.

GantMan:
LGTM! 👍

PR #539, opened 2016-12-14
Update react-native-maps and simplify MapViewExample 

javiercr:
- [x] `fireDrill.sh` passed

## Describe your PR

1. Updated react-native-maps from 0.11.0 to 0.12.2
2. Simplified `MapViewExample.js` by replacing `calculateRegion` with `fitToElements` method offered by the  `<MapView>` component.
3. Removed `MapHelpers.js` as it is not needed anymore.


GantMan:
I like this.  We need to **ask** the user if they want the fancy map example or not.  That way we can keep maps clear from the bells and whistles.  I might revisit this in a different light soon.  Will reference this ticket.

skellock:
Hey @javiercr , as you know, we're plugging away at Ignite 2.   We're going to be merge v2 into master shortly which will conflict with your commit.

RN 0.40 isn't compatible with 0.12.2 of maps (we need 0.13 for that).  

Soooo.... we'll be updating maps & making a `ignite-maps` plugin for that shortly!

Wanted to thank you for your time and fixes here!

When it comes time for the maps plugin (later this week), I'll refer to your changes again to make we don't leave this behind.

:1st_place_medal: my friend!

javiercr:
No worries! Makes sense, looking forward to v2. Let me know if there is anything I can help with.

PR #545, opened 2016-12-16
Template Registry

skellock:
Alrighty.

Now, people can create their own plugins for ignite.  

This is template registry.  In your app, people can specify which plugins drive the screens, components, and more.

There will be more ux around this via `ignite add`, but the engine as is... works.

image

A few side effects to address shortly once we discuss.

Should 3rd party generators show up when a type `ignite`?   I vote yes, but let's specify some restrictions.  We don't want them overriding things like `add` and `generate` themselves.

Also, how do we want to disambiguate between **multiple** plugins overriding the same template? For example `ignite generate screen`.   My vote is to prompt with a list of all options and tell the user how to permanently choose in the `ignite.toml`.

Pretty excited to see this coming together!


skellock:
This PR was to add the ability for 3rd parties to add their own generators.

GantMan:
understood, I have a few questions of course.   for instance the TOML entry is a bit confusing.  Stub?

GantMan:
Does this mean I can/should start 3rd party plugins?  🎉 ?

skellock:
Ready for liftoff.

giphy

PR #550, opened 2016-12-18
exclude .env from generated gitignore

imjared:
- [x] `fireDrill.sh` passed

## Describe your PR
Not really changing much here (I think) but when I went to update an environment variable, I noticed that the file was tracked by git. This should keep your secrets safe and sound.

### DIFF
--- a/ignite-generator/.gitignore
+++ b/ignite-generator/.gitignore
@@ -47,3 +47,7 @@ redux/index.js
 saga/index.js
 screen/index.js
 testing/index.js
+
+# Secrets
+#
+.env


GantMan:
hrmmm, you have a .env in the generators?

not sure this is the right place.  Shouldn't this catch/fix the .env file?

https://github.com/infinitered/ignite/blob/master/ignite-base/.gitignore#L49-L51

imjared:
I should have included steps to reproduce the issue.

Steps using ignite v1.12.0

1. `$ ignite new myapp`
1. `$ cd myapp`
1. `$ git status`

Expected behavior:
1. `.env` should be ignored

Actual behavior:
1. `.env` is untracked but not ignored.
image

The generated `.gitignore` in the root of project `myapp` does not include a line for `.env`. I can add `.env` to `.gitignore`, run `$ git-status` a second time, and I'll see that `.env` is not listed in untracked files.

image

skellock:
Summoning @GantMan for a quick gut check here.

jamonholmgren:
Merging this, because I agree this should be done.

nonameolsson:
What's the status on this one? Just encountered the same issue when I started using `react-native-config`.

jamonholmgren:
@nonameolsson I don't think we're using `react-native-config` anymore, which is why we .gitignored the .env file.

nonameolsson:
@jamonholmgren Oh? But it comes when I generate a project, and the Readme mentions it.

jamonholmgren:
My bad, that's what I get for responding in a hurry. You're right that we are indeed using `react-native-config` in the template.

Can you explain a little further what the issue is that you're running into?

nonameolsson:
The readme states:
`The .env file is ignored by git keeping those secrets out of your repo.`.

But the `.gitignore` doesn't contain anything that has to do with `.env`. I had to manually add it to it.

The problem: The readme says that the `.gitignore` will ignore `.env` but it doesn't do thay. 

Solution 1): Remove that line from the Readme
Solution 2): Make sure that git ignores the `.env`-file. 

I would go for suggestion 2, because things added in it probably wouldn't be checked into the repo. In my project I have created a `.env.example` and added the keys needed, but without its value. Then when someone clones the project they need to rename it to `.env`, and populate it the API-keys, and other things.

jamonholmgren:
@GantMan Can you weigh in here?

GantMan:
@nonameolsson - that's awesome, would you consider a PR to have the `.env.example` by default in a newly ignited project?

I'll be happy to do so if you're not comfortable, but this is a good opportunity to contribute.

nonameolsson:
@gantman Sure! I'll create a PR! It will take some days, because my internet connection is down. But as soon as it works again I'll try to make one.

PR #498, opened 2016-11-21
Screen generators via Plugin

GantMan:
### Screen generators moved on to external generator.

This moves all screen generators (except maps) to use [this](https://github.com/GantMan/ignite-basic-generators) instead.

When you ignite new screens/components etc, you're using the plugin system.  This allows composition and extendability.  This begins the plugin system for features.

Closes: #488 
References: #153 - #170 - #191 - #358 - #447 - #457

PR #551, opened 2016-12-19
Ignite add is working!

GantMan:
[28 days ago](https://github.com/infinitered/ignite/pull/498), I had this same thing working, but this time... GLUE GUN!

### Normal add process
image

### When there's a generator collision
image

image

**some caveats:**
* to make my life manageable I only worried about adding generators
* no docs
* code reads like a bad script, needs to become functions
* npm hardcoded over yarn for now

### DIFF
--- a/packages/ignite-empty-playground/package.json
+++ b/packages/ignite-empty-playground/package.json
@@ -7,7 +7,7 @@
   },
   "devDependencies": {
     "ignite": "0.2.0",
-    "ignite-basic-generators": "0.2.0",
-    "ignite-ui": "^16.2.0-RC.2040"
-  }
+    "ignite-basic-generators": "0.2.0"
+  },
+  "dependencies": {}
 }
--- a/packages/ignite-empty-playground/package.json
+++ b/packages/ignite-empty-playground/package.json
@@ -7,6 +7,7 @@
   },
   "devDependencies": {
     "ignite": "0.2.0",
-    "ignite-basic-generators": "0.2.0"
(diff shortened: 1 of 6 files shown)


packages/ignite-empty-playground/package.json
```diff
@@ -7,6 +7,7 @@
   },
   "devDependencies": {
     "ignite": "0.2.0",
-    "ignite-basic-generators": "0.2.0"
+    "ignite-basic-generators": "0.2.0",
+    "ignite-ui": "^16.2.0-RC.2040"
```
skellock: You'll probably want to remove this.  =)

packages/ignite/package.json
```diff
@@ -27,9 +27,12 @@
   "license": "MIT",
   "dependencies": {
     "gluegun": "0.7.3",
+    "json2toml": "^1.0.6",
```
skellock: Didn't realize our parser didn't write TOML.

skellock:
I don't think this reads like a bad script at all dude.

packages/ignite/package.json
GantMan: yeah, imagine my shock!  I was loke CMON!!!!!! MANNNNNNN!!!!!

packages/ignite-empty-playground/package.json
GantMan: bahahaha, I searched npm for something that started with `ignite-`

PR #565, opened 2016-12-26
Vector icons

GantMan:
theoretical addition of vector icons via plugin.

### DIFF
--- a/packages/ignite/package.json
+++ b/packages/ignite/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ignite",
-  "version": "0.2.0",
+  "version": "2.0.0",
   "description": "Ignite - The next version.",
   "main": "src/index.js",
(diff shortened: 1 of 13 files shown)


packages/ignite/package.json
```diff
@@ -27,6 +27,7 @@
   "license": "MIT",
   "dependencies": {
     "gluegun": "0.7.3",
+    "gluegun-patching": "^0.3.0",
```
GantMan: @skellock - I tried using this, but saw no way to actually use it, so I just brought in patching first class.  Need to fix this.

skellock: npm install it.  then add it as a gluegun plugin in the cli.

GantMan: make a note to show me when you get back, gonna keep moving forward

PR #572, opened 2016-12-28
Decouple Ignite themes from ignited app

derekgreenberg:
- [x] `fireDrill.sh` passed

## Describe your PR
This PR decouples default themes (Colors, Fonts, Images) generated by ignite-base from themes that developers need to create for their Ignited apps. Each of these theme files now provide an empty object with a comment that tells the developer to add values for their own Ignited app.

### DIFF
--- a/ignite-base/App/Themes/Colors.js
+++ b/ignite-base/App/Themes/Colors.js
@@ -1,7 +1,9 @@
 // @flow
 
-const colors = {
-  background: '#1F0808',
+import { merge } from 'ramda'
+
+const igniteColors = {
+  igniteBackground: '#1F0808',
   clear: 'rgba(0,0,0,0)',
   facebook: '#3b5998',
   transparent: 'rgba(0,0,0,0)',
@@ -21,5 +23,9 @@ const colors = {
   fire: '#e73536',
   drawer: 'rgba(30, 30, 29, 0.95)'
 }
+
+const colors = merge(igniteColors, {
+  // add colors for your Ignited app here
+})
 
 export default colors
--- a/ignite-base/App/Themes/Colors.js
+++ b/ignite-base/App/Themes/Colors.js
@@ -1,7 +1,5 @@
 // @flow
 
-import { merge } from 'ramda'
-
 const igniteColors = {
   igniteBackground: '#1F0808',
   clear: 'rgba(0,0,0,0)',
@@ -24,8 +22,9 @@ const igniteColors = {
   drawer: 'rgba(30, 30, 29, 0.95)'
 }
 
-const colors = merge(igniteColors, {
+const colors = {
+  ...igniteColors
   // add colors for your Ignited app here
-})
+}
 
 export default colors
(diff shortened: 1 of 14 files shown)


ignite-base/App/Themes/Colors.js
```diff
@@ -22,4 +24,8 @@ const colors = {
   drawer: 'rgba(30, 30, 29, 0.95)'
 }
 
+const colors = merge(igniteColors, {
+  // add colors for your Ignited app here
+})
+
```
GantMan: I'm playing with the idea in my head of all the ignite colors (and other flavors, like metrics, strings etc) coming from a node module, so that they would not be a distracting 20line block in this code?  Or is that too petty?

ignite-base/App/Themes/Colors.js
```diff
@@ -1,7 +1,9 @@
 // @flow
 
-const colors = {
-  background: '#1F0808',
+import { merge } from 'ramda'
+
+const igniteColors = {
+  igniteBackground: '#1F0808',
```
GantMan: What was the motivation for renaming background color?

derekgreenberg: background is a common name for the background color of Ignited apps. This enables devs to use 'background' as a color for their own apps without changing the background color of the Ignite-generated screens such as ignite/ignite-base/App/Containers/Styles/ListviewExampleStyle.js

ignite-base/App/Themes/Colors.js
derekgreenberg: No, not too petty at all. This furthers my intent of decoupling all ignite flavors from the ignite-generated app. I want users to be able to open ignite screens such as the presentationScreen and access all of the useful screens, such as componentExamples, while being able to set up the flavors for their own app, with minimal conflicts.

derekgreenberg: I do like the generated Metrics, however, for use in my Ignited projects

ignite-base/App/Themes/Colors.js
jamonholmgren: @derekgreenberg Isn't it namespaced within `igniteColors` in this case?

skellock:
Take what @derekgreenberg has done one step further.  I say we remove all metrics, colors, and fonts entirely.  Leave only comments.

This PR seems to exist because of our "dev screens".  So, I vote to move dev screen colours and images and metrics over to that new plugin instead of in the root here.

So a freshly ignited app starts with *no* colours or metrics.

ignite-base/App/Themes/Colors.js
derekgreenberg: @jamonholmgren no, it is not namespaced at all. Colors.background is currently set to the values set in Colors, which will not match what a user wants for their own ignited app. The goal here is to decouple Ignite's default settings for colors - in particular Colors.background, from the Color.background that a user might want for their own Ignited app. So when you first generate a project via Ignite, the pre-ignited screens (which I love and want to preserve!) use the background color generated for all pre-ignited screens. But an Ignited app should be able to use Colors.background for their own app without breaking the presentation of pre-ignited screens such as ignite/ignite-base/App/Containers/Styles/ListviewExampleStyle.js

kevinvangelder:
I think this is a good fix for 1.0 until we get 2.0 up and running. As mentioned by Steve, the method of addressing this in 2.0 is going to be a totally different implementation.

derekgreenberg:
Removed merge() in place of simple use of spread operator.

jamonholmgren:
@GantMan @skellock  Merge this?

Issue #578, opened 2017-01-03
Redux generator is not upcasing anymore


skellock:
When you `ignite g redux Foo`, the `FooRedux.js` has the `createReducer()` block that looks like this:

```js

/* ------------- Hookup Reducers To Types ------------- */

export const reducer = createReducer(INITIAL_STATE, {
  [Types.Foo_REQUEST]: request,
  [Types.Foo_SUCCESS]: success,
  [Types.Foo_FAILURE]: failure
})

```

Those `Types.Foo` need to be uppercase for this to work.

This fix https://github.com/infinitered/ignite/pull/531 should be uppercased after splitting on the _ boundaries.

Found by Thiemo Bleeken @thiemo--bleeken in the Gitter channel.

skellock:
Just a heads up.  This is merged, but not released yet.

projoneftw:
I know it may seem trivial to the experienced guys, but maybe you should add it to troubleshooting section in readme until release - I burned a good couple of hours on it :) 
Found this issue only after I fixed it, the error is "take(patternOrChannel): patternOrChannel is undefined"

PR #579, opened 2017-01-03
Ensures uppercase when creating reducers with the redux generator.

skellock:
This is a fix for #578.

The redux generator now will uppercase the `createReducer()` keys.


GantMan:
I assume you'll do a second PR on https://github.com/infinitered/ignite/blob/next/packages/ignite-basic-generators/templates/redux.ejs#L43-L45

skellock:
Not sure.  Depends on what the `name` is.

PR #580, opened 2017-01-04
Adds ignite plugin path and improves ignite add support.

skellock:
This PR adds a round of fun to the `ignite add` command.

* Gluegun now supports configurable template directories, so here's support for that + 1 example (vector-icons). Basic structure is next, but not part of this PR because I want to do something awesome with bulk template generation in gluegun.

* Switched `Shell.exec` to `context.system.run`.

* Added a switch called `--npm` to `ignite add` because `yarn` insists on validating existing modules against npm.  Since ignite 2.0 isn't shipped, this blows up `yarn`.  But not `npm`.

* Adds better code docs inside the `ignite` gluegun plugin.

* Adds support for ignite 3rd party folk to opt-in to `react-native link` and installing modules as dev dependencies.

* Added some try/catches in order to centralize error messages.

### DIFF
--- a/packages/ignite/src/plugins/default/commands/add.js
+++ b/packages/ignite/src/plugins/default/commands/add.js
@@ -1,14 +1,10 @@
 // @cliDescription  Add a new thingy
 // ----------------------------------------------------------------------------
-const Shell = require('shelljs')
 const Exists = require('npm-exists')
 const Toml = require('toml')
 // Yeah, why would toml include this? :(
 const json2toml = require('json2toml')
 const R = require('ramda')
-
-// use yarn or use npm? hardcode for now
-const useYarn = false
 
 // used for changes warnings
 const detectedChanges = (oldObject, newObject) => {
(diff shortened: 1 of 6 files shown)


packages/ignite/src/plugins/default/commands/add.js
```diff
@@ -42,20 +42,23 @@ module.exports = async function (context) {
...
+  info(`🔎    Finding ${moduleName} on npmjs.com`)
   const moduleExists = await Exists(moduleName)
   // it exists?  Let's install it else warn
   if (moduleExists) {
-    success(`${checkmark}    Installing`)
+    info(`${checkmark}    Installing npm module`)
 
     if (ignite.useYarn) {
-      Shell.exec(`yarn add ${moduleName} --dev`, {silent: true})
+      Shell.exec(`yarn add ${moduleName}`, {silent: true})
```
GantMan: "this doesn't look right" - Steve

GantMan: We should change this to also support the config flags.  Should be dev dep by default.

GantMan: add comments plz, to make sure we stop getting confused here.

GantMan:
shit just got real - http://i.imgur.com/FpltJvO.gifv

skellock:
I lol'ed!

Issue #521, opened 2016-12-01
rehydrate/load from asyncstorage too slow?


pvinis:
hello. i use some ignite stuff in an app i develop. i have the loginredux and loginsaga similar to the base ignite app. i have an if isLoggedIn to show the login screen otherwise show another screen. i log in in the beginning, then close the app. next time i open it, i get the login screen for a few milliseconds and then it jumps to the main screen. i console logged some stuff, and it seems like the login data is loaded but a bit later than the first render. how can i fix that? how can i get more info about it?

rt2zz:
while I do not know specifically in the context of ignite, the traditional way to solve this problem is to render a loading state until persistence has completed. You can do this either by 
1. have one of your reducers store when boostrap is complete
```
persistStore(store, config, () => {
  store.dispatch({ type: BOOTSTRAP_COMPLETE })
})
// in some reducer
case BOOTSTRAP_COMPLETE: 
  return {...state, boostraped: true}
```

2. use react lifecycle to delay render (https://github.com/rt2zz/redux-persist/blob/master/docs/recipes.md#delay-render-until-rehydration-complete)

pvinis:
aha. i will try both and see. thanks :)

pvinis:
ok i did the first thing you suggested and it worked. should i maybe make a pr to include this in the `startup` redux part of ignite?

pvinis:
ok i actually made a pr. it was small enough to do it quickly today.

skellock:
see #582 and #715 .

PR #582, opened 2017-01-06
during startup, we now have a check if everything is loaded already from storage.

pvinis:
- [X] `fireDrill.sh` passed

## Describe your PR
for example, before, if a user was logged in, the log in button would show up for a split second while the async storage loaded, and when it was finally loaded, it would swap for a log out button. now, we can have a check if the storage is loaded, so we can display an animation or a spinner, and then immediately show the log out button. no more flashing from log in -> log out.

### DIFF
--- a/ignite-base/App/Containers/PresentationScreen.js
+++ b/ignite-base/App/Containers/PresentationScreen.js
@@ -1,7 +1,9 @@
 // @flow
 
-import React from 'react'
+import React, { PropTypes } from 'react'
 import { ScrollView, Text, Image, View } from 'react-native'
+import { connect } from 'react-redux'
+import { isLoadedFromStorage } from '../Redux/StartupRedux'
 import { Images } from '../Themes'
 import RoundedButton from '../Components/RoundedButton'
 import { Actions as NavigationActions } from 'react-native-router-flux'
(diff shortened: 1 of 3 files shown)


pvinis:
this was inspired by #521.

jamonholmgren:
@skellock Should we merge?

skellock:
Haven't tried it yet.  Can you give it a go?

skellock:
Hey @pvinis , we're about to merge Ignite version 2 into master which will cause this PR to conflict.

If you were to rebase, you'd see the whole structure changed, and since I don't want you to flip your table, I wanted to give you a heads up that I'm closing this (and thank you!).

These changes are still relevant... so we'll absorb them before 2.0 launches.  Not quite sure how that will work without rebasing, so it might be easier to copy & paste this sucker.  Which sucks because this is yours to get credit for.

We're going to revisit Redux Persist as it's a valuable plugin.  I feel like some of these issues are us not holding it right.

Thanks for your time and expertise Pavlos. :clinking_glasses:

pvinis:
no problem. I just want that to be there. it was useful for me and I guess it will be for others too. looking forward to the new version.

PR #584, opened 2017-01-09
Use Husky rather than deprecated ghooks

daviscabral:
`ghooks` is now deprecated and recommends the use of `Husky` instead. This PR replaces it.

See gtramontina/ghooks#166 for more info.


GantMan:
fine by me!

GantMan:
We need to bring this in to Unholy for 2.0 - or make it a plugin.

skellock:
Just tested this.  Works great!

Issue #583, opened 2017-01-09
[2.0] Proposed structure of the Ignite folder and example usage


GantMan:
### Proposed structure of the Ignite folder and example usage:
```
/ignite
  ignite.toml
  Examples/
    index.js
    ignite-vector-icons
      Components/vectorExample.js
      Components/Styles/vectorExampleStyle.js
      Containers/VectorsGoneWild.js
      Containers/Styles/VectorsGoneWild.js
    some-other-thing
      Components/stuff.js
  DevScreens/
  Spork/
    ignite-basic-screen-generators/
      screen.ejs
      screen-style.ejs
```

GantMan:
cc: @kevinvangelder - this is what we talked about this morning via `next-devscreens`.  Ping Steve or I to get more detail.

skellock:
This is in place.

PR #590, opened 2017-01-13
Devscreens Functionality

GantMan:
Dev screens plugin.  Currently copying to app folder, but soon to copy to correct ignite folder with all other goodies as proposed in https://github.com/infinitered/ignite/issues/583


GantMan:
wtf

PR #591, opened 2017-01-14
Start ignite folder

GantMan:
Now when you create basic structure ignite items are stored in the `/ignite` folder.
image

This PR also fixes:
* copy of templates not having/using template generator
* setGlobal/removeGlobal from toml using new location
* toml in root as secondary option OK
* small aesthetic changes
* copy launch screen in dev screens
* copyBatch error when no setting is found


GantMan:
Got everybody using `./ignite/ignite.toml` now.

PR #600, opened 2017-01-22
Adds additional ways to show the version number.

skellock:
Adds a few additional ways to ask for the version number.

```sh
ignite version
ignite --version
ignite -version
ignite -v
```

The last 3 are what this PR adds.  And they do that by manually calling the version command.  :)

Requested by https://github.com/infinitered/gluegun/issues/131.

### DIFF
--- a/packages/ignite/src/cli/cli.js
+++ b/packages/ignite/src/cli/cli.js
@@ -25,9 +25,10 @@ module.exports = async function run (argv) {
   // parse the commandLine line
   const commandLine = minimist(argv.slice(2))
 
-  // are we asking for --version|-version|-v and nothing else?
-  const showVersionNumber = isEmpty(commandLine._) && commandLine.version || commandLine.v
-  if (showVersionNumber) {
+  // should we show the version number & jet?
+  const hasNoArguments = isEmpty(commandLine._)
+  const hasVersionOption = commandLine.version || commandLine.v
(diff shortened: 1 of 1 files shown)


packages/ignite/src/cli/cli.js
```diff
@@ -24,10 +22,20 @@ module.exports = async function run (argv) {
     .token('extensionName', 'contextExtension')
     .createRuntime()
 
+  // parse the commandLine line
+  const commandLine = minimist(argv.slice(2))
+
+  // are we asking for --version|-version|-v and nothing else?
+  const showVersionNumber = isEmpty(commandLine._) && commandLine.version || commandLine.v
```
kevinvangelder: Can you group these so it is more clear?

skellock: We use `commandLine` down below as well.  See that section with `commandLine.wtf`?

kevinvangelder: Sorry, I meant the `&&` and `||`, it's not clear which takes priority.

skellock: The latter.  Expressions evaluation goes left to right.  I think it's the `isEmpty()` parenthesis that's making this distracting.  I just checked in a simpler version which breaks it down a bit. Hope that's easier to read.

PR #608, opened 2017-01-25
Modalized DevScreens

kevinvangelder:
@GantMan @skellock


packages/ignite-dev-screens/templates/DevscreensButton.ejs
```diff
@@ -1,21 +1,89 @@
 // @flow
```
skellock: Let's name this `DevscreensButton.js.ejs`. <3

PR #609, opened 2017-01-25
spork copies template files

GantMan:
spork


GantMan:
@skellock  - gonna help the last leg of this spork/dir check.

skellock:
Sporked templates are now activated.

I'm not entirely happy with my implementation, but it was really straight forward this way.  I'll let it sit for a few days and see how I feel about it.

What I did was made a `context.ignite.generate()` which looks for a sporked version of the template before handing it off to `context.template.generate()`.

It works well, and we only have to make sure people call `context.ignite.generate()` from their generators.  Which is fine because it's our API... I'm going to think a bit more on it, but I might promote it up to gluegun.

packages/ignite-basic-generators/templates/redux.ejs
```diff
@@ -40,7 +40,7 @@ export const failure = state =>
 /* ------------- Hookup Reducers To Types ------------- */
 
 export const reducer = createReducer(INITIAL_STATE, {
-  [Types.{{ props.name | upperCase }}_REQUEST]: request,
-  [Types.{{ props.name | upperCase }}_SUCCESS]: success,
-  [Types.{{ props.name | upperCase }}_FAILURE]: failure
+  [Types.<%= snakeCase(props.name).toUpperCase() %>_REQUEST]: request,
```
skellock: I think a screamingSnakeCase() might be appropriate here.

packages/ignite/src/plugins/default/commands/spork.js
```diff
@@ -0,0 +1,55 @@
...
+    type: 'checkbox',
+    choices
+  })
+
+  // TODO: This will be wonky if you're not in root of your project
+  copyFiles.selectedTemplates.map((template) => {
+    filesystem.copyAsync(`${directory}/templates/${template}`, `ignite/Spork/${selectedPlugin}/${template}`)
+  })
+
+  success('Sporked!')
```
skellock: This fires even if we didn't select anything.

packages/ignite/src/plugins/default/commands/spork.js
```diff
@@ -0,0 +1,55 @@
...
+  const copyFiles = await context.prompt.ask({
+    name: 'selectedTemplates',
+    message: 'Which templates would you like to spork?',
+    type: 'checkbox',
+    choices
+  })
+
+  // TODO: This will be wonky if you're not in root of your project
+  copyFiles.selectedTemplates.map((template) => {
+    filesystem.copyAsync(`${directory}/templates/${template}`, `ignite/Spork/${selectedPlugin}/${template}`)
```
skellock: This clobbers any existing sporked template without warning.  Which is fine in my books, but in other places we prompt for confirmation.

PR #618, opened 2017-01-26
CLI aliases and hidden commands.

skellock:


jamonholmgren:
Could we update the `docs/` folder too?

PR #627, opened 2017-01-27
ignite-i18n package

jamonholmgren:
Here's the `ignite-i18n` package. I'll make notes on relevant parts.


ignite-base/App/Containers/App.js
```diff
@@ -1,14 +1,11 @@
 // @flow
```
jamonholmgren: We're loading `../Config` at the top because load order matters and we want to make sure all the config has a chance to be applied before we load anything else.

kevinvangelder: Are you sure you want to update an Ignite 1.0 file here?

PR #630, opened 2017-01-27
Update android shake command to not fail

Blackclaws:
- [x ] Everything works on iOS/Android - Nothing changed in that regard
- [x ] `ignite-base` **ava** tests pass
- [x ] `fireDrill.sh` passed

## Describe your PR

Android shake command fails for me as the previous command greps "List" out of the result and adb fails with unknown device "List"

This change greps all devices out of the device List and then omits the line containing "List".

### DIFF
--- a/ignite-base/package.json
+++ b/ignite-base/package.json
@@ -19,7 +19,7 @@
     "android:hockeyapp": "cd android && ./gradlew assembleRelease && puck -submit=auto app/build/outputs/apk/app-release.apk",
     "android:devices": "$ANDROID_HOME/platform-tools/adb devices",
     "android:logcat": "$ANDROID_HOME/platform-tools/adb logcat *:S ReactNative:V ReactNativeJS:V",
-    "android:shake": "$ANDROID_HOME/platform-tools/adb devices | grep '\\t' | awk '{print $1}' | sed 's/\\s//g' | xargs -I {} $ANDROID_HOME/platform-tools/adb -s {} shell input keyevent 82",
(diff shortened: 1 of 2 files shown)


GantMan:
@skellock - check this out.  
BTW< I just used this command recently, as for some reason command+M wasn't working

skellock:
That's why I assigned myself.  I'll check it out as soon as I can.

Ya that's strange that it's not working.  The `\t` should skip the "List" text.  Had not encountered that yet.

Blackclaws:
Interestingly enough it doesn't. Even worse it only prints that line. 
I've found out why though now.
\t does not mean tab. At least not in the linux version of grep. It instead simply means an escaped t. Hence it matches List and devices that contain "emulator"

An alternative fix (which is much easier) is to just replace grep with grep -P which uses perl regex.

I'll update this accordingly.

skellock:
Ohhhhhhhh!    Ok.  Ha!  Thx man!  Appreciate this.

Blackclaws:
Ok I got it.

The way package.json is currently delivered it is: grep "\\t" which expands to grep "\t" which doesn't work.

There are two possible fixes. Either turn it into grep -P "\\t" or get rid of the second \ which causes the \t to expand to a tab.

Blackclaws:
I've updated the pull request. I am however unsure as to why the command does not fail for you without this modification.

Maybe it depends on the OS? Does the new command fail for you? If it does we might need to use my original fix which should be OS independent.

skellock:
Thanks.  Both of these work.  With and without the extra backlash.  (osx)

If this fixes things on Linux, then I'm down for merging!

PR #632, opened 2017-01-29
Next prompt

### LINKED ISSUE #620: All `new` prompts should be multiple choice or multi-select
To streamline the user experience we want to only present multiple choice or multi-select questions. This also helps future-proof for when we start getting more plugins of a given type.

kevinvangelder:
Closes #620.

- Make all prompts use `list` for consistency and ease of use.
- Make all prompts default to `none` or `no` for consistency and ease of use.
- Fixed lingering I18n references in Unholy that only appear if you chose not to install I18n.
- Fixed lingering drawer navigation items for moved dev screens.

### DIFF
--- a/packages/ignite-basic-structure/templates/Unholy/Transforms/ConvertFromKelvin.js
+++ b/packages/ignite-basic-structure/templates/Unholy/Transforms/ConvertFromKelvin.js
@@ -1,14 +1,8 @@
 // @flow
 
-import I18n from 'react-native-i18n'
-
 export default (kelvin: number) => {
   const celcius = kelvin - 273.15
   const farenheit = (celcius * 1.8000) + 32
 
-  if (I18n.t('tempIndicator') === 'F') {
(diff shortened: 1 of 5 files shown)


packages/ignite-basic-structure/templates/Unholy/Transforms/ConvertFromKelvin.js
```diff
@@ -1,14 +1,8 @@
 // @flow
 
-import I18n from 'react-native-i18n'
-
 export default (kelvin: number) => {
   const celcius = kelvin - 273.15
   const farenheit = (celcius * 1.8000) + 32
 
-  if (I18n.t('tempIndicator') === 'F') {
```
kevinvangelder: This was a clever way to display the temperature in Celsius if the user's language was anything but English, but since I18n is now a generator we can't do this by default and besides, not all English speaking countries even use Fahrenheit...

PR #635, opened 2017-01-30
Ignite animatable

GantMan:
animatable


packages/ignite-animatable/index.js
```diff
@@ -0,0 +1,33 @@
...
+ * Add ourself to the project.
+ */
+const add = async function (context) {
+  const { ignite } = context
+
+  // install a npm module
+  await ignite.addModule(NPM_MODULE_NAME)
+
+  // copy the example file (if examples are turned on)
+  await ignite.addComponentExample(EXAMPLE_FILE, { title: 'Animatable' })
```
kevinvangelder: Should this not be `addPluginExample`?

skellock: I think we only have [addComponentExample](https://github.com/infinitered/ignite/blob/next/packages/ignite/src/plugins/default/extensions/ignite.js#L178) right now.

GantMan: let's open a ticket on renaming this function.

PR #641, opened 2017-02-03
Fixes outstanding 2.0 project setup issues

skellock:
This brings us back into a good place again with 2.0.

* fixes `react-native link` does something weird which causes a hang.
* temporary comments out the directory safety checks.  will sync up with @kevinvangelder to fix 'er up.
* fixes & re-enables yarn support

Don't forget to `cd packages/ignite && yarn` to grab the latest deps.

### DIFF
--- a/packages/ignite/src/plugins/default/commands/new.js
+++ b/packages/ignite/src/plugins/default/commands/new.js
@@ -54,47 +54,51 @@ module.exports = async function (context) {
   if (!parameters.options.min) {
     answers = await context.prompt.ask(installWalkthrough)
   }
-  // then we kick off (TODO: Would be awesome to have this kick off during questions)
+
   // we need to lock the RN version here
-  info('Creating new RN project')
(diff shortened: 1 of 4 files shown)


packages/ignite/src/plugins/default/commands/new.js
```diff
@@ -54,47 +54,51 @@ module.exports = async function (context) {
...
     answers = await context.prompt.ask(installWalkthrough)
   }
-  // then we kick off (TODO: Would be awesome to have this kick off during questions)
+
   // we need to lock the RN version here
-  info('Creating new RN project')
-  // TODO make sure `react-antive --version has react-native-cli 2.x otherwise failure`
+  // TODO make sure `react-native --version has react-native-cli 2.x otherwise failure`
   const reactNativeVersion = '0.40.0'
+  info(`🌮  creating project with react native ${reactNativeVersion}`)
```
GantMan: bahahahaha

PR #650, opened 2017-02-04
Clears the build folder in ios when newclearing.

### LINKED ISSUE #624: Improve iOS cache clear in newclear
Right now in our `package.json` in the newclear command, we're not aggressive enough to clear the Xcode cache.  It should be `rm -rf ios/build` instead of `rm -rf ios/build/ModuleCache/*`.

skellock:
This clears the whole build folder under ios when using `newclear`.  Fixes #624.


GantMan:
LGTM, we should really make newclear a command in ignite

PR #653, opened 2017-02-04
Upgrades ignite doctor.

skellock:
This started by moving `which` into gluegun.  Ended up reworking the whole command.

I think there's even more we can do as well.  Like list the current installed android sdks and more.

```
❯ ignite doctor
-----------------------------------------------
  (                  )   (
  )\ )   (        ( /(   )\ )    *   )
 (()/(   )\ )     )\()) (()/(  ` )  /(   (
  /(_)) (()/(    ((_)\   /(_))  ( )(_))  )\
 (_))    /(_))_   _((_) (_))   (_(_())  ((_)
 |_ _|  (_)) __| | \| | |_ _|  |_   _|  | __|
  | |     | (_ | | .` |  | |     | |    | _|
 |___|     \___| |_|\_| |___|    |_|    |___|
-----------------------------------------------

An unfair headstart for your React Native apps.
https://infinite.red/ignite

-----------------------------------------------

System
  platform           darwin
  arch               x64
  cpu                8 cores      Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz

JavaScript
  node               7.5.0        /usr/local/bin/node
  npm                4.1.2        /usr/local/bin/npm

React Native
  react-native-cli   2.0.1
  app rn version     0.40.0

Ignite
  ignite             2.0.0        /usr/local/bin/ignite

Native Mobile Platforms
  xcode              8.2.1
  java               1.8.0_45     /usr/bin/java
  android home       -            /Users/steve/Library/Android/sdk
```

I removed the triple backtick formatting as I think we can do something a little different/better for people trying to submit issues.

🚨 @jamonholmgren @kevinvangelder @GantMan 🚨  -> once this is merged in, we'll need to `cd packages/ignite && yarn` as gluegun got bumped to `0.11.0`.

### DIFF
--- a/packages/ignite/src/commands/doctor.js
+++ b/packages/ignite/src/commands/doctor.js
@@ -1,49 +1,82 @@
 // @cliDescription Checks your dev environment for dependencies.
 // ----------------------------------------------------------------------------
 
-const R = require('ramda')
-const Shell = require('shelljs')
+const { split, last, replace, head, match } = require('ramda')
+const header = require('../brand/header')
+const os = require('os')
 
 module.exports = async function (context) {
-  const { filesystem: { exists, read } } = context
-  const { debug } = context.print
+  // fistful of features
+  const {
+    filesystem: { read },
+    system: { run, which },
+    print: { colors, info, table },
+    strings: { padEnd }
+  } = context
 
   // ... be the CLI you wish to see in the world
   const platform = process.platform
-  const ignitePath = R.trim(Shell.which('ignite') || '')
-  const igniteVersion = R.trim(Shell.exec('ignite version', { silent: true }).stdout)
-  const nodePath = R.trim(Shell.which('node'))
-  const npmVersion = R.trim(Shell.exec('npm --version', { silent: true }).stdout)
-  const npmPath = R.trim(Shell.which('npm'))
-  const nodeVersion = R.trim(Shell.exec('node --version', { silent: true }).stdout)
-  const ggPackageFile = `${process.cwd()}/node_modules/gluegun/package.json`
-  const ggVersion = exists(ggPackageFile) ? read(ggPackageFile, 'json').version : '¯\\_(ツ)_/¯'
+  const arch = os.arch()
+  const cpus = os.cpus() || []
+  const firstCpu = head(cpus) || {}
+  const cpu = `${firstCpu.model}`
+  const cores = `${cpus.length} cores`
+  const ignitePath = which('ignite')
+  const igniteVersion = await run('ignite version', { trim: true })
+  const nodePath = which('node')
+  const npmVersion = await run('npm --version', { trim: true })
+  const npmPath = which('npm')
+  const nodeVersion = replace('v', '', await run('node --version', { trim: true }))
+  const rnCli = split(/\s/, await run('react-native --version', { trim: true }))[1] // lulz
+  const xcodeVersion = split(/\s/, await run('xcodebuild -version', { trim: true }))[1] // lulz
+  const rnPkg = read(`${process.cwd()}/node_modules/react-native/package.json`, 'json')
+  const appReactNativeVersion = rnPkg ? rnPkg.version : '-'
+  const androidPath = process.env['ANDROID_HOME']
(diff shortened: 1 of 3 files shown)


packages/ignite/src/commands/doctor.js
```diff
@@ -1,49 +1,82 @@
...
-\`\`\`
-`
-  debug(body, 'Ignite Info')
+  const arch = os.arch()
+  const cpus = os.cpus() || []
+  const firstCpu = head(cpus) || {}
+  const cpu = `${firstCpu.model}`
+  const cores = `${cpus.length} cores`
+  const ignitePath = which('ignite')
+  const igniteVersion = await run('ignite version', { trim: true })
```
GantMan: I think it would be good to print the version a project was Ignited with (in the Toml file) - that way we can see if they are running a command from X+1 on a project ignited at X.

packages/ignite/src/commands/doctor.js
```diff
@@ -1,49 +1,82 @@
...
+    [column1('platform'), column2(platform)],
+    [column1('arch'), column2(arch)],
+    [column1('cpu'), column2(cores), column3(cpu)]
+  ])
+
+  info('')
+  info(colors.cyan('JavaScript'))
+  table([
+    [column1('node'), column2(nodeVersion), column3(nodePath)],
+    [column1('npm'), column2(npmVersion), column3(npmPath)]
```
GantMan: we should probably check if they have `yarn`, and if so what version, since our CLI will try to use Yarn and yarn has had quite a few bugs.

packages/ignite/src/commands/doctor.js
```diff
@@ -1,49 +1,82 @@
...
+  const arch = os.arch()
+  const cpus = os.cpus() || []
+  const firstCpu = head(cpus) || {}
+  const cpu = `${firstCpu.model}`
+  const cores = `${cpus.length} cores`
+  const ignitePath = which('ignite')
+  const igniteVersion = await run('ignite version', { trim: true })
+  const nodePath = which('node')
+  const npmVersion = await run('npm --version', { trim: true })
+  const npmPath = which('npm')
```
GantMan: some people using ignite said it failed bc they didn't even have npm on their machines at all.   Should we branch only if which actually returns?  That way we don't have failing commands if there's no NPM anyhow?

GantMan: Should we check that npm has proper permissions?

`test -w $(npm config get prefix)/lib/node_modules && echo "npm is totes tacos" || echo "shiiiiiiiit :("`

Actually, probably overkill to check that since 2.0 doesn't need the same magic 1.0 needed.

packages/ignite/src/commands/doctor.js
```diff
@@ -1,49 +1,82 @@
...
+  table([
+    [column1('ignite'), column2(igniteVersion), column3(ignitePath)]
+  ])
+
+  info('')
+  info(colors.cyan('Native Mobile Platforms'))
+  table([
+    [column1('xcode'), column2(xcodeVersion)],
+    [column1('java'), column2(javaVersion), column3(javaPath)],
+    [column1('android home'), column2('-'), column3(androidPath)]
```
GantMan: @kevinvangelder - should we print any Windows project info here?

packages/ignite/src/commands/doctor.js
skellock: Sure. Let's add that when we start recording the version number.

packages/ignite/src/commands/doctor.js
skellock: npm comes with node since version 4.  I'd be surprised to see someone without a valid npm, but we could check for that too if you want.

Ya special permissions aren't required anymore because we're not writing to any system directory.

packages/ignite/src/commands/doctor.js
skellock: good call.  overlooked that one.

skellock:
Ok, I added:

* `yarn` detection
* safety checks to all `run` commands so it doesn't die
* better organization
* a windows section for our man @kevinvangelder when he's ready to check that out.

Good suggestions @GantMan .

skellock:
image

GantMan:
perfect 🙌

PR #654, opened 2017-02-06
Fixes unholy package.json clobbering RN version.

skellock:
The unholy template copies it's `package.json` overtop of the one installed by React Native so we end up losing the user's choice.

This fixes that.


GantMan:
good catch.

PR #655, opened 2017-02-06
Ignite app configuration is now JSON.

skellock:
### JSON

Ignite's configuration (`ignite/ignite.json`) is now `json`.  

image

I also dropped the `ignite` key which removes 1 level.  It wasn't needed because we're not sharing this with gluegun any longer.

### New API

If you need to work with this config from a plugin:

```js
// to read
const config = context.ignite.loadIgniteConfig()

// to write
config.omg = 'yussss'
context.ignite.saveIgniteConfig(config)
```

### Breaking API

This also changes the API a bit.  These two functions were renamed.

```js
`setGlobalConfig` -> `setIgniteConfig`
`removeGlobalConfig` -> `removeIgniteConfig`
```

I kept wonder just what config `Global` meant.  It means the Ignite config.

`setGlobalConfig` also no-longer has a 3rd parameter.  I couldn't find any examples of it in action. @kevinvangelder lemme know if that's the case.

### Package Changes

See all those `package.json` changes?  They all have new version numbers now.

I had a lot of issues with caching that even `npm cache clean` couldn't fix... turns out we're going to need to start bumping `version` inside `package.json` of each package.json.

This was blocking me, so I fixed it here since I was testing everything end to end.

Nice side-effect now is `yarn` is now fully supported again.

### DIFF
--- a/packages/ignite-basic-structure/index.js
+++ b/packages/ignite-basic-structure/index.js
@@ -28,8 +28,8 @@ const add = async function (context) {
       target: 'README.md'
     },
     {
-      template: 'ignite.toml',
-      target: 'ignite/ignite.toml'
+      template: 'ignite.json.ejs',
+      target: 'ignite/ignite.json'
     },
     {
       template: '.editorconfig',
@@ -59,7 +59,8 @@ const add = async function (context) {
 
   await ignite.copyBatch(context, copyJobs, {
     name: parameters.third,
-    reactNativeVersion
+    reactNativeVersion,
+    igniteVersion: ignite.version
   })
 }
 
--- a/packages/ignite-dev-screens/index.js
+++ b/packages/ignite-dev-screens/index.js
@@ -30,7 +30,7 @@ const add = async function (context) {
   const { warning } = print
 
   // Set Examples to "classic" in Ignite config
-  context.ignite.setGlobalConfig('examples', 'classic')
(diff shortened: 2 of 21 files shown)


packages/ignite-basic-structure/index.js
```diff
@@ -28,8 +28,8 @@ const add = async function (context) {
       target: 'README.md'
     },
     {
-      template: 'ignite.toml',
-      target: 'ignite/ignite.toml'
+      template: 'ignite.json.ejs',
```
GantMan: yay!

packages/ignite-basic-structure/index.js
```diff
@@ -59,7 +59,8 @@ const add = async function (context) {
 
   await ignite.copyBatch(context, copyJobs, {
     name: parameters.third,
-    reactNativeVersion
+    reactNativeVersion,
+    igniteVersion: ignite.version
```
GantMan: woohoo!  I like it

packages/ignite-dev-screens/index.js
```diff
@@ -30,7 +30,7 @@ const add = async function (context) {
   const { warning } = print
 
   // Set Examples to "classic" in Ignite config
-  context.ignite.setGlobalConfig('examples', 'classic')
+  context.ignite.setIgniteConfig('examples', 'classic')
```
GantMan: I like this naming better.

PR #662, opened 2017-02-10
[OMG READY] Ignite 2 On Final Approach

skellock:
We took a poll.  Should we merge v2 into v1 now or later?

Team now won.

Ignite 2 has been living in `next`.  This is compatible with `master` because v2 lives under `/packages/` and shares nothing in common.  This will cause some minor nuisances and confusion with contributors, so we should mention something on the README.

Maybe we could move `v1` out of the way sooner rather than later?

But tonight I say move forward, not backward... https://www.youtube.com/watch?v=zGmfiKowrYU


GantMan:
dat commit log

jamonholmgren:
I'd like to get this README shined up before we merge. Make sure people know where we are and where we aren't, how to use Ignite 2 vs 1, etc.

skellock:
@GantMan I know right?  We kicked some ace.

@jamonholmgren I agree.  We should just wait before merging.  There's no reason yet.

packages/ignite-cli/readme.md
```diff
@@ -0,0 +1,6 @@
+# ignite-cli
+
+Currently this is an unused.  But we may move here in the future if we decide to split
```
jamonholmgren: > this is an unused.

unused what

Un3qual:
Maybe add a `--v2` option to the generator for now so that people can decide which version to use until v2 has been out for a while.

Un3qual:
Will `react-native` be updated from `38` to the latest (`42` I belive) in 2.0?

skellock:
@Un3qual We'll be publishing under a different name: `ignite-cli`, so you can keep Ignite 1.  Ignite 1 will remain at 0.39.2.  Ignite 2 will be tracking the latest official release (currently 0.41.2).

Un3qual:
Sounds good. Thanks!

PR #665, opened 2017-02-10
Adds `ignite plugin new` to CLI

jamonholmgren:
This adds basic plugin generation to Ignite 2.0.

```
$ ignite plugin new ignite-foo
```

screen shot 2017-02-09 at 10 46 32 pm

screen shot 2017-02-09 at 10 46 16 pm

### DIFF
--- a/packages/ignite-basic-structure/index.js
+++ b/packages/ignite-basic-structure/index.js
@@ -54,6 +54,9 @@ const add = async function (context) {
     filesystem.copy(`${sourceFolder}/App`, `${process.cwd()}/App`, { overwrite: true })
   }
 
+  // Create ./ignite/plugins
+  filesystem.dir(`${process.cwd()}/ignite/plugins`)
+
   // grab the react native version from the command line
   const reactNativeVersion = parameters.options['react-native-version'] || REACT_NATIVE_VERSION
 
--- a/packages/ignite-basic-structure/index.js
+++ b/packages/ignite-basic-structure/index.js
@@ -8,7 +8,7 @@ const add = async function (context) {
 
   // validation
   if (isBlank(parameters.third)) {
-    print.info(`${context.runtime.brand} add basic-structure <name>\n`)
+    print.info(`ignite add basic-structure <name>\n`)
     print.error('App name is required to properly brand structure')
     process.exit(1)
     return
@@ -54,8 +54,8 @@ const add = async function (context) {
     filesystem.copy(`${sourceFolder}/App`, `${process.cwd()}/App`, { overwrite: true })
   }
 
-  // Create ./ignite/plugins
-  filesystem.dir(`${process.cwd()}/ignite/plugins`)
+  // Create ./ignite/plugins/.gitkeep
+  filesystem.write(`${process.cwd()}/ignite/plugins/.gitkeep`, '')
 
   // grab the react native version from the command line
   const reactNativeVersion = parameters.options['react-native-version'] || REACT_NATIVE_VERSION
--- a/packages/ignite/src/commands/new.js
+++ b/packages/ignite/src/commands/new.js
@@ -34,6 +34,25 @@ const installWalkthrough = [
   }
 ]
 
+const minOptions = {
+  'dev-screens': 'No',
+  'vector-icons': 'none',
+  'i18n': 'none',
+  'animatable': 'none'
+}
+const maxOptions = {
+  'dev-screens': 'Yes',
+  'vector-icons': 'react-native-vector-icons',
+  'i18n': 'react-native-i18n',
+  'animatable': 'react-native-animatable'
+}
+
+const walkthrough = async (context) => {
+  if (context.parameters.options.min) { return minOptions }
+  if (context.parameters.options.max) { return maxOptions }
+  return await context.prompt.ask(installWalkthrough)
+}
+
 module.exports = async function (context) {
   const { parameters, strings, print, system } = context
   const { isBlank } = strings
@@ -58,10 +77,7 @@ module.exports = async function (context) {
   const igniteDevPackagePrefix = parameters.options.live || path.resolve(`${__dirname}/../../..`) + '/ignite-'
 
   // First we ask!
-  let answers = {}
-  if (!parameters.options.min && !parameters.options.max) {
-    answers = await context.prompt.ask(installWalkthrough)
-  }
+  const answers = await walkthrough(context)
 
   // we need to lock the RN version here
   // TODO make sure `react-native --version has react-native-cli 2.x otherwise failure`
@@ -99,19 +115,19 @@ module.exports = async function (context) {
   await system.spawn(`ignite add ${igniteDevPackagePrefix}basic-generators`, { stdio: 'inherit' })
 
   // now run install of Ignite Plugins
-  if (answers['dev-screens'] === 'Yes' || parameters.options.max) {
+  if (answers['dev-screens'] === 'Yes') {
     await system.spawn(`ignite add ${igniteDevPackagePrefix}dev-screens`, { stdio: 'inherit' })
   }
 
-  if (answers['vector-icons'] === 'react-native-vector-icons' || parameters.options.max) {
+  if (answers['vector-icons'] === 'react-native-vector-icons') {
     await system.spawn(`ignite add ${igniteDevPackagePrefix}vector-icons`, { stdio: 'inherit' })
   }
 
-  if (answers['i18n'] === 'react-native-i18n' || parameters.options.max) {
+  if (answers['i18n'] === 'react-native-i18n') {
     await system.spawn(`ignite add ${igniteDevPackagePrefix}i18n`, { stdio: 'inherit' })
   }
 
-  if (answers['animatable'] === 'react-native-animatable' || parameters.options.max) {
+  if (answers['animatable'] === 'react-native-animatable') {
     await system.spawn(`ignite add ${igniteDevPackagePrefix}animatable`, { stdio: 'inherit' })
   }
 
--- a/packages/ignite/src/commands/plugin.js
+++ b/packages/ignite/src/commands/plugin.js
@@ -1,0 +1,80 @@
+// @cliDescription Manages ignite plugins
+// @cliAlias p
+// ----------------------------------------------------------------------------
+
+const exitCodes = require('../lib/exitCodes')
+
+/**
+ * Checks whether a plugin name was given and errors if not.
+ *
+ * @param {Object} context The gluegun context.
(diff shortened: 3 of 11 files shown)


jamonholmgren:
This will almost certainly need improvement before we release, but that's probably best discovered by using it.

packages/ignite-basic-structure/index.js
```diff
@@ -54,6 +54,9 @@ const add = async function (context) {
     filesystem.copy(`${sourceFolder}/App`, `${process.cwd()}/App`, { overwrite: true })
   }
 
+  // Create ./ignite/plugins
+  filesystem.dir(`${process.cwd()}/ignite/plugins`)
```
skellock: Woo!

Maybe replace with this?

```js
filesystem.write(`${process.cwd()}/ignite/plugins/.gitkeep`, '')
```

Currently reading plugins from this directory isn't enabled yet.  We'd have to add another line like this to tell gluegun to load all plugins found in that directory.

https://github.com/infinitered/ignite/blob/next/packages/ignite/src/cli/cli.js#L18

This time it would be `hidden: false` though.  We still have some work to do convincing @GantMan that this is a good idea, but gluegun supports this out of the box.  🎉

packages/ignite/src/commands/new.js
```diff
@@ -34,6 +34,25 @@ const installWalkthrough = [
...
+  'vector-icons': 'none',
+  'i18n': 'none',
+  'animatable': 'none'
+}
+const maxOptions = {
+  'dev-screens': 'Yes',
+  'vector-icons': 'react-native-vector-icons',
+  'i18n': 'react-native-i18n',
+  'animatable': 'react-native-animatable'
+}
```
skellock: @kevinvangelder had pointed this out as a good candidate for refactoring.  Thanks for cleaning this up.  Love this.

packages/ignite/src/commands/plugin.js
```diff
@@ -0,0 +1,145 @@
...
+ * Checks whether a plugin name was given and errors if not.
+ *
+ * @param {Object} context The gluegun context.
+ * @returns {Boolean}
+ */
+const checkName = (pluginName, context) => {
+  const { strings, print, runtime, parameters } = context
+  
+  if (strings.isBlank(pluginName)) {
+    print.info(`${runtime.brand} plugin new <${runtime.brand}-plugin-name>\n`)
```
skellock: `runtime.brand` is 100% correct.  We use the `ignite` everywhere.  I wouldn't worry about using `runtime.brand` at all in ignite.

SEE, I DON"T HATE BRANDS JAMON.

packages/ignite/src/commands/plugin.js
```diff
@@ -0,0 +1,145 @@
...
+    print.error('Plugin name is required')
+    process.exit(exitCodes.PLUGIN_NAME)
+    return false
+  }
+
+  if (pluginName.indexOf('ignite-') !== 0) {
+    // As passive-aggressively as possible
+    print.info(`Ignite plugins typically start with ignite-* -- but you do you.`)
```
skellock: ROFL.  Get this out of here.  You guys win.  You can force `ignite-*` prefixes.  FFS.

packages/ignite/src/commands/plugin.js
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  // TODO: Make this better at detecting invalid plugin names
+  if (!/^[a-z0-9].*/i.test(pluginName)) {
+    print.error('Plugin name should be a valid folder name')
+    process.exit(exitCodes.PLUGIN_NAME)
+    return false
```
skellock: Thx for using exit codes.  You can remove this line.

packages/ignite/src/commands/plugin.js
```diff
@@ -0,0 +1,145 @@
...
+    { template: 'plugin/ignite.toml.ejs', target: `${pluginName}/ignite.toml` },
+    { template: 'plugin/package.json.ejs', target: `${pluginName}/package.json` },
+    (answers.template === 'Yes') &&
+      { template: 'plugin/templates/Example.js.ejs', target: `${pluginName}/templates/${name}Example.js` },
+    (answers.command === 'Yes') &&
+      { template: 'plugin/commands/example.js.ejs', target: `${pluginName}/commands/example.js` },
+  ]
+
+  // copy over the files
+  await ignite.copyBatch(context, copyJobs, {name, pluginName})
```
skellock: I think maybe gluegun's `context.template.generate` (and `context.ignite.generate`) should support this variant too. Instead of `copyBatch`, we could call `ignite.generate()` with the same arguments.

packages/ignite/src/commands/plugin.js
```diff
@@ -0,0 +1,145 @@
...
+
+module.exports = async function (context) {
+  // They need to be in the folder /ignite or /packages to create plugins
+
+  // const { parameters, strings, print, system } = context
+  const { parameters, print } = context
+  // const { debug, info, error } = print
+
+  // context.print.debug(context.parameters)
+  switch (parameters.second) {
```
skellock: Gluegun has support for doing what you're doing here.  So, instead of rolling your own `switch` statement, you would create a folder called `plugin`.  In that folder, you'd create another folder called `commands`.  In that commands folder, you'd have `new.js`, `list.js`, and `help.js`.

Then, we'd tell the CLI to load this plugin right around here:  https://github.com/infinitered/ignite/blob/next/packages/ignite/src/cli/cli.js#L17

```js
   .load(`${__dirname}/../plugin`)
```

packages/ignite/src/templates/plugin/commands/example.js.ejs
```diff
@@ -0,0 +1,26 @@
...
+    print.info(`ignite generate example <name>\n`)
+    print.info('A name is required.')
+    return
+  }
+
+  const name = pascalCase(parameters.first)
+
+  const copyJobs = [{
+    template: 'example.ejs',
+    target: `App/Examples/${name}.js`
```
skellock: Just a heads up, we're putting this in `ignite/Examples/${name}.js`.

packages/ignite/src/commands/plugin.js
GantMan: :FACT:

packages/ignite-basic-structure/index.js
GantMan: We'd have to change `add` to not use the node-modules folder.  Maybe I'm not understanding.  This feels like it might get complex pretty fast.

packages/ignite/src/commands/new.js
GantMan: 👍

packages/ignite/src/lib/exitCodes.js
```diff
@@ -31,10 +31,15 @@ module.exports = {
...
   /**
    * You're trying to spork, but there's nothing to eat
    */
-  SPORKABLES_NOT_FOUND: 6
+  SPORKABLES_NOT_FOUND: 6,
 
+  /**
+   * Invalid Plugin name.
+   */
+  PLUGIN_NAME: 7
```
GantMan: INVALID errors sometimes have the word invalid in them... we should standardize.

packages/ignite/src/templates/plugin/commands/example.js.ejs
GantMan: yup this should be changed before merge.

packages/ignite/src/templates/plugin/index.js.ejs
```diff
@@ -0,0 +1,54 @@
...
+
+  // install a npm module and link it
+  await ignite.addModule(NPM_MODULE_NAME, { link: true })
+
+  // copy the component example file (if examples are turned on)
+  await ignite.addComponentExample(EXAMPLE_FILE, { title: '<%= props.name %> Example' })
+
+  // Example of copying templates/<%= props.name %> to App/<%= props.name %>
+  // if (!filesystem.exists(`${APP_PATH}/App/<%= props.name %>`)) {
+  //   filesystem.copy(`${PLUGIN_PATH}/templates/<%= props.name %>`, `${APP_PATH}/App/<%= props.name %>`)
```
GantMan: Would an example of copybatch be good?

packages/ignite/src/templates/plugin/templates/Example.js.ejs
```diff
@@ -0,0 +1,12 @@
+// @flow
+
+import React from 'react'
+import { View, Text } from 'react-native'
+import ExamplesRegistry from '../../../App/Services/ExamplesRegistry'
```
GantMan: @skellock - we should really see about adding that `~/App` plugin by default so these become readable.

packages/ignite/src/lib/exitCodes.js
skellock: Probably could leave off the word INVALID.  These are exit error codes.

packages/ignite/src/templates/plugin/templates/Example.js.ejs
skellock: 👍  as long as it doesn't break things.  We should try pushing a production build with it to ensure there's no side effects.  It seems great.

packages/ignite-basic-structure/index.js
jamonholmgren: Let's do a Zoom about this. If it's as easy as Steve says then I don't see any reason not to support it.

skellock: 1 line of code in the CLI gets us custom plugins.   Boosh!  I'm available to chat anytime.

packages/ignite/src/commands/plugin.js
jamonholmgren: Good call. I'll leave this for a future update.

packages/ignite/src/templates/plugin/commands/example.js.ejs
jamonholmgren: This isn't actually an Ignite example like you guys are thinking. I sort of poorly named this file. This is a template file that is generated along with the plugin, intended for modification.

Replace "Examples" with "Components" to get a sense for what we're doing in this file. You could create a Component generator with this file. The "Examples" folder is just a placeholder that will be renamed by the plugin author.

packages/ignite/src/templates/plugin/index.js.ejs
jamonholmgren: Probably -- I'd imagine this file will see the most modifications as we start using this generator. I'll leave that until later if that's cool.

jamonholmgren:
w00t!

PR #671, opened 2017-02-11
Adds collapsing spinners

skellock:
image

The subtasks show as spinners and once complete, they collapse, leaving behind a nice checkmark.

I like it.  

You see the details of the subtasks when they're happening, but go away when the task is finished.  The end result is clean.

But take a test drive before merging to see what you think.

We still need a bit more work on each of the components, but you'll get the gist.


GantMan:
looks good to me!
spinner2

Issue #672, opened 2017-02-11
Clean out Devscreens 


GantMan:
As of right now, dev screens depend on app styles.  The schism needs to happen! ➗ 

1. Move styles over to dev screens
2. Make dev screens point to new styles
3. Delete unused styles from copied files
4. Delete unused styles from original themes folder

kevinvangelder:
Technically we've only done 1 & 2.

skellock:
Oops.  I did it again.

skellock:
Are we ok to close this issue now?  I feel like we are.

PR #673, opened 2017-02-11
Devscreens Cleanup

kevinvangelder:
Copied DevScreen styles and assets from Unholy. We'll need to go back and clean up the remaining DevScreen styles and assets from Unholy in a separate PR.


packages/ignite-dev-screens/index.js
```diff
@@ -1,27 +1,4 @@
...
-        template: `${screen}Style.js.ejs`,
-        target: `ignite/DevScreens/Styles/${screen}Style.js`
-      }
-    )
-  })
-
-  // make the templates
-  await ignite.copyBatch(context, jobs, {})
-}
+const sourceFolder = `${process.cwd()}/node_modules/ignite-dev-screens/templates`
```
jamonholmgren: This isn't related to this PR necessarily, but I think we should replace any instances of `process.cwd()` with `SOURCE_PATH` or something, and set it further up the file. I never really know what `process.cwd()` refers to. cc @GantMan @skellock for discussion

jamonholmgren:
Ref #672

packages/ignite-dev-screens/index.js
skellock: It'll always be the root of your ignited app.  In fact I think for most operations, you replace `${process.cwd()}/` with `./` or hopefully even `` (blank).

Symlinks are the only think I would imagine would shit the bed.  I don't think we're using symlinks yet.

`SOURCE_PATH` still is ambiguous to me since we're in the business of breeding plugins.

I will agree with you `process.cwd()` makes me stop & think sometimes too.  Someone should see if it's optional.

jamonholmgren: `ROOT_OF_IGNITED_APP` sounds good to me. heh

PR #675, opened 2017-02-12
Adds long-running integration tests.

skellock:
This adds `packages/ignite-integration-tests` so a `npm run bootstrap` will be required to set this up.

The idea of this package is to run some end-to-end integration tests.  @jamonholmgren was asking about these.

I've set up a handful to show how these can work.

They run the `ignite` cli with various arguments and look for expected output.

Have a look at `packages/ignite-integration-tests/test/ignite-new/newMin.test.js`.  That one will `ignite new MinApp --min` and test that right artifacts come out.

image

I've found some issues in our CLI with it already.  🙀


GantMan:
before tests we were all like
http://i.imgur.com/Lrkm9nP.gifv

GantMan:
1 question?  Why is this kept outside of ignite's package?

jamonholmgren:
I'm super happy with this!

skellock:
@GantMan 

First thing that crossed my mind was... bloody hell, I don't want to run these integration tests when I type `npm test`.  So instead of configuring `ava` to ignore, I made a new package.

Lazy. I know.

I planned on cheating too.  

We technically don't need a React Native app to test some of integrations.  Like generators, parts of the plugin system, etc. So that would buy us speed! We'd still need all of the `ignite-*` plugins so the game plan was to set dependencies.

TL;DR: I flipped a coin.  It came up tails.

@jamonholmgren - I'm super happy with you!

Issue #680, opened 2017-02-14
Add more fields to generated plugins


jamonholmgren:
Generated plugins should also include `description`, `license`, `devDependencies` in their generated `package.json`.

skellock:
Heads up; `devDependencies` are optional. I could be convinced of peer dependencies, but we decided not to for now.

jamonholmgren:
We should also add `author`, `url`, and also add a sample README.md file.

GantMan:
also the generated file has blah.js instead of blah.js.ejs for examples

jamonholmgren:
@robinheinze Could you fix these things in Ignite?

robinheinze:
Working on this now.

PR #683, opened 2017-02-15
Fix/update generated package json

robinheinze:
- [x] `fireDrill.sh` passed

## Describe your PR

Fixes https://github.com/infinitered/ignite/issues/680

This adds additional fields to the generated `package.json` for new plugins, and also corrects the issue where generated example files were `.js` instead of `.js.ejs`. 

I also added a sample README to new plugins.

### DIFF
--- a/packages/ignite/src/commands/plugin.js
+++ b/packages/ignite/src/commands/plugin.js
@@ -81,6 +81,7 @@ const createNewPlugin = async (context) => {
     { template: 'plugin/index.js.ejs', target: `${pluginName}/index.js` },
     { template: 'plugin/ignite.toml.ejs', target: `${pluginName}/ignite.toml` },
     { template: 'plugin/package.json.ejs', target: `${pluginName}/package.json` },
+    { template: 'plugin/README.md', target: `${pluginName}/README.md` },
     (answers.template === 'Yes') &&
       { template: 'plugin/templates/Example.js.ejs', target: `${pluginName}/templates/${name}Example.js` },
     (answers.command === 'Yes') &&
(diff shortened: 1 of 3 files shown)


packages/ignite/src/commands/plugin.js
```diff
@@ -81,10 +81,11 @@ const createNewPlugin = async (context) => {
     { template: 'plugin/index.js.ejs', target: `${pluginName}/index.js` },
     { template: 'plugin/ignite.toml.ejs', target: `${pluginName}/ignite.toml` },
     { template: 'plugin/package.json.ejs', target: `${pluginName}/package.json` },
+    { template: 'plugin/README.md', target: `${pluginName}/README.md` },
     (answers.template === 'Yes') &&
-      { template: 'plugin/templates/Example.js.ejs', target: `${pluginName}/templates/${name}Example.js` },
+      { template: 'plugin/templates/Example.js.ejs', target: `${pluginName}/templates/${name}Example.js.ejs` },
     (answers.command === 'Yes') &&
-      { template: 'plugin/commands/example.js.ejs', target: `${pluginName}/commands/example.js` }
+      { template: 'plugin/commands/example.js.ejs', target: `${pluginName}/commands/example.js.ejs` }
```
jamonholmgren: I don't think we should put the `.ejs` on this, correct @skellock ?

skellock: Oops.  Correct.  It should arrive as a js file, not a template.

robinheinze: Fixed!

PR #703, opened 2017-02-20
removes redux-logger from unholy

GantMan:
This PR removes redux-logger.   This isn't something we've been using much since the birth of Reactotron.   But don't worry!  It's not dead, I went ahead and made a plugin for it.

I also began our `PLUGINS.md` file to list the many plugins that will come.

So, should you want to add redux logger back into any Ignite project, you can do so with the instructions from here:  https://github.com/GantMan/ignite-redux-logger


GantMan:
image

PR #704, opened 2017-02-20
Changed packages/ignite to packages/ignite-cli

skellock:
Also:

* turns off the integration tests as they're not liking this change
* deleted the old instructions for how to get started


skellock:
I'm going to self merge this bad boy because I'll be working on the build scripts tonight.

And by build scripts, I mean `npm run publish` if all goes well.

PR #725, opened 2017-02-22
Removes __tests__ during an unholy ignite.

### LINKED ISSUE #724: Fix --skip-jest in the Unholy App
This was working.  Then I unfixed it.

skellock:
For Unholy, we use `ava` currently.  This deletes that `__tests__` directory.

If we do switch to `jest` it won't be before 2.0 launches.

Fixes #724


GantMan:
I merged in the wrong direction and broke your shiiiii

:( I'm sorry

skellock:
Np!  All good.  Surprised that git got caught with whitespace there to be honest.

PR #728, opened 2017-02-23
Adds plugin list, search, and info support to the CLI

jamonholmgren:
Adds support for the following plugin commands:

```
ignite plugin list
ignite plugin search <search string>
ignite plugin info <plugin name>
```

Also adds a directory that we can add to as we find out about and test plugins.

screen shot 2017-02-22 at 10 12 56 pm


packages/ignite-cli/src/commands/plugin.js
```diff
@@ -105,29 +106,110 @@ Generally, you would run this from ./YourApp/ignite/plugins/
 Commands:
   ignite plugin help
   ignite plugin new <your-plugin>
-  ignite plugin list (coming soon)
+  ignite plugin list
+  ignite plugin search <search string>
+  ignite plugin info <plugin name>
```
skellock: I'd like to promote these to `ignite list`, `ignite search`, and `ignite info` just like we already have `ignite add` and `ignite remove`?

Ignite 2 is really all about the plugins.

packages/ignite-cli/src/commands/plugin.js
```diff
@@ -105,29 +106,110 @@ Generally, you would run this from ./YourApp/ignite/plugins/
...
+  info(colors.highlight(`Ignite Plugin: '${name}'`))
+  newline()
+
+  if (plugin) {
+    table([
+      [ "Name", name ],
+      [ "Description", plugin.description ],
+      [ "Author", plugin.author ],
+      [ "URL", plugin.url ],
+      [ "Email", plugin.email ]
```
skellock: I'd vote to drop this field.

packages/ignite-cli/src/commands/plugin.js
```diff
@@ -105,29 +106,110 @@ Generally, you would run this from ./YourApp/ignite/plugins/
...
 
-  // context.print.debug(context.parameters)
   switch (parameters.second) {
     case 'new':
       await createNewPlugin(context)
       break
 
     case 'list':
-      print.error(`ignite plugin list is not implemented yet.`)
+      showPluginDirectory(context, null)
```
skellock: If we do end up keeping the `plugin` namespace, could we move this to a gluegun plugin?  It handles this switch statement and organizes your code at the same time.

jamonholmgren:
Good feedback @skellock -- we have some stuff to discuss on this. Good idea to bump back to 2.1 & let's table this until we've had a chance to chat about it.

packages/ignite-cli/src/commands/plugin.js
franzejr: I liked that!

franzejr:
Just to mention, I really liked this idea of searching for packages. I think it's necessary!

packages/ignite-cli/src/commands/plugin.js
maxfindel: Wouldn't it be better to use a centralized server like npm or bundler? That way you don't have to push code every time someone creates a cool plugin. Who's up for the challenge?

skellock: @ryanlntn is taking this task on now. 🎉

packages/ignite-cli/src/commands/info.js
```diff
@@ -0,0 +1,45 @@
...
+const validateName = require('../lib/validateName')
+const fetchPluginRegistry = require('./fetchPluginRegistry')
+
+/**
+ * Shows info about a particular plugin.
+ *
+ * @param {Object} context The gluegun context.
+ */
+const showPluginInfo = async function (context) {
+  console.log('I was called!')
```
skellock: Hey, I just logged you.
And this is crazy!
So here's my console.
So cull me maybe?

ryanlntn:
screen shot 2017-04-25 at 12 50 11 pm

I've moved the list, search, and info commands out of the plugin namespace and moved the plugin registry data to https://github.com/infinitered/ignite-plugins

screen shot 2017-04-25 at 12 53 43 pm
I'm thinking of simplifying list output to just the shortNames. Thoughts?

skellock:
@ryanlntn I like the short names for displaying. ya.  👍

Issue #740, opened 2017-02-26
Switch to Github for examples


skellock:
The weather service thing isn't working out.

I've used Github as an API example on Reactotron and it's really nice.

Let's switch the default examples over to that.

Issue #570, opened 2016-12-27
Mapview broken in Android base application


perqa:
<img width="289" alt="screen shot 2016-12-27 at 18 07 14" src="https://cloud.githubusercontent.com/assets/960796/21504495/c08a8d36-cc5f-11e6-94ad-7aba523ded90.png">

Application created by 

``ignite new TestApplication``

Everything seems to work, except Mapview on Android.

GantMan:
You can use our key for now:  
https://github.com/infinitered/ignite/blob/master/ignite-base/android/app/src/main/AndroidManifest.xml#L51-L54

I'll make sure this works in what I'm working on for Ignite 2.0

jamonholmgren:
This will be part of the `ignite-mapview` plugin, which I don't think has been started yet.

jamonholmgren:
Fixed in #741 I think?

jamonholmgren:
Okay, never mind -- this is maps, that was weather service.

jamonholmgren:
Moving to `ignite-maps`.

PR #741, opened 2017-02-26
[2.0] - move to Github API for examples

GantMan:
#### It's not fantastic like Reactotron, but it works!

**Changes include:**
* Using new API endpoint
* Updating Fixtures
* Updating sagas

**Issues:**   
* Could be a visual thing like in reactotron? 
* Feels Invasive - will fix when we do API as a plugin

closes: #740


jamonholmgren:
Should fix #570 too.

PR #743, opened 2017-02-26
Changes ignite plugin config to JSON.

### LINKED ISSUE #688: Convert toml to json for ignite plugin authors
We did this recently for the `ignite/ignite.toml` for apps, but now let's finish this up for `ignite.toml` for plugin authors.

Currently we only use this for defining generators, so it's only present in `ignite-basic-generators`.

The game plan is to call this `ignite.json`.  It'll be optional to have this.  #681 has that fix, so I'd like to get that merged before I do this one.

skellock:
Ignite plugins are now configured with `json` instead of `toml`.

Good bye TOML, we'll meet again in another project.

Fixes #688


GantMan:
is this backwards compatible?  Does this effect these guys? https://github.com/infinitered/ignite/blob/master/PLUGINS.md

skellock:
Ya, these can all just delete their TOML files.

GantMan:
but then how do they description though?

¯\\_(ツ)_/¯

skellock:
`package.json` in the `description` field.

skellock:
^ also, I may have just invented that just now.  I can't remember if it's thing or not.  😆

PR #744, opened 2017-02-26
Refactors the ignite extension.

### LINKED ISSUE #689: Seperate ignite extensions into individual files
https://github.com/infinitered/ignite/blob/next/packages/ignite/src/extensions/ignite.js grew up way too fast.

Time to chop it up into multiple files:

* findIgnitePlugins
* appConfig (flashbacks to my days as a .NET dev... shudder)
* addModule
* removeModule
* copyBatch
* addComponentExample
* removeComponentExample
* igniteConfig
* debugConfig
* generate

(see if we're still using this patchInFile... i don't think we are now)

skellock:
Once there lived a giant file and within, functions flourished.  That is, until the explorers came.  They did not know, nor understand what they were seeing.  And they were not pleased.  They demanded order and separation.

And they were right.

Thus began the journey of this PR. And it ends here.  Now.  When you press Merge.

Good luck champion.  The repo (and perhaps the universe) depends on you.

Fixes #689

### DIFF
--- a/packages/ignite-cli/src/extensions/ignite.js
+++ b/packages/ignite-cli/src/extensions/ignite.js
@@ -1,25 +1,23 @@
-// Steeeeeeeve I'm starting to not love this giant deconstruct - leaving it here to prove why
-const { dissoc, merge, pipe, prop, sortBy, propSatisfies, filter, reduce, flatten, map, takeLast, split, replace } = require('ramda')
-const { startsWith } = require('ramdasauce')
-const exitCodes = require('../lib/exitCodes')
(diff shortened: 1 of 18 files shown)


GantMan:
wow, I'm in love

skellock:
I was thinking (in the future), that we might be able to automatically assemble this at the filesystem level within gluegun.  

If your extensions have subfolders, then the files there will form 1 extension with each file as functions.

Seems pretty clear in hindsight.

jamonholmgren:
@skellock I actually assumed that was how gluegun worked when I was making the i18n plugin changes recently and was confused when it didn't work. Assuming we're thinking about the same thing.

PR #746, opened 2017-02-27
Updates ignite new to check/install react-native-cli

### LINKED ISSUE #612: Give friendly error if react-native CLI not installed
If `react-native-cli` is not installed, Ignite currently fails with a lame error. Let's make it less lame.

jamonholmgren:
Fixes #612.

When `react-native-cli` is not installed:

screen shot 2017-02-26 at 4 35 12 pm

When `react-native-cli` is already installed:

screen shot 2017-02-26 at 4 35 17 pm

### DIFF
--- a/packages/ignite-cli/src/extensions/reactNative.js
+++ b/packages/ignite-cli/src/extensions/reactNative.js
@@ -56,6 +56,17 @@ function attach (plugin, command, context) {
       rnOptions.push('--skip-jest')
     }
 
+    // install React Native CLI if it isn't found
+    const rncliSpinner = print.spin(`installing react-native-cli`)
+    try {
+      await system.exec('which react-native', { stdio: 'ignore' })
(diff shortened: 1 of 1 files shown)


packages/ignite-cli/src/extensions/reactNative.js
```diff
@@ -56,6 +56,17 @@ function attach (plugin, command, context) {
       rnOptions.push('--skip-jest')
     }
 
+    // install React Native CLI if it isn't found
+    const rncliSpinner = print.spin(`installing react-native-cli`)
+    try {
+      await system.exec('which react-native', { stdio: 'ignore' })
```
skellock: We also have `system.which('react-native')` as well.

jamonholmgren: TIL, thanks.

PR #748, opened 2017-02-27
Adds temporary spinner during Unholy install

### LINKED ISSUE #717: nothing happening in install in some spots
I only caught a second of it.  But there are long pauses of nothing

nothing

jamonholmgren:
We'll need to do the same for other templates when we expand those, but this makes the standard unholy app template look better during `ignite new`. Fixes #717.

ignite-new2

### DIFF
--- a/packages/ignite-unholy-app-template/index.js
+++ b/packages/ignite-unholy-app-template/index.js
@@ -1,5 +1,12 @@
 const { merge, pipe, assoc, omit, __ } = require('ramda')
 
+// Hack to get around `ignite add <x>` spin-up delay
+const addTemporarySpinner = (package, spinner) => {
+  spinner.text = `initializing ${package}`
+  spinner.start()
+  setTimeout(() => spinner.stop(), 8000)
+}
+
 // Questions to ask during install if the chatty route is chosen.
 const installQuestions = [
   {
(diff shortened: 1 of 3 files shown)


packages/ignite-unholy-app-template/index.js
```diff
@@ -1,5 +1,12 @@
 const { merge, pipe, assoc, omit, __ } = require('ramda')
 
+// Hack to get around `ignite add <x>` spin-up delay
+const addTemporarySpinner = (package, spinner) => {
+  spinner.text = `initializing ${package}`
+  spinner.start()
+  setTimeout(() => spinner.stop(), 8000)
```
jamonholmgren: It's almost exactly 9 seconds on my machine.

jamonholmgren:
@skellock It does start the spinner before the npm install command. But it takes about 9 seconds on my machine before it even hits that line of code. The issue is that the spinup time for the new `ignite add` process is slow.

jamonholmgren:
Okay, I feel pretty lame now -- an hour wasted on it, when you're absolutely right, @skellock.

Now, is there any reason we really need to protect the spinner startup?

skellock:
I can't remember exactly atm, but I think it had something to do with deciding what to show when it gets installed via `new` or directly when using `add`.

Don't feel lame dude.  I've spent more than a few hours on this stuff, and it's still not pretty.

PR #751, opened 2017-02-27
Fixes spinner during ignite new

### LINKED ISSUE #717: nothing happening in install in some spots

jamonholmgren:
Better fix than #748, thanks to @skellock . Should fix #717. Is there any reason why we need the guard around these lines of code? I can't find any.


jamonholmgren:
@GantMan You might need to show me why this block of code is necessary...if it is. The output seems good now.

skellock:
I'm ok with this.  I thought maybe it had something to do with when you run `add` directly.  I can't remember anymore.  It's getting blurry. haha. 

Last one in wins.

PR #753, opened 2017-02-27
Unholy -> Blessed

### LINKED ISSUE #745: Rename "Unholy" template to "Infinite Red"
Up until now, the "Unholy" template has been sort of a kitchen sink approach, where we put sample code of the decisions we've made so far.

In reality, "Unholy" is the stack that Infinite Red currently thinks is the best way to build React Native projects. Given that we have a lot of experience spinning up greenfield apps, our decisions are pretty informed, but they are of course opinions.

I'd like to rename "Unholy" to "Infinite Red" to make it more clear what this set of opinions represents. We build apps a certain way at Infinite Red and are proud of the work we've done, but there are a lot of very viable alternatives out there. If you want to sort of go down the "Blessed" IR route, then that's why the "Infinite Red" template will exist. Ignite 2 lets you use that or any other custom or preconfigured set of decisions.

cc @GantMan @skellock @robinheinze @derekgreenberg

jamonholmgren:
Renames the unholy app template to `infinite-red-app-template`. Resolves #745.


packages/ignite-infinite-red-app-template/index.js
```diff
@@ -36,7 +36,7 @@ const add = async function (context) {
   const { filesystem, parameters, ignite, reactNative, print, system, prompt, template } = context
   const name = parameters.third
   const igniteDevPackagePrefix = parameters.options['ignite-dev-package-prefix'] // NOTE(steve): going away soon
-  const spinner = print.spin(`using Infinite Red's ${print.colors.cyan('unholy')} app template`).succeed()
+  const spinner = print.spin(`using the ${print.colors.red('Infinite Red')} app template`).succeed()
```
skellock: lol

PR #769, opened 2017-02-27
Implements auto-running of generated minimal app tests

jamonholmgren:
Also fixes some issues with the generated minimal app that were uncovered by these tests.


packages/ignite-cli/src/templates/plugin/package.json.ejs
```diff
@@ -4,9 +4,5 @@
   "version": "0.0.1",
   "license": "MIT",
   "author": "Your Name",
-  "url": "https://example.com/path/to/your/plugin",
-  "devEngines": {
-    "node": ">=7.x",
-    "npm": ">=4.x"
-  }
+  "url": "https://example.com/path/to/your/plugin"
```
jamonholmgren: Removed the devEngines parameter -- per @skellock

PR #772, opened 2017-02-28
Plugin Directory Love

skellock:
## Ignite Plugin Authors Are A Go

You can now run `ignite add elements` to install 3rd party plugins. 

If it's published on npm, you can install it.  Today.

Most of ours plugins aren't there yet, but this is the 👍 to start if they have their proper home & version number (spoiler, ours don't just yet, but other people can!! @dabit3 already has 2 of 'em and @GantMan has 1 or 2, and i believe @robinheinze has one too for radio buttons?).

## Overriding Plugins

Now, because we're still developing, this PR introduces another thing: **the ability to override plugin installs.** 🌹 

Lets say you're building `ignite-maps` like @jamonholmgren is.  End users will install with:

```sh
ignite add maps
```

But you don't want to publish that first while you're testing.  If you're building this in `/Users/jamonsux/ignite/maps`, You can:

```sh
export IGNITE_PLUGIN_PATH=/Users/jamonsux/ignite
```

Any subdirectory now in there will become the first thing `ignite add` will check before going out to npm.

🚨  I highly recommend you add: `export IGNITE_PLUGIN_PATH=/path/to/your/ignite/repo/packages` to the shell of your choice. <sub>the correct answer is zsh.</sub>

In fact, you HAVE TO once this gets merged.  In a few days when we go beta, we can just remove that ENV variable. 🚨 

## Prettier Things For Prettier People

This PR also pretties up the error message when we can't `ignite add` something properly.  This was killing Gant inside.  I could see it in his eyes.

image

## Getting Started Docs

It also updates the docs on how to get started with this alpha.

Thanks, and sorry the dev experience has been bumpy so far... it'll get better when we're deploying to npm.  This is that first step.

## Spammy CCs

Just wanted to make sure you read this PR because your 💩 is about to break if you don't set that ENV var once this lands.

@jamonholmgren gets a /cc
@mlaco get a /cc
@robinheinze gets a /cc
@GantMan was already mentioned.  no /cc for you.
@codeithuman gets a /cc
@derekgreenberg gets a /cc cuz i miss him
and I forgot @kevinvangelder , sorry @kevinvangelder ....  you get a /cc

### DIFF
--- a/docs/advanced-guides/using-development-build.md
+++ b/docs/advanced-guides/using-development-build.md
@@ -6,23 +6,33 @@ This guide will walk you through how to setup and use a pre-release or experimen
 
 * Node 7.6+
 * NPM 4 (ships with Node 7)
+* Yarn 20.3+
 
-> Note: Requirements may change based on the Ignite version you are targeting.
+If you're on a Mac, you can use one of these techniques:
+
+* [n](https://github.com/tj/n) - with `npm i -g n && n 7.6`
+* [homebrew](https://brew.sh/) - with `brew install node`
+* [nvm](https://github.com/creationix/nvm) - with `nvm install v7.6.0`
+
 
 ## Getting Started
 
+#### Out With The Old
+
 First, if you already have a copy of Ignite installed with `npm`, uninstall it with the commands listed below.
 
-For Ignite 2.0 and above:
+For Ignite 2.0 and above: (exact one depends on what you may have installed previously)
 ```sh
-npm rm -g react-native-ignite
+npm rm -g ignite
+npm rm -g ignite-cli
 ```
 
 For Ignite 1.x:
 ```sh
 npm rm -g react-native-ignite
-rm /usr/local/bin/ignite && rm /usr/local/lib/node_modules/ignite
 ```
+
+#### In With The New
 
 Clone Ignite's repository from https://github.com/infinitered/ignite to your local development machine.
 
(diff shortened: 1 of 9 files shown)


docs/advanced-guides/using-development-build.md
```diff
@@ -67,3 +95,31 @@ ignite new MyApp
...
+
+
+
+
+
+
+
+
+
+
```
jamonholmgren: holy \n batman

packages/ignite-cli/src/commands/add.js
```diff
@@ -125,7 +139,24 @@ Examples:
...
+        print.info('')
+        print.info(print.colors.muted('-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-'))
+        print.error('  We also searched in these directories:\n')
+        ignite.pluginOverrides.forEach(dir => {
+          print.info(`    ▸ ${dir}`)
+        })
+        print.error('')
+        print.error('  During Ignite 2 alpha, please consider setting this\n  environment variable:\n')
+        print.info(`  export IGNITE_PLUGIN_PATH=${path.resolve(__dirname, '..', '..', '..')}`)
+        print.info(print.colors.muted('-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-'))
```
jamonholmgren: This error message is brilliant.

docs/advanced-guides/using-development-build.md
skellock: whoa...

codeithuman:
Well done sir!

PR #787, opened 2017-02-28
Adds ignite remove <plugin> --confirm

jamonholmgren:
This will be especially useful when we do CI, checking to see if a plugin cleans up properly after itself.

### DIFF
--- a/docs/quick-start/ignite-commands.md
+++ b/docs/quick-start/ignite-commands.md
@@ -98,10 +98,11 @@ command or generator.
 ### Remove
 
 ```
-ignite remove {ignite plugin name}
+ignite remove {ignite plugin name} [--confirm]
 ```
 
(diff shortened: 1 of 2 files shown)


skellock:
Would you be ok calling this -f and --force?

jamonholmgren:
@skellock Actually yes ... believe it or not I scoured the web trying to figure out what the convention was for this sort of thing and couldn't find anything. I know `-i` means `interactive`.... Happy to change it if you'd like me to.

skellock:
I've seen -y used as well.  As in `npm init -y` to answer "yes" to all questions.

I was thinking `-f` like some unix commands.

It's totally up to you.

jamonholmgren:
Looks like `apt-get` and `yum` also use `-y`.

skellock:
I love how you research things like this.

PR #797, opened 2017-02-28
Relocates ignite-i18n.

### LINKED ISSUE #791: Split ignite-i18n into its own repo
Per #759, I would like to split this plugin into its own repo. We can host it on our infinitered Github org, just need someone to make the necessary changes. @robinheinze, want to do this one? (lower priority though)

skellock:
Moves this plugin to [its own repo](https://github.com/infinitered/ignite-i18n).

Fixes #791


skellock:
That inline resolve editor in github is pretty slick.  💯

Issue #758, opened 2017-02-27
Yarn version check (0.20+?)


GantMan:
![image](https://cloud.githubusercontent.com/assets/997157/23368428/6f825352-fcd3-11e6-9829-f3aa7c667d60.png)

### Upgrade
![image](https://cloud.githubusercontent.com/assets/997157/23368466/8e5cd66c-fcd3-11e6-999a-8b8fa3cc9b59.png)

skellock:
i think i saw somewhere (create-react-app source code?) where they redirect stderr to /dev/null.. i wonder if thats where the `npm cache dir` version number line comes from?  I dunno... the easier way is to just say `yarn@>=0.20` or whatever.

PR #808, opened 2017-03-02
Requires Yarn 0.20.0+, if installed.

jamonholmgren:
I'm not exactly sure if this is what @skellock was going for in #758, but I think it was?

screen shot 2017-03-02 at 12 18 13 am

cc @GantMan


kongpop113:
I have a very similar issue.  

capture

jamonholmgren:
@kongpop113 Hmm, let me see if @kevinvangelder can look into that. He's our resident Windows expert.

Issue #752, opened 2017-02-27
CLI output - remove header after first usage on a project?


jamonholmgren:
When you jump into a project and type in `ignite`, you'll get this:

![screen shot 2017-02-26 at 7 43 45 pm](https://cloud.githubusercontent.com/assets/1479215/23348082/f8d20d3c-fc5b-11e6-87f6-c158fb3f2506.png)

This is cool looking. However, on subsequent `ignite` calls, I'd like to discuss suppressing the header output.  That way it looks more like this:

![screen shot 2017-02-26 at 7 45 20 pm](https://cloud.githubusercontent.com/assets/1479215/23348110/2038b826-fc5c-11e6-82c6-e0ae16232fda.png)

Perhaps we add an entry into the `ignite/ignite.json` file to indicate if it's been shown.

Low priority, just something I thought of tonight.

skellock:
Yay!  I agree!!!  

I thought it was just me.  When you guys started making fun of me for being "not a fan of branding", I was just going to keep my mouth shut.  Haha.

jamonholmgren:
An easier way to do this is only output the header if we are not in an Ignite project.

PR #812, opened 2017-03-02
Adds ignite --help and modifies the output slightly.

jamonholmgren:
More judicious use of the Ignite header too.

screen shot 2017-03-02 at 1 10 08 pm


jamonholmgren:
Fixes #752.

PR #813, opened 2017-03-02
Minimal template returns to its original home

skellock:
So once upon a time, we had this.  @jamonholmgren even did the min/max argument refactoring.

Then, the evil @skellock, Destroyer of Repos swooped in from the north and tore things apart.

He had realized weeks later that this was wrong.  Min and Max deserve to live together.  And this is his attempt to remedy his unjust actions.

Forgive me.


packages/ignite-integration-tests/test/ignite-new/newMin.test.js
```diff
@@ -1,79 +0,0 @@
-const test = require('ava')
```
jamonholmgren: I'm super sad to see this integration test go. Can we get a `--max` integration test instead?

skellock: Oh sorry... ya i should have said something.

This is going to move over to `ignite-ir-app`.   

It'll come back strong, faster, and more glorious than before.

I removed it because `--max` now comes with husky support (which is git hooks).  And when you have .git inside .git, things don't work well.   We might have to move the running directory into a `/tmp` directory.  

I didn't want to deal with that change right now, so I panicked and hit the delete button instead.

PR #825, opened 2017-03-03
Removes flow from generated app

jamonholmgren:
Removes Flow.

The generated app seems to work fine, but I am getting these warnings. Any suggestions? @GantMan 

screen shot 2017-03-03 at 12 02 57 am

Fixes #536.

Issue #798, opened 2017-02-28
Rename a few projects


skellock:
Since these generators rely on `ignite-infinite-red-app-template`, anyone object to me renaming this to `ignite-infinite-red-generators`?

/cc @GantMan + @jamonholmgren

jamonholmgren:
I thought this over this morning. I think I'd rather keep the current name in this case. While the IR-AT is definitely all Infinite Red opinions, these generated components and whatnot are less opinionated. And Ignite needs some basic generators as something to default to. I'd vote to keep as-is.

skellock:
`ListView` and `Component` are almost opinion free.  The other are full-on Infinite Red.

Split 'em up?

I like the idea of creating `ignite-infinite-red-generators` because people who create their own boilerplates will be doing the same.  It'd be nice to show folks an example of where and why.

If we have some core generators, we'll get people contributing more.  That way we can build up a library of things everyone needs.  Just like we're doing with plugins.

skellock:
That last message was very cryptic.  I was typing it mid-meeting.

Here's what we decided:

* `ignite-minimal-app-template` will disappear.  i've just rolled that into `ignite-infinite-red-app-template`.
* `ignite-infinite-red-app-template` is now called `ignite-ir-app`.
* we're recommending people call these things `ignite-{boilerplate-name}-app`.
* `ignite-empty-app-template` is going to be dismantled and run entirely from inside `new.js`.
* `ignite-basic-generators` is going to be rolled into `ignite-ir-app`.
* `ignite-dev-screens` will remain here, but will be stripped of all `ir` opinions and hooks.

jamonholmgren:
Update: `ignite-ir-app` will be renamed to `ignite-ir-boilerplate` and we will have a slightly different model for boilerplate plugins.

jamonholmgren:
I re-opened #819 to track the final thing that remains here. Closing this.

PR #827, opened 2017-03-03
Boilerplates become first class citizen

### LINKED ISSUE #818: Bring ignite-basic-generators into ignite-ir-app
These generators are very opinionated (well most are), but they belong to the Infinite Red stack.  Let's move these inside `ignite-ir-app`.

What makes this a bit funky is that the `ignite-ir-app` will now have to include itself as a dependency within it's `index.js`.  Not a huge deal, but we can optimize this later.

skellock:
Boilerplates are still ignite plugins, but they have some properties that make them unique.

In order to not become confused with plugins, there's now a `boilerplate.js` file which is expected to exist.  This will have the an interface of:

```js
async function install (context) {
  // do  your thing.
}

module.exports = { install }
```

This will also change the interface of the CLI a bit for advanced users.

```sh
# this remains the same
ignite new MyInteractiveApp
ignite new MyPimpedOutApp --max
ignite new MySlimmerApp --min

# but this changes
ignite new MyCustomApp --boilerplate ignite-wacky-boilerplate

# and so does this
ignite new MyUninspiredApp --no-boilerplate
```

Internally, a few things have changed around.  `new.js` now calls `boilerplate-install.js` instead of `add.js`.

Also, this changes `ignite-infinite-red-app-template` to `ignite-ir-boilerplate` and moves `ignite-basic-generators` into `ignite-ir-boilerplate` as they're tightly coupled.

Fixes #818 
Addresses the rest of the critical issues with #798 
Totally screws up #825


skellock:
Might make sense to rename `index.js` to `plugin.js` in light of these changes.

docs/quick-start/ignite-commands.md
```diff
@@ -57,30 +57,33 @@ generates.
...
 Generate a new React Native project with Ignite.
 
 `ignite n` uses `react-native init`, then adds files specific to Ignite.
-  
+
 When you execute this command, Ignite 2.0 will give you a series of options for
 your new project regarding what libraries you would like to use.
 
-`ignite new` has various -- commands. ex:
+With `ignite new`, you have the option to pick your own boilerplate to install for your project.  The default is `ignite-ir-boilerplate`, however you can change this by providing your own boilerplate available on `npm`.
```
jamonholmgren: -2016

docs/quick-start/using-boilerplates.md
```diff
@@ -0,0 +1,15 @@
...
+You can use an boilerplate as the starting point for your app by creating a new
+Ignite project with the `--boilerplate` flag:
+
+```
+ignite new MyIgniteApp --boilerplate {boilerplate name}
+```
+
+Ignite comes with a default boilerplate.
+
+It creates your app with a host of opinions and options. This boilerplate reflects the way that we at Infinite Red prefer to start our apps. This is also Ignite's default boilerplate; if you don't select a one, Ignite will use this.
```
jamonholmgren: Add verbiage about beyond 2016

PR #832, opened 2017-03-04
Adds boilerplate to plugin generator - fixes #829

### LINKED ISSUE #829: Refactor the plugin generator to allow for generating boilerplate plugins easily
Per #824 and #827, we now need to update the plugin generator to generate boilerplate plugins.

On `ignite plugin new mything`, we'll ask if the plugin will be a boilerplate plugin. Answering Yes will generate a `boilerplate.js` file and `boilerplate` template folder.

jamonholmgren:
To test this, I spun up a new plugin:

```
$ ignite plugin new ignite-test-boilerplate
$ export IGNITE_PLUGIN_PATH=$(cwd)
$ ignite new Test -b ignite-test-boilerplate
```

It works!


packages/ignite-cli/src/templates/plugin/boilerplate.js.ejs
```diff
@@ -0,0 +1,110 @@
+/**
```
jamonholmgren: This is the most minimal boilerplate generator script I could come up with that still had enough functionality. We might want to refactor some of this out into boilerplate helper functions at some point.

packages/ignite-cli/src/commands/plugin.js
```diff
@@ -20,8 +20,14 @@ const walkthrough = async (context) => {
   // Okay, we'll ask one by one, fine
   return await context.prompt.ask([
     {
+      name: 'boilerplate',
+      message: 'Is this an app boilerplate plugin?',
```
GantMan: Don't hate me, but I think this will be more confusing than helpful :/

I think we can include this, but skip the question and move to a documented flag?

jamonholmgren: You're probably right. I like having a question, can you think of another wording that might make sense?

jamonholmgren: This will be well documented in our guides, btw.

PR #838, opened 2017-03-04
Temporary fix for mock-fs.

skellock:
Until mock-fs accepts that patch, this will hook us up.

This is what eslint did (https://github.com/eslint/eslint/commit/0fd9d2712aeeebc65399083f5249729da91c7bf6) as a temporary work around.

I've tested this on 7.6 and 7.7.1.


jamonholmgren:
NICE!

PR #848, opened 2017-03-05
fix proptypes warning

GantMan:
Fixes these two warnings:
image

image


packages/ignite-dev-screens/templates/ButtonBox.js
```diff
@@ -5,8 +5,8 @@ import styles from './Styles/ButtonBoxStyles'
 export default class ButtonBox extends React.Component {
   static propTypes = {
     onPress: PropTypes.func,
-    image: PropTypes.string,
-    style: PropTypes.object,
+    image: PropTypes.number,
+    style: PropTypes.number,
```
skellock: could we make this one number or object?

GantMan: would that code be `style: PropTypes.number || Proptypes.object,` ??

GantMan: nm, it's `style: PropTypes.oneOfType([PropTypes.object, PropTypes.number]),`

PR #851, opened 2017-03-05
Changes createdBy to boilerplate to match naming scheme

jamonholmgren:
Tiny change to generated boilerplate plugins.


jamonholmgren:
You're right, I think I got those spots.

Issue #864, opened 2017-03-05
Git config difference causes infinite spinner


GantMan:
![image](https://cloud.githubusercontent.com/assets/997157/23589728/f6693c5a-0197-11e7-9260-692d979bcb8f.png)

Igniting a new project, I get this ^

My EOL crlf config is set to force unix style in git.   I don't know when/how I set this, but man!  this is some helluva experience if you have this set.

I can turn it off, [as shown here](http://stackoverflow.com/questions/20168639/git-commit-get-fatal-error-fatal-crlf-would-be-replaced-by-lf-in), but I don't know if everyone will accept that answer.

PR #865, opened 2017-03-05
added check with fail + cleanup

GantMan:
This should help people with their configs set like me.  I think the real fix needs to come from upstream, which I've filed a ticket for.

image

This PR pertains to #864 
It should only affect a small portion of users.  Might frustrate them, but it's better than making the error fall on us.

Lagniappe:  I fixed the return codes and punctuation along the way.

### DIFF
--- a/packages/ignite-cli/src/cli/check.js
+++ b/packages/ignite-cli/src/cli/check.js
@@ -20,7 +20,7 @@ let angryGit = Shell.exec('git config --global core.safecrlf', {silent: true})
 if (angryGit.stdout === 'true\n') {
   console.error(`
     Git safecrlf with the current version of react-native is likely to fail.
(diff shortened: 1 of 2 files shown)


skellock:
Would you be ok with appending this to `.gitattributes` instead?

```
*.bat	text eol=crlf
```

GantMan:
I believe it is several files, tbh.   I think a `.gitattributes` update should be upstreamed instead of appended here.

skellock:
I agree it should be upstream.  How long though?  Month?  Months?   A few lines in `.gitattributes` mean people with this super-strict check enabled can use Ignite today.

Can we at least amend the message to explain how to fix this by placing that fix in `.gitattributes`?

GantMan:
https://github.com/infinitered/ignite-ir-boilerplate-2016/pull/5 instead.

Issue #862, opened 2017-03-05
Nothing is generated on Windows 10


johntom:
Hi
Installed version 2 on win10 pro and checked all dependecies. With exception of maps I had ignite 1 sample running using genymotion. When I run ignite new PizzaApp nothing is generated in folder. Note: I also posted on slack,
Thanks
John

skellock:
Thanks for the report John.  We'll take a look at running from Windows.  Probably something small I overlooked.

johntom:
I'm looking to move and app from Native Starter Pro(4.0) and already find the support community (slack and github) a notch above. Also I just found out about the following  and tracking now.

C:\React\ignite2>ignite doctor
System
  platform           win32
  arch               x64
  cpu                4 cores            Intel(R) Core(TM) i5-4460  CPU @ 3.20GHz
  directory          C:\React\ignite2

JavaScript
  node               7.7.1        C:\PROGRAM FILES\NODEJS\NODE.EXE
  npm                4.1.2        C:\PROGRAM FILES\NODEJS\NPM.CMD
  yarn               0.21.3       C:\USERS\JOHN\APPDATA\ROAMING\NPM\YARN.CMD

React Native
  react-native-cli   2.0.1

Ignite
  ignite             2.0.0-beta.1   C:\USERS\JOHN\APPDATA\ROAMING\NPM\IGNITE.CMD
vvv -----[ DEBUG ]----- vvv
{ Error: Command failed: java -version
Error: Registry key 'Software\JavaSoft\Java Runtime Environment'\CurrentVersion'
has value '1.8', but '1.7' is required.
Error: could not find java.dll
Error: Could not find Java SE Runtime Environment.

    at ChildProcess.exithandler (child_process.js:210:12)
    at emitTwo (events.js:106:13)
    at ChildProcess.emit (events.js:194:7)
    at maybeClose (internal/child_process.js:899:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:226:5)
  killed: false,
  code: 2,
  signal: null,
  cmd: 'java -version',
  stderr: 'Error: Registry key \'Software\\JavaSoft\\Java Runtime Environment\'\\CurrentVersion\'\r\nhas value \'1.8\', but \'1.7\' is required.\r\nError: could not find java.dll\r\nError: Could not find Java SE Runtime Environment.\r\n' }
^^^ -----[ DEBUG ]----- ^^^

skellock:
[...]

PR #873, opened 2017-03-06
We love Windows too!

skellock:
Just kidding, no we don't.  Only @kevinvangelder does.  

But!!! We will support it because it's damn popular!

So, some of the commands we shell out and don't work on windows.  

Specifically `react-native link` (again... we had issues with this on the mac which is why we had to switch to `stdio: ignore`).

This PR bumps gluegun as I've added `cross-spawn` to power `context.system.spawn` now.  It works the same as Node's spawn but ensures things work properly with Windows.

I don't know the who story, but apparently grandchild processes don't inherit the path, and there's some issues with escaping with the out-of-the-box node spawn experience.

Cross-spawn makes this patching invisible to us.  👍 

Address #862 .  I won't say "fixed" just yet, but I will say, it's working on my machine.  Will push a beta 2 once this is merged and circle back with the folks in that ticket to follow up.

### DIFF
--- a/packages/ignite-cli/src/extensions/ignite/addModule.js
+++ b/packages/ignite-cli/src/extensions/ignite/addModule.js
@@ -23,10 +23,14 @@ module.exports = (plugin, command, context) => {
     // install the module
     if (useYarn) {
       const addSwitch = options.dev ? '--dev' : ''
-      await system.run(`yarn add ${moduleFullName} ${addSwitch}`)
+      const cmd = `yarn add ${moduleFullName} ${addSwitch}`
+      ignite.log(cmd)
+      await system.spawn(cmd, { stdio: 'ignore' })
     } else {
       const installSwitch = options.dev ? '--save-dev' : '--save'
-      await system.run(`npm i ${moduleFullName} ${installSwitch}`)
+      const cmd = `npm i ${moduleFullName} ${installSwitch}`
+      ignite.log(cmd)
+      await system.spawn(cmd, { stdio: 'ignore' })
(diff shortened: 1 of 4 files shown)


packages/ignite-cli/src/extensions/ignite/addModule.js
```diff
@@ -23,10 +23,14 @@ module.exports = (plugin, command, context) => {
     // install the module
     if (useYarn) {
       const addSwitch = options.dev ? '--dev' : ''
-      await system.run(`yarn add ${moduleFullName} ${addSwitch}`)
+      const cmd = `yarn add ${moduleFullName} ${addSwitch}`
+      ignite.log(cmd)
+      await system.spawn(cmd, { stdio: 'ignore' })
```
GantMan: maybe we can wrap `spawn` with {stdio: 'ignore} and call it `runSafe`.  Perhaps make it the prescribed way to run commands in gluegun, with a short explanation why?  I could see this being an issue for anyone who uses gluegun to use the wrong path.

skellock: It's not safe though. It's `ignore`.  I have no clue what `react-native link` is doing that would cause this, and I've looked. 😆 

Now that we have `spawn`, `exec` and `run`, I do agree this needs order.

That was my intention behind `run` originally.  Maybe we switch to `run` and `runAdvanced`?  

`run` would just return `stdout`.  And `runAdvanced` could return an object like `spawn` does if you need extra details?

Issue #882, opened 2017-03-07
`ignite init` fails for dashed-project-names


chapati23:
**What's going on?**
The failure itself happens during `react-native init`, i suppose. When you try `react-native init demo-app` you'll get an error message telling you that project names can only have alphanumeric characters.

---
**Steps to reproduce**
* Try running `ignite new demo-app`
* Watch it fail with no idea what's wrong

---
**Proposed solution**
* Forward the error message from `react-native init` so people don't have to walk through the source code to find out what's wrong :)

skellock:
Agreed!  Thanks for opening this.

skellock:
Related to #874

PR #902, opened 2017-03-15
Check for kebab case project names and return useful error

fvonhoven:
## Describe your PR
React Native doesn't allow kebab case project names, and neither will we...
Check for kebab case (dashes) project naming when a user is igniting a new project, and return a useful error if found.
kebab-error


skellock:
Fixes #882

Issue #874, opened 2017-03-06
Add sanity checks before running new


skellock:
We have `ignite doctor` yay!

But...

We need to run environmental checks before allowing people to run `new`.

So far I've seen people have issues with `react-native-cli`, `yarn` and `xcode 7`.

Let's check these things silently before allowing them to spit out an app.

* [ ] Check presence of react-native & error if it's installed
* [ ] Check presence of npm 2, upgrade to npm 4 after installing ignite (see [below comment](https://github.com/infinitered/ignite/issues/874#issuecomment-285345838))

skellock:
Another thing to add to the list.  The presence of `react-native` installed globally will trip up some steps but not others.  :(

skellock:
Also, people who have npm 2 installed, then upgrade to npm 4 after installing ignite will have issues.  We can detect this by looking for the `gluegun` dependency underneath the `node_modules` in the `ignite-cli` directory.

skellock:
Come to think of it, that's exactly what [should be happening](https://github.com/infinitered/ignite/blob/master/packages/ignite-cli/package.json#L52).  Investigate why this isn't working?

skellock:
Might want to consider #883 detection as well if it turns out to be an issue.

skellock:
Bumping up to higher priority as this is devouring lots of hours.

jamonholmgren:
PR on the way for checking presence of `react-native`.

skellock:
I did a check for `react-native` already.  We still need a few others still.

jamonholmgren:
Should be good to go on this one.

PR #934, opened 2017-03-23
Adds sanity check for presence of `react-native` conflict

jamonholmgren:
screen shot 2017-03-22 at 11 36 35 pm

Ref #874.

### DIFF
--- a/packages/ignite-cli/src/cli/check.js
+++ b/packages/ignite-cli/src/cli/check.js
@@ -1,7 +1,7 @@
 // NOTE:  this file is intentionally written with es3
 
-// check the node version
 var Sniff = require('gluegun/sniff')
+var shell = require('shelljs')
 
 // check the node version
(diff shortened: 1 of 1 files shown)


packages/ignite-cli/src/cli/check.js
```diff
@@ -30,12 +30,27 @@ var rnCli = enforceGlobalDependency({
   optional: false,
   range: '>=2.0.0',
   which: 'react-native',
-  package: 'react-native-cli',
+  packageName: 'react-native-cli',
```
jamonholmgren: Noticed this was missed from my previous #929 PR.

skellock:
Since we are targetting es3, @jamonholmgren we will have to remove the multilines in that console.log.  Unless you can test that in node 4.  If its good, we can add an linter ignore directive.

jamonholmgren:
Fixes #940 too

PR #972, opened 2017-04-14
References proper react-native binary - Fixes #949

### LINKED ISSUE #949: Error: not found: yarn
Hello,
On Win7 I followed the guide to install ignite, and I did it, but when I try to create a new project I get this error:

Error: not found: yarn

and

(node:5520) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): Error: not found: yarn

Should I install yarn? It does not say anywhere.

jamonholmgren:
Fixes #949

[According to @skellock](https://github.com/infinitered/ignite/issues/949#issuecomment-291283847):

> when we shell out, node is giving precedence to the node_modules/.bin/react-native executable instead of the global one on the path.

This change calls `$(which react-native) --version` rather than `react-native --version` which should address the problem in issue #949 and others.

### DIFF
--- a/packages/ignite-cli/src/cli/check.js
+++ b/packages/ignite-cli/src/cli/check.js
@@ -31,7 +31,7 @@ var rnCli = enforceGlobalDependency({
   range: '>=2.0.0',
   which: 'react-native',
   packageName: 'react-native-cli',
-  versionCommand: '$(which react-native) --version',
+  versionCommand: '--version',
   installMessage: 'To install: npm i -g react-native-cli'
 })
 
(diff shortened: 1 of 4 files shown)


skellock:
https://github.com/npm/node-which  <-- this bad boy, right thurrr...

jamonholmgren:
On it.

packages/ignite-cli/src/cli/check.js
```diff
@@ -31,7 +31,7 @@ var rnCli = enforceGlobalDependency({
   range: '>=2.0.0',
   which: 'react-native',
   packageName: 'react-native-cli',
-  versionCommand: 'react-native --version',
+  versionCommand: '--version',
```
jamonholmgren: The `versionCommand` now only takes the actual version flag, not the whole command. This is because we find the executable using the new `which` call.

jamonholmgren:
@skellock @GantMan Should be good to go now.

PR #976, opened 2017-04-17
Add clarification about final npm install command after saying to install Yarn

fvonhoven:
## Describe your PR
Added note about Yarn global not working on Windows.
I thought it was strange that we say to install Yarn, and then the final install command uses npm.
This adds a little clarification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,13 +82,11 @@ You can check your version of node by running
 node -v
 ```
 
-_Currently, Yarn global doesn't work on Windows - so feel free to skip installing Yarn if you're running on Windows_
-_...otherwise..._
-
-install [Yarn](https://yarnpkg.com/lang/en/docs/install/) for your system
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -82,7 +82,10 @@ You can check your version of node by running
 node -v
 ```
 
-then install [Yarn](https://yarnpkg.com/lang/en/docs/install/) for your system
+_Currently, Yarn global doesn't work on Windows - so feel free to skip installing Yarn if you're running on Windows_
```
skellock: Yarn global doesn't work on windows?

GantMan: Perhaps: "All global installs below are using `npm` over `yarn` because of [this bug](https://github.com/yarnpkg/yarn/issues/859) on Windows, which is known for global installs with Yarn."

PR #987, opened 2017-04-21
Update calls to insertInFile to patchInFile

ryanlntn:

Next step will be moving `patchInFile` up into gluegun.


skellock:
@ryanlntn I can show you how that's done next week if ya like?  Would only take like 10 mins.

ryanlntn:
@skellock Awesome!

PR #989, opened 2017-04-24
Adds Tour of Ignite documentation, updates other docs

jamonholmgren:
Also removes old testRelease.sh script


docs/advanced-guides/tour.md
```diff
@@ -0,0 +1,70 @@
...
+
+## gluegun
+
+`ignite-cli` (described [next](#ignite-cli)) uses another Infinite Red npm package called [gluegun](https://github.com/infinitered/gluegun). Gluegun is sort of like [Yeoman](http://yeoman.io/) in that it gives us a bunch of CLI tools, but it's much lighter and doesn't take over your whole CLI.
+
+Whenever you see `context` in `ignite-cli`, it is a gluegun runtime context. This is a JS object that has a bunch of useful functions and properties attached to it, like `print` and `parameters`.
+
+In addition, Ignite adds some of its own Ignite-specific functions and parameters to `context` ... those are found under `context.ignite`.
+
+If you're curious what's in the context object, check out Gluegun's [nice context docs](https://infinitered.github.io/gluegun/#/context-api.md), or just inspect it with `console.dir(context, {depth: 3, colors true})`.
```
skellock: woah.  i never knew about `depth: 3, colors: true`.

docs/advanced-guides/tour.md
```diff
@@ -0,0 +1,70 @@
...
+  const { print, filesystem, prompt, ignite, parameters, strings } = context
+  // ...
+}
+```
+
+... then you're looking at the main function. The `context` object is described above in the [gluegun section](#gluegun). Here, we're destructuring several parts of `context` such as `print`, `filesystem`, etc. Those get used in the main body of the function to perform the various tasks.
+
+#### Extensions
+
+Any file exports in this directory will get automatically added to the `context` object as additional properties. Let's say we wanted a `context.foo()` function. Just add `foo.js` into `./src/extensions` and export a function from there. You'll then be able to access it anywhere as `context.foo()`.
```
skellock: Maybe add a note about how we often use extensions for namespacing or organizing?

docs/advanced-guides/tour.md
jamonholmgren: Neither did I. I literally looked up `console.dir` to find out what the frig it was, since I'd seen you use it a lot.

https://nodejs.org/api/console.html#console_console_dir_obj_options

Lots of cool stuff.

skellock: What's funny was, I've often wanted a depth setting!  I just never had the presence of mind to look up the docs.  Nice tip.  Thx.

PR #990, opened 2017-04-24
Minor fixes to tour documentation per @skellock

jamonholmgren:
Updates #989.


jamonholmgren:
My docs are so dry compared to yours, @skellock. 🤦‍♂️

PR #994, opened 2017-04-25
Adds ability to customize paths for generators

jamonholmgren:
Previously, generator paths were set by the host plugin and not customizable. This allows developers to edit their `ignite/ignite.json` files to specify other paths.

```json
{
  "paths": {
    "app": "App",
    "tests": "Tests",
    "components": "Components",
    "config": "Config",
    "containers": "Containers",
    "fixtures": "Fixtures",
    "images": "Images",
    "navigation": "Navigation",
    "services": "Services"
  }
}
```

I'll annotate various parts of this fairly wide-ranging pull request for easier review. Please review carefully -- I touched many parts of Ignite in this update.

### DIFF
--- a/docs/quick-start/project-structure.md
+++ b/docs/quick-start/project-structure.md
@@ -193,5 +193,35 @@ For example, you may:
 
 ## Ignite directory
 
-TODO:
 The Ignite directory stores all things Ignite. Here you will find generators, plugins and examples to help you get started with React Native.
+
+The `ignite.json` file is where you can customize how Ignite behaves in your project to some degree. For example, you can edit the paths that your plugin generators use when creating new components, screens, and the like.
+
+```json
+{
+  "generators": {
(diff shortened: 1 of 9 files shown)


jamonholmgren:
Thanks @skellock! I agree that the boilerplate is the right place to decide where to put files (hence this PR: https://github.com/infinitered/ignite-ir-next/pull/27). However, if we want the end-user to be able to customize the paths, ignite-cli has to support that, and that's what most of this PR is about.

The fallback opinions in this PR (which is the part I think you object to?) could be removed, for sure. I added them there for backwards compatibility, but now that I think about it, they should be in the boilerplate too.

Let's discuss in our meeting.

jamonholmgren:
After further discussion, we're going to move to a different option -- allowing a command-line flag to specify a custom path. I'll track it in #996.

PR #997, opened 2017-04-26
Integration tests

GantMan:
adding integration tests to CI

### DIFF
--- a/bin/testRelease
+++ b/bin/testRelease
@@ -89,6 +89,9 @@ check_builds()
   echo '~ Test adding maps plugin from npm'
   test_command ignite add maps
 
+  echo '~ LINT ALL THE THINGS!'
+  test_command npm run lint
+
   echo '~ Build ios'
   test_command react-native bundle --entry-file index.ios.js --bundle-output test.ios.js --dev false --platform ios --assets-dest iosAssets
   if [ ! -d "iosAssets" ]; then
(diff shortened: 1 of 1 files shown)


skellock:
I'm more in favour of using `ava` to test this stuff, but bash is fine. The important thing is that it's works.  

Since we're testing the ignite boilerplate, thoughts of moving this there?

Also, may as well delete `packages/ignite-integration-tests` now that we've switched to shell-based testing.

GantMan:
@skellock - let's talk about this.  I might have taken us backwards in time.  Since we did not have any e2e integration tests, I added this.  Are you saying `packages/ignite-integration-tests` was doing this already?

skellock:
The tests in our boilerplate are end-to-end tests.

We moved it over to `ignite-ir-boilerplate-2016` when we moved the boilerplate out of here.  What's left in `packages/ignite-integration-tests` is low value and can become unit tests.

Most likely, @jamonholmgren can delete that when he de-lernas this repo.

The reason I like our `ava` tests is because we can use things like `jetpack` and things to do fancy stuff instead of bash tricks.  It's not without it's annoyances though (e.g. `{preferLocal: false}` when calling `execa`.

That's just my preference.  I care more about working tests than I do my preference though.

GantMan:
I'm a fan of moving to JS over bash as well.  I don't think anything was testing like this was.  Perhaps we re-write this test in JS at some later date?

jamonholmgren:
I didn't see this until now. Whoops.

Issue #1013, opened 2017-05-05
Move React Native version dependency over to boilerplates


jamonholmgren:
Related to #957, we should move the version dependency to our boilerplates.

• Issue a warning if `opts.version` is not set [here](https://github.com/infinitered/ignite/blob/master/src/extensions/reactNative.js#L34)
• Update [ir-next](https://github.com/infinitered/ignite-ir-next/blob/091c4ad2616722eaa7e57ae577d014de7185477d/boilerplate.js#L86) and [default](https://github.com/infinitered/ignite-ir-boilerplate-2016/blob/master/boilerplate.js#L96) to properly call out which RN version they want

juicycleff:
This is the best route as it has been one issue I've been struggling with.

Issue #957, opened 2017-04-03
Support React Native 0.44.0


skellock:
React Native 0.43.0 is now available.

Also, let's [remove ignite depending](https://github.com/infinitered/ignite/blob/master/src/extensions/reactNative.js#L6) on a specific version and move that off to the boilerplates if they care.

MyTotoro:
so I just ncu updated and it broke everything... does ignite generally break when updating react/react-native?

skellock:
> so I just ncu updated and it broke everything

What is ncu? What errors are you seeing?

> does ignite generally break when updating react/react-native?

I'd say not often.  It depends on React Native's changes though.  RN is pretty good with letting us know what the breaking changes are.

MyTotoro:
https://www.npmjs.com/package/npm-check-updates

I just did a fresh install of an ignite template and updated React and RN (it broke only when I updated RN, when I updated React to 15.5.2, it was fine)

![image](https://cloud.githubusercontent.com/assets/3621921/24825109/c04996c0-1be5-11e7-824d-d1a5ccf8a18a.png)

Ignore the change in RootContainer, this is the error I'm seeing

![screenshot_20170407-230103](https://cloud.githubusercontent.com/assets/3621921/24825146/b5e712e2-1be6-11e7-8c93-01baca9d56d3.png)

skellock:
I googled and this came up: https://github.com/facebook/react-native/issues/13291 .  Related to your issue?

MyTotoro:
no luck, I've tried playing with different permutations - I think I'll just stick with the original 15.4.1 and .42

kevinvangelder:
React Native 0.43 is using React 16.0.0-alpha which isn't really ready for use in React Native.

skellock:
It's ready. There's just warnings out the wazoo.  :(  Wasn't their best release.  I say we wait for 44 =) =)

skellock:
44 is here.  It still uses 16.0.0-alpha.6.  Hilarity continues to ensue.  Recommend waiting for 0.45.0.

jamonholmgren:
Okay, thanks @skellock. Opened a related issue in #1013.

PR #1014, opened 2017-05-05
Add condition for rn version deprecation & fallback to default

fvonhoven:
## Describe your PR
We are going to require boilerplates to specify a RN version during their `install()`.
Add warning of deprecation and a fallback version.


jamonholmgren:
Fixes #1013 and #957. Also related to https://github.com/infinitered/ignite-ir-next/pull/34. We do need a similar PR for `ignite-ir-boilerplate-2016` before we can deploy this change, though.

PR #1021, opened 2017-05-09
Warn on plugin addModule if no version specified

### LINKED ISSUE #1006: Encourage Users To Specify Versions When Using `ignite.addModule`
Deprecation warning:
`ignite.addModule` should require version number.  Unless `yolo: true`

### Update Docs
https://github.com/infinitered/ignite/blob/master/docs/advanced-guides/creating-plugins.md#edit-the-pluginjs

jamonholmgren:
Fixes #1006.

image

### DIFF
--- a/src/extensions/ignite/addModule.js
+++ b/src/extensions/ignite/addModule.js
@@ -1,6 +1,16 @@
 const { trim } = require('ramda')
 
 module.exports = (plugin, command, context) => {
+  const getModuleName = (moduleName, options) => {
+    if (options.version) {
+      return `${moduleName}@${options.version}`
+    } else if (!options.yolo) {
+      context.print.warning(`Plugin should specify specific version for NPM module ${moduleName} in addModule call.`)
+      context.print.warning(`Please add version: "<version here>" or to suppress this warning, pass in yolo: true`)
+    }
+    return moduleName
+  }
+
   /**
    * Adds a npm-based module to the project.
    *
@@ -13,9 +23,7 @@ module.exports = (plugin, command, context) => {
   async function addModule (moduleName, options = {}) {
     const { print, system, ignite } = context
     const { useYarn } = ignite
-    const moduleFullName = options.version
-      ? `${moduleName}@${options.version}`
-      : moduleName
+    const moduleFullName = getModuleName(moduleName, options)
 
     const depType = options.dev ? 'as dev dependency' : ''
     const spinner = print.spin(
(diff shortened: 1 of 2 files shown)


skellock:
I'm:

* 👍  on the plugin changes
* 👎  on the warning
* 👎  on the`--yolo` option

### Strawman-mode Engage!

So, run this: `react-native init FunTown`

image

See that yellow warning in the middle?

As a person building a new React Native project, that's meaningless.

Here's the same thing with `create-react-app`:

image

So, if I was given the power to `--yolo` my way thru it, I'd never do it. It would never fix anything.

What I need is for upstream to fix their stuff.  I don't even know how far upstream either.

### Jump To Conclusions Mat

And, so this is what this PR is bringing to the table, forcing info on end-users where no action can be taken.

### Recommendations

* Keep the `plugin.js.ejs` changes. As plugin & boilerplate authors, it's on us to responsibly install our dependencies.

* Drop the `--yolo` enforcement.

I could be convinced (or at least be upgraded to 😐 ) to keep the warning (if the wording was more along the lines of

> Dearest innocent bystander,
>
> Go to {repo} and file an issue with {ignite-plugin} to tell them to add {version} to their addModule({dependency}) like we recommend in {link to docs}.
>
> -- ignite team

jamonholmgren:
I'm all with you on this, @skellock, makes sense.

I thought you were for the warning, though...?

image

skellock:
I thought i was too.

GantMan:
I think we're saying "warning" so we can do a deprecation phase.  I'm fine if it says what @skellock says, but we need something to identify it's not ok... other than just forcing it.

Or maybe we just force it?

src/extensions/ignite/addModule.js
```diff
@@ -1,6 +1,16 @@
...
 
 module.exports = (plugin, command, context) => {
+  const getModuleName = (moduleName, options) => {
+    if (options.version) {
+      return `${moduleName}@${options.version}`
+    } else if (!options.yolo) {
+      context.print.warning(`Plugin should specify specific version for NPM module ${moduleName} in addModule call.`)
+      context.print.warning(`Please add version: "<version here>" or to suppress this warning, pass in yolo: true`)
+    }
+    return moduleName
```
jamonholmgren: Okay, okay, I'll change it. 🙄 

Sidenote: I like multiple returns -- short circuit the rest of the function if it's not necessary, like a guard statement.

But I'm not dogmatic about it. I don't want your skin crawling. 😁

jamonholmgren:
I'm solving the problem by only outputting the warning if we're adding the plugin from a path, like a plugin author might.

image

skellock:
That's a good compromise.  Thank you.

PR #1038, opened 2017-05-18
Adds deprecation warning when no .ejs extension for template

### LINKED ISSUE #690: Prevent addComponentExample from magically appending .ejs
This is the wrong level of magic:

~https://github.com/infinitered/ignite/blob/next/packages/ignite/src/extensions/ignite.js#L201~
https://github.com/infinitered/ignite/blob/master/packages/ignite-cli/src/extensions/ignite/addPluginComponentExample.js#L23

Not all component examples are templates, and there's no harm in running all files through the generator, but let's not have the user type `addComponentExample('foo.js')` when it's `foo.js.ejs` on the file system.

Let's change our current usage to pass in `foo.js.ejs`.  Let's also print a `warning` that the template was not found, but not bomb the install of the plugin.

Found by @robinheinze

jamonholmgren:
Fixes #690 and adds deprecation warning so we can fix any plugins soon.

### DIFF
--- a/src/extensions/ignite/addPluginComponentExample.js
+++ b/src/extensions/ignite/addPluginComponentExample.js
@@ -9,6 +9,14 @@ module.exports = (plugin, command, context) => {
     const { filesystem, ignite, print, template } = context
     const { ignitePluginPath } = ignite
     const config = ignite.loadIgniteConfig()
+    
+    let templateFile
+    if (fileName.endsWith('.ejs')) {
+      templateFile = fileName
+    } else {
+      print.warning(`DEPRECATION WARNING: addPluginComponentExample called with '${filename}' and no .ejs extension. Add .ejs to your filename when calling this function.`)
+      templateFile = `${fileName}.ejs`
+    }
 
     // do we want to use examples in the classic format?
     if (config.examples === 'classic') {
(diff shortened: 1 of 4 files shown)


src/extensions/ignite/addPluginComponentExample.js
```diff
@@ -9,6 +9,14 @@ module.exports = (plugin, command, context) => {
     const { filesystem, ignite, print, template } = context
     const { ignitePluginPath } = ignite
     const config = ignite.loadIgniteConfig()
+    
+    let templateFile
+    if (fileName.endsWith('.ejs')) {
+      templateFile = fileName
+    } else {
+      print.warning(`DEPRECATION WARNING: addPluginComponentExample called with '${filename}' and no .ejs extension. Add .ejs to your filename when calling this function.`)
```
skellock: fileName

src/extensions/ignite/addPluginScreenExamples.js
```diff
@@ -38,6 +38,14 @@ module.exports = (plugin, command, context) => {
         files
       )
 
+      let templateFile
+      if (fileName.endsWith('.ejs')) {
+        templateFile = fileName
+      } else {
+        print.warning(`DEPRECATION WARNING: addPluginScreenExample called with '${filename}' and no .ejs extension. Add .ejs to your filename when calling this function.`)
```
skellock: fileName

Issue #1040, opened 2017-05-18
Upgrading gluegun causes error: "extension.setup" not a function


jamonholmgren:
Upgrading from gluegun 0.17.1 to 0.18.0 results in this error when running `yarn run integration`:

![image](https://cloud.githubusercontent.com/assets/1479215/26189229/5611419e-3b58-11e7-8608-ed22e130bd0e.png)

@skellock ... halp... ?

skellock:
We changed the interface in https://github.com/infinitered/gluegun/pull/184 and my breaking change radar didn't go off.  Will fix in the morning.  Sorry.

jamonholmgren:
Okay, thank you @skellock !

skellock:
Slipped my radar.

Good to go now @jamonholmgren once you [pull the merge trigger](https://github.com/infinitered/ignite/pull/1046).

PR #1041, opened 2017-05-18
Upgrades dependencies and adds npm run shipit

jamonholmgren:
This also pins all dependencies. We should run `npm outdated` before every release. Couldn't update gluegun because of #1040.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,8 +41,7 @@
     "ramdasauce": "1.2.0",
     "semver": "5.3.0",
     "shelljs": "0.7.6",
-    "which": "1.2.14",
-    "np": "2.12.0"
+    "which": "1.2.14"
   },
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -30,25 +31,26 @@
...
+    "gluegun": "0.17.1",
+    "gluegun-patching": "0.3.0",
+    "minimist": "1.2.0",
+    "pretty-error": "2.0.2",
+    "ramda": "0.23.0",
+    "ramdasauce": "1.2.0",
+    "semver": "5.3.0",
+    "shelljs": "0.7.6",
+    "which": "1.2.14",
+    "np": "2.12.0"
```
skellock: Could you move `np` to a dev dependency please?

PR #1049, opened 2017-05-23
Adds overwrite project on ignite new

jamonholmgren:
Allows for `--overwrite` flag on `ignite new MyApp` to overwrite an existing folder. Useful for testing over and over. Will also ask (rather than just abort) if it encounters an existing folder without the flag.


jamonholmgren:
Tests pass locally but keep timing out on Semaphore. I think I know why.

PR #1053, opened 2017-05-24
Allows setting @ version on ignite add

jamonholmgren:
Prior to this change, doing `ignite add i18n@0.1.2` would result in an error.

Related to https://github.com/infinitered/ignite-ir-next/issues/57.

### DIFF
--- a/src/commands/add.js
+++ b/src/commands/add.js
@@ -59,6 +59,12 @@ Examples:
     process.exit(exitCodes.OK)
   }
 
+  // we don't (yet) support scoped npm packages, so bail
+  if (parameters.second.startsWith('@')) {
+    print.error(`Error: Ignite CLI doesn't support prefixed npm packages yet, sorry about that. Try it without the prefix.`)
+    process.exit(1)
+  }
+
   // find out the type of install
   const specs = detectInstall(context)
   const { moduleName } = specs
--- a/src/lib/detectInstall.js
+++ b/src/lib/detectInstall.js
@@ -32,25 +32,27 @@ function detectInstall (context) {
 
   // the plugin we're trying to install
   let plugin = parameters.second
+  let packageName = plugin.split('@')[0]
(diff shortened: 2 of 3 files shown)


jamonholmgren:
Hold off, have an issue with this. Will resubmit.

jamonholmgren:
Ready to review now.

src/lib/detectInstall.js
```diff
@@ -32,25 +32,27 @@ function detectInstall (context) {
 
   // the plugin we're trying to install
   let plugin = parameters.second
+  let packageName = plugin.split('@')[0]
```
skellock: Just a heads up.  I know I lost the argument to allow non-ignite prefixes, but this is a common thing nowadays to use scoped packages.  They look like `@infinitered/ignite`.  That `@` will mess with this code you've written here.  Maybe we could just do a quick bailout if it `startsWith` `@`?

That's bonus points, though.  I'm ok with this PR as is.

src/commands/add.js
```diff
@@ -59,6 +59,12 @@ Examples:
     process.exit(exitCodes.OK)
   }
 
+  // we don't (yet) support scoped npm packages, so bail
+  if (parameters.second.startsWith('@')) {
+    print.error(`Error: Ignite CLI doesn't support prefixed npm packages yet, sorry about that. Try it without the prefix.`)
```
skellock: Can you change the message to call them scoped packages? https://docs.npmjs.com/getting-started/scoped-packages

src/commands/add.js
```diff
@@ -59,6 +59,12 @@ Examples:
...
+    process.exit(1)
```
skellock: Can you use a code from [exitCodes.js](https://github.com/infinitered/ignite/blob/master/src/lib/exitCodes.js)?

skellock:
Lol, merged.  Oh well.  RIP.

PR #1054, opened 2017-05-25
Use proper terminology and exit code per @skellock

jamonholmgren:
Follow up for #1053.


jamonholmgren:
Self-merging for now, but appreciate eyes on this.

PR #1055, opened 2017-05-25
Fixes bug where .ejs was left on plugin examples

jamonholmgren:
Plugin examples retained their `.js.ejs` extension, causing a crash. This fixes that.


jamonholmgren:
Self-merging for now, but still appreciate eyes on this.

Issue #1043, opened 2017-05-20
Sporked templates not being used


SRandazzo:
What's going on?

After running `ignite spork` and selecting the templates I want to spork, they are copied into `ignite/Spork/ignite-ir-boilerplate-2016`

I've modified the template (redux.ejs in this case), saved and then run `ignite generate redux MyState`

The new MyState.js that is created is not based on my modified .ejs template, but instead reflects the default template

---
Steps to reproduce

run `ignite spork`
select the templates you want to spork (in this case redux)
modify the template
run `ignite generate redux MyState`
Observe the newly created MyState.js is using the original format, not your custom sporked template

thanks!

---
`ignite doctor` results:

```
System
  platform           darwin

  arch               x64

  cpu                8 cores                                                 Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz
  directory          /Users/sal/Sites/paperlesspost/pages/packages/PagesRN


JavaScript
  node               7.10.0       /Users/sal/.nvm/versions/node/v7.10.0/bin/node
  npm                4.2.0        /Users/sal/.nvm/versions/node/v7.10.0/bin/npm
  yarn               0.24.4       /usr/local/bin/yarn

React Native
  react-native-cli   2.0.1
  app rn version     0.44.0

Ignite
  ignite             2.0.0-beta.9   /Users/sal/.nvm/versions/node/v7.10.0/bin/ignite

Android
  java               1.8.0_131    /usr/bin/java
  android home       -            undefined

iOS
  xcode              8.3.2
```

jamonholmgren:
Thanks @SRandazzo , I'll look into this before we release 2.0.0.

jamonholmgren:
Verified this is happening.

jamonholmgren:
The reason this isn't working is because in most plugins we switched our generators over to `copyBatch`, which doesn't include spork capability.

I'll work on implementing that now.

SRandazzo:
thanks!!! 👏 🎉

PR #1056, opened 2017-05-25
Fixes sporking issue, adds better tests

### LINKED ISSUE #1050: Create e2e Tests for spork functionality
it seems Spork might have broken and is reported in #1043 

We need tests to make sure we don't have this regression again.

jamonholmgren:
Spork was broken in `copyBatch` generators. This fixes issue #1043.

I also added an end-to-end test. The assertions can be filled out much more, but I'm happy to have what we have here. Should be very useful to catch `ignite new` issues, and it also tests a generator and of course sporking. This fixes #1050.

I also had to add the ability to specify a spork file from the initial command. Example: `ignite spork component.ejs` should spork just that file and not wait for input. Truthfully, that's how I thought it worked already!

### DIFF
--- a/src/extensions/ignite/copyBatch.js
+++ b/src/extensions/ignite/copyBatch.js
@@ -1,4 +1,13 @@
 module.exports = (plugin, command, context) => {
+  
+  function findSpork(context, template) {
+    // console.dir(context, {colors: true, depth: 1})
+    const { filesystem } = context
+    const sporkDirectory = `${filesystem.cwd()}/ignite/Spork/${context.plugin.name}`
+
+    return filesystem.exists(`${sporkDirectory}/${template}`) ? sporkDirectory : false
+  }
+
   /**
    * Runs a series of jobs through the templating system.
    *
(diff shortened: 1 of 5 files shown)


src/extensions/ignite/copyBatch.js
```diff
@@ -38,8 +47,11 @@ module.exports = (plugin, command, context) => {
       // generate the React component
       if (await shouldGenerate(job.target)) {
         const currentPluginPath = ignitePluginPath()
+
+        const sporkDirectory = findSpork(context, job.template)
+
         await template.generate({
-          directory: directory || (currentPluginPath && `${currentPluginPath}/templates`),
+          directory: directory || sporkDirectory || (currentPluginPath && `${currentPluginPath}/templates`),
```
jamonholmgren: I assume someone passing in `directory` should take precedence over a spork. This is an unusual circumstance, I think.

tests/integration/ignite-new/new.test.js
```diff
@@ -0,0 +1,41 @@
...
+const execa = require('execa')
+const tempy = require('tempy')
+const jetpack = require('fs-jetpack')
+
+const IGNITE = `${process.cwd()}/bin/ignite`
+const APP_DIR = 'Foo'
+
+test.before(t => {
+  const tempDir = tempy.directory()
+  process.chdir(tempDir)
```
jamonholmgren: Now creates tests in a temporary directory, like we should have been doing all along. :)

tests/integration/ignite-new/new.test.js
```diff
@@ -0,0 +1,41 @@
...
+})
+
+test('spins up a min app and performs various checks', async t => {
+  await execa(IGNITE, ['new', APP_DIR, '--min'])
+  process.chdir(APP_DIR)
+
+  // check the contents of ignite/ignite.json
+  const igniteJSON = jetpack.read('ignite/ignite.json')
+  t.is(typeof igniteJSON, 'string')
+  t.regex(igniteJSON, /"generators": {/)
```
skellock: TIL

Issue #1047, opened 2017-05-23
Release plan for 2.0.0


jamonholmgren:
## Checklist for release 2.0.0-beta.10 and subsequently 2.0.0

* [x] ignite-i18n update merged https://github.com/infinitered/ignite-i18n/pull/7
* [x] ignite-vector-icons update merged https://github.com/infinitered/ignite-vector-icons/pull/6
* [x] ignite-animatable update merged https://github.com/infinitered/ignite-animatable/pull/6
* [x] ignite-ir-next ejs update merged https://github.com/infinitered/ignite-ir-next/pull/54
* [x] ignite-ir-next skip git init merged https://github.com/infinitered/ignite-ir-next/pull/53
* [x] ignite-ir-next refine the walkthrough process https://github.com/infinitered/ignite-ir-next/issues/31 
* [x] ignite-ir-next better success message https://github.com/infinitered/ignite-ir-next/pull/55
* [x] ignite-ir-next pin plugin versions https://github.com/infinitered/ignite-ir-next/issues/57
* [x] ignite overwrite flag https://github.com/infinitered/ignite/pull/1049
* [x] release ignite-dev-screens 2.0.0
* [x] release ignite-i18n 1.0.0
* [x] release ignite-vector-icons 1.0.0
* [x] release ignite-animatable 1.0.0
* [x] release ignite-standard 1.0.0
* [x] *discuss and settle on a good strategy for boilerplate-2 release*
* [x] release ignite-ir-boilerplate 2.0.0-rc.1
* [x] fix sporking issue https://github.com/infinitered/ignite/issues/1043 (pending PR #1056)
* [x] point Ignite CLI master to boilerplate-2 by default (pending PR #1057)
* [x] release ignite CLI 2.0.0-rc1
* [x] release ignite CLI 2.0.0 and ir-boilerplate 2.0.0
* [x] release Jamon's blog post (may need freshening)

skellock:
I did the release notes first, but github won't let me save a new release it until I give it a tag.

Soo.... gonna park this here until I do that!

# 2.0.0-beta.10

## :tada: Features
* `ignite add` now enforces version numbers - so we don't break for semver incompatibilites in the future. #1053, #1054 by @jamonholmgren
* Boilerplates must now include a React Native target version so things accidently break on new releases. #1014 by @fvonhoven
* `ignite new` has a new flag `--overwrite` which will forceable clobber the destination directory. #1049 by @jamonholmgren
[...]

PR #1057, opened 2017-05-25
Switches default boilerplate from ir-boilerplate-2016 to ir-boilerplate

jamonholmgren:
Moving from the venerable `ignite-ir-boilerplate-2016` to the fresh, hot `ignite-ir-boilerplate` version 2.0.0 (Andross), which is currently released as `2.0.0-rc.1`.

### DIFF
--- a/BOILERPLATES.md
+++ b/BOILERPLATES.md
@@ -4,7 +4,7 @@
 
 | Name | Description |
 |------|-------------|
-| [ignite-ir-next](https://github.com/infinitered/ignite-ir-next) | The latest and greatest of Infinite Red opinions |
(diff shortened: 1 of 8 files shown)


jamonholmgren:
Once #1056 is merged and subsequently this one, let's release `ignite-cli@2.0.0-rc.1` here (@skellock, can you handle that tomorrow?). Then we're ready to test the two release candidates for a week or so and then release 2.0.0 next week. Ref #1047.

skellock:
I can release this yes. I would prefer to show someone how though.

jamonholmgren:
Updated to `ignite-ir-boilerplate`.

PR #1067, opened 2017-05-26
Fix for wrong example files being generated.

### LINKED ISSUE #1066: Example files end with a `.`
On a fresh `ignite new Test --max`, the example files are all appended with a period.

image

skellock:
Fixes #1066


skellock:
The failure is caused by a missing dependency `tempy`.  I will fix that next, but it's not part of this PR.

Also self-merging.

PR #1068, opened 2017-05-26
Adds missing `tempy` dependency from package.json.

skellock:
This dependency was missing causing the integration tests to fail.


skellock:
Important bug fix, so I'm self merging.  Got the greenlight from CI.  That's a real person right?

jamonholmgren:
Gah my bad.

PR #1072, opened 2017-05-29
Detect failed imports

GantMan:
As of right now, if an import has a typo, no tests will catch it.

This will help us with end to end testing of failed import statements, BUT, it will also possibly be a good solution for us to add to all projects.


skellock:
This seems to have gone through a lot of revisions.  Is this ready to review? I see that runs the generators. Is that enough to detect failed imports?

skellock:
Can you rebase this please?

skellock:
Ok, i had to select the drop down box next to the greyed out button. lol.

Issue #1058, opened 2017-05-25
bin/ci cleanup -- remove hooks


jamonholmgren:
When you run `bin/ci` it leaves behind some git hooks. Let's clean those up (I usually do `rm -rf .git/hooks` after a ci run).

PR #1091, opened 2017-06-24
Updates ci script to run in a temp directory

jamonholmgren:
Rather than running integration tests in `./testgrounds`, this change will run those tests in a temporary directory. This solves #1058 and other potential issues.


jamonholmgren:
I committed `package-lock.json` ... should I have? @skellock

PR #1095, opened 2017-06-25
lots of windows love

### LINKED ISSUE #1093: [Windows] Can't find boilerplate template from flag
image

GantMan:
This fixes #1093 100% and probably fixes a good bit of other Windows issues.

❤️


GantMan:
@skellock I started off being selective on slashes fixes and tests, but then errors still existed and tests went red, so I kind of flip-flopped over to "Let me kill it with fire" mode.

Issue #1107, opened 2017-07-13
Jest tests run against local code, not against the bundle


aschrijver:
### Local project files are used for Jest tests, not transpiled bundle code

---
I have created a project using the standard boilerplate, then added `babel-plugin-module-resolver`: "2.7.1"` to the package and in `.babelrc`:

```json
  "plugins": [
    ["module-resolver", {
      "alias": {
        "mock": "./config/mock",
        "fs": "./config/mock",
        "sodium-universal": "libsodium",
        "crypto": "react-native-crypto",
        "path": "path-browserify"
      }
    }]
  ]
```

This because I have some node-based dependencies (without UI) that I want to reuse in my RN app.
With this plugin Babel neatly replaces imports and requires throughout `node_modules`.

When running the emulator I get an error complaining that `fs.readFileSync` is undefined, which is completely expected, as `fs` is just a mock, but is invoked from one of those node-related dependencies.

Now when running Jest unit tests that invoke the same methods, using `yarn test` (defined as `"NODE_ENV=test jest --no-cache"` in package.json) or in any other way I could think of, the tests succeed, where they should also fail!

So seems like Jest is completely ignoring the bundle and happily uses the original sources from disk.

---
Steps to reproduce:
- Create new Ignite project
- Add babel-plugin-module-resolver and .babelrc
- Point alias `fs` to an empty js file
- Invoke readFileSync in App
- Create and run a test
---
`ignite doctor` results:
```
System
  platform           linux                                                                                                
  arch               x64                                                                                                  
  cpu                4 cores                                          Intel(R) Core(TM) i7-4500U CPU @ 1.80GHz  
  directory          /home/arnold/Development/fullcircle/Fullcircle                                                       

JavaScript
  node               7.10.1       /usr/local/bin/node  
  npm                4.2.0        /usr/local/bin/npm   
  yarn               0.24.6       /usr/bin/yarn        

React Native
  react-native-cli   2.0.1       
  app rn version     0.45.1      

Ignite
  ignite             2.0.0        /usr/local/bin/ignite  

Android
  java               1.8.0_111    /usr/bin/java  
  android home       -            undefined 
```

PS. This may be more of a Jest or babel-jest issue, but I'm filing here to start with :)

skellock:
Do you experience the same issue without ignite as well?  `react-native init Strangeness --version 0.45.1`

aschrijver:
[...]

PR #1108, opened 2017-07-13
Update README.md Node V8 debugging warning #1107

aschrijver:
Added the warning on Node V8 debugging as discussed in #1107


GantMan:
thanks for this!!!

PR #1125, opened 2017-07-29
Add redux-thunk to plugins list

kusti8:
Update the docs to add redux-thunk to the plugins list. This is an ignite plugin that provides redux-thunk.


jamonholmgren:
Nice, thanks @kusti8 !

PR #1132, opened 2017-08-02
Prepares the option of using Ignite structure in web projects

sidferreira:
A first step to have Ignite boilerplates for Web projects

This should change nothing on RN projects except  the "React" option on Install context

### DIFF
--- a/src/extensions/react.js
+++ b/src/extensions/react.js
@@ -1,0 +1,120 @@
+const { test, trim } = require('ramda')
+const exitCodes = require('../lib/exitCodes')
+
+// DEPRECATED: Please specify React version when invoking install
+// Example: const rnInstall = await reactNative.install({ name, version: '0.42.0' })
+const REACT_VERSION = '15.6.1'
+
+/**
+ * Attach this extension to the context.
+ *
+ * @param   {any} plugin  - The current running plugin.
+ * @param   {any} command - The current running command.
+ * @param   {any} context - The active context.
+ * @returns {any}         - The public-facing features this supports.
+ */
+function attach (plugin, command, context) {
+  // fist-full o features
+  const { parameters, print, system, filesystem, strings, ignite } = context
+  const { log } = ignite
+
+  /**
+   * Installs React Native.
+   *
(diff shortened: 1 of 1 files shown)


jamonholmgren:
Would you mind providing a test or two for this, @sidferreira ?

sidferreira:
Sure! Will do tomorrow morning!

jamonholmgren:
@skellock @GantMan Any additional thoughts before we merge?

skellock:
I vote no for now. (and i don't mean that in a YOU SUCK way either)

### superficial reasons

* this uses `create-react-app`.  which is a boilerplate and brings it's own opinions.
* we don't check that `create-react-app` is installed
* i'm not sure if `create-react-app` supports `--skip-jest`
* if people use `ignite` for web apps now, they still have to have `react-native` installed
* once it is in, we're now supporting it and nobody uses it here

### slightly bigger picture

Here's what the landscape looks like now:

image

With this PR, it becomes this:

image

Where we're hopefully going is like this:

image

@sidferreira I think it's *awesome* you're working on the `create-react-app` version!  And it's not that this PR is wrong, I just think it's a bit too soon to put into the core.

If there's anything I can help you do in your boilerplate, don't hesitate to ask!

GantMan:
yeah, I'm going to be doing boilerplate soon, too.  Best to make the flavors of react stay in the boilerplate.  Let's just move this code there.

sidferreira:
I really agree with @skellock . 
At first, I was expecting to modify only the boilerplate. After that, I noticed I had to modify it in the cli.
At the same time, not proceeding with this PR will basically freeze the ignite-web.
So my opinion depends on how long the CLI change would take...

skellock:
@sidferreira in your `boilerplate.js`, instead of calling https://github.com/infinitered/ignite-ir-boilerplate/blob/master/boilerplate.js#L48, you'd call the function this file instead.

skellock:
Closing for now.  Thx @sidferreira.  Don't hesitate to summon me if you need help on your boilerplate.

Issue #1134, opened 2017-08-07
ignite-cli does not play well with symlinks


l1br3:
---
Linked npm packages are not available as boilerplate
---
You can try this by using a `npm link`ed package or even by replacing the ignite-{yourBoilerplateName} folder by a symlink.

Create a boilerplate plugin, link it then try to ignite a new app with that boilerplate
```bash
#Linking result
/home/libre/.nvm/versions/node/v8.2.1/lib/node_modules/ignite-AutoRnApp -> 
/home/libre/dev/AutoRnApp/ignite/plugins/ignite-AutoRnApp

# Then ignite the app
libre@libre:~/dev$ ignite new BoilerTest -b ignite-AutoRnApp #Also tried without "ignite-" prefix

🔥 igniting app BoilerTest
✖ ignite-AutoRnApp is not available on npm.

-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-
  We also searched in these directories:
```

---
```
libre@libre:~/dev$ ignite doctor
System
  platform           linux                                                                 
  arch               x64                                                                   
  cpu                4 cores           Intel(R) Core(TM) i5-4690K CPU @ 3.50GHz  
  directory          /home/libre/dev                                                       

JavaScript
  node               8.2.1        /home/libre/.nvm/versions/node/v8.2.1/bin/node  
  npm                5.3.0        /home/libre/.nvm/versions/node/v8.2.1/bin/npm   
  yarn               0.27.5       /home/libre/.nvm/versions/node/v8.2.1/bin/yarn  

React Native
  react-native-cli   2.0.1       

Ignite
  ignite             2.0.0        /home/libre/.nvm/versions/node/v8.2.1/bin/ignite  

Android
  java               1.8.0_131    /usr/bin/java            
  android home       -            /home/libre/Android/Sdk  

```

l1br3:
If I install the new app with relative path it works but then if I delete manualy or with npm uninstall the boilerplate plugin and then link my local plugin, it does not work. The generate command does nothing.

Could you elaborate on how boilerplates packages are managed by ignite cli ?

l1br3:
Well, I understand that ignite cli does not detect symlinks

skellock:
It’s not that it doesn’t work with symlinks, it’s that it doesn’t check with your global npm packages for boilerplates.

Which is perfectly reasonable to expect.

[...]

PR #1141, opened 2017-08-15
Upgrades a few dependencies.

skellock:
The main purpose of this upgrade was to get us new `gluegun` features.  There were a few others i noticed were aging so i brought them up too.

Finally the `fs-mock` now works.  (tested on node 8.2.1)

This gives us `symlink` support #1134 (for example using `npm link`).


skellock:
I didn't explain this well.

What this allows us to do is:  

1. develop plugins/boilerplates within a project.
2. open up that project's `node_modules/ignite-thing/` files, hit save, and it gets picked up immediately.

A quality-of-life PR.

PR #1256, opened 2018-04-16
📝 Adds mst documentation overview

skellock:
This some prelim notes about the `mobx-state-tree` document that @robinheinze and I are putting together.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -5,7 +5,8 @@
 * Installing
 * [Getting started](./quick-start/getting-started.md)
 * [Ignite CLI commands](./quick-start/ignite-commands.md)
-* [Project Structure](./quick-start/project-structure.md)
+* [Project Structure - Andross](./quick-start/project-structure-andross.md)
+* [Project Structure - Bowser](./quick-start/project-structure-bowser.md)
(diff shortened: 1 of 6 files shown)


robinheinze:
I think 2 documents should come out of this.

1) A new `project-structure.md`, replacing https://github.com/infinitered/ignite/blob/master/docs/quick-start/project-structure.md. This is where we'll detail our new directory structure, along with our (IR's) specific patterns around MST. 

- In addition to that, I think we should have a separate `intro-to-mobx.md`, also under quick-start that gives the general overview to MST, including gotchas, that is totally non-specific to IR and intended for the people who may never have been exposed to MobX or MobX State Tree. I think this is important because there's far fewer resources on the internet at large for MST issues and troubleshooting, so users will most likely turn back to us. 

I'm definitely willing to take on either or both of these.

robinheinze:
@GantMan @skellock I just added a commit which changes `project-structure.md` to talk about the new structure rather than the old one. It's very sparse now, but it should suffice for release. Let me know if you want me to add or change anything. 

I think tomorrow i'll go through Steve's `mst` document and just polish it up and make it more document-y and less brainstorm-y. 

Thoughts?

PR #1263, opened 2018-04-27
Update README for new Bowser world

robinheinze:
Updating the example in the README for Bowser

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,14 +34,12 @@ If you really liked Ignite 1 as it was, then not much has changed! You just run 
 ```
 $ npm install -g ignite-cli
 $ ignite new PizzaApp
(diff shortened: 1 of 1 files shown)


GantMan:
What are your thoughts on not pushing the user to choose bowser, but telling them commands they can run IF they choose bowser and command IF they choose Andross?

robinheinze:
So like:

"If you'd like the latest and greatest Infinite Red Bowser boilerplate with React Native 0.55.1, React Navigation, Mobx State Tree, Typescript, and more: 

```
Example goes here
```

If you'd prefer the tried and true Infinite Red Andross boilerplate with React Native 0.55.1, React Navigation, Redux, Redux Sagas, and more:

```
Example goes here
```

@gant ^^^^

PR #1277, opened 2018-05-21
A little linting cleanup

juddey:
Hey guys, 
I'm preparing to try to add support for `create-react-app` thru ignite, first though I thought I would lint things, and whaddya know.... found there were these errors:

```
  /home/justin/projects/node/ignite/src/cli/check.js:37:78: Trailing spaces not allowed.
  /home/justin/projects/node/ignite/src/cli/check.js:90:1: Too many blank lines at the end of file. Max of 0 allowed.
  /home/justin/projects/node/ignite/src/commands/new.js:74:47: Redundant use of `await` on a return value.
  /home/justin/projects/node/ignite/src/commands/new.js:187:1: Too many blank lines at the end of file. Max of 0 allowed.
  /home/justin/projects/node/ignite/src/commands/plugin.js:21:10: Redundant use of `await` on a return value.
  /home/justin/projects/node/ignite/src/extensions/ignite.js:36:23: 'system' is assigned a value but never used.
  /home/justin/projects/node/ignite/src/extensions/ignite/copyBatch.js:1:48: Block must not be padded by blank lines.
  /home/justin/projects/node/ignite/src/extensions/ignite/copyBatch.js:2:1: Trailing spaces not allowed.
  /home/justin/projects/node/ignite/src/extensions/ignite/copyBatch.js:3:21: Missing space before function parentheses.
  /home/justin/projects/node/ignite/src/extensions/ignite/copyBatch.js:37:14: Redundant use of `await` on a return value.
  /home/justin/projects/node/ignite/src/extensions/ignite/generate.js:21:12: Redundant use of `await` on a return value.
  /home/justin/projects/node/ignite/src/lib/importPlugin.js:58:1: Trailing spaces not allowed.
  /home/justin/projects/node/ignite/tests/fast/lib/detectInstall.test.js:28:6: Strings must use singlequote.
  /home/justin/projects/node/ignite/tests/integration/ignite-new/new.test.js:26:1: Trailing spaces not allowed.
  /home/justin/projects/node/ignite/tests/integration/ignite-new/new.test.js:41:1: Too many blank lines at the end of file. Max of 0 allowed.
```
so thought I would fix so everything is nice and clean. :rotating_light:

### DIFF
--- a/bin/ci
+++ b/bin/ci
@@ -60,7 +60,7 @@ check_builds()
   cd $(mktemp -d)
 
   echo '~~~🌟 Generating Project'
-  test_command ignite new TestProj --min --skip-git --overwrite
+  test_command ignite new TestProj --min --skip-git --overwrite -b ignite-ir-boilerplate-andross
 
   echo '~~~🌟 Checking Builds'
   cd ./TestProj
--- a/src/cli/check.js
+++ b/src/cli/check.js
@@ -34,7 +34,7 @@ try {
   process.exit(3)
 }
 
-// Only check full dependencies if we're running in `--debug` or `--wtf` mode 
+// Only check full dependencies if we're running in `--debug` or `--wtf` mode
 var debugMode = (process.argv.indexOf('--debug') !== -1 || process.argv.indexOf('--wtf') !== -1)
 
 if (debugMode) {
@@ -87,4 +87,3 @@ if (debugMode) {
     process.exit(exitCodes.INVALID_GLOBAL_DEPENDENCY)
   }
 }
-
--- a/src/commands/new.js
+++ b/src/commands/new.js
@@ -184,4 +184,3 @@ async function command (context) {
 }
 
 module.exports = command
-
--- a/src/commands/new.js
+++ b/src/commands/new.js
@@ -71,7 +71,7 @@ async function command (context) {
   // verify the directory doesn't exist already
   if (filesystem.exists(projectName) === 'dir') {
     print.error(`Directory ${projectName} already exists.`)
-    const askOverwrite = async () => { return await prompt.confirm('Do you want to overwrite this directory?') }
+    const askOverwrite = async () => { return prompt.confirm('Do you want to overwrite this directory?') }
 
     if (parameters.options.overwrite || await askOverwrite()) {
       print.info(`Overwriting ${projectName}...`)
--- a/src/commands/plugin.js
+++ b/src/commands/plugin.js
@@ -18,7 +18,7 @@ const walkthrough = async (context) => {
   if (context.parameters.options.max) { return maxOptions }
 
   // Okay, we'll ask one by one, fine
-  return await context.prompt.ask([
+  return context.prompt.ask([
     {
       name: 'boilerplate',
       message: 'Is this an app boilerplate plugin?',
--- a/src/extensions/ignite.js
+++ b/src/extensions/ignite.js
@@ -33,7 +33,7 @@ const pluginOverridesExt = require('./ignite/pluginOverrides')
  * @return {Function} A function to attach to the context.
  */
 function attach (plugin, command, context) {
-  const { parameters, system } = context
+  const { parameters } = context
 
   // determine which package manager to use
   const forceNpm = parameters.options.npm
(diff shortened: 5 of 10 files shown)


juddey:
I broke stuff! :smiley: 

image

Has the :robot: made a mistake with this or have I missed something?

juddey:
Hey so I have been working on this a bit further today. While I was writing an integration test for my changes for removing react-native as a dependency, I noticed that the integration tests didn't complete at all. They just sort of hung there with the lovely spinner showing. I reckon this is why the semaphore build has failed above.

Running locally, the integration test (tests/integration/ignite-new/new.test.js) never returns from this line:
```
await execa(IGNITE, ['new', APP_DIR, '--min'])
```
Seems like the `--min` switch doesn't work like it used to (?) I tried adding `-b ignite-ir-boilerplate-bowser` in various combinations in place of and as well as the `--min` but no good.

What should work here?

juddey:
Hey so I have been working on this a bit further today. While I was writing an integration test for my changes for removing react-native as a dependency (aren't I good writing the test first!), I noticed that the integration tests didn't complete at all. They just sort of hung there with the lovely spinner showing. Maybe this is why the semaphore build has failed above, because I don't think its because of my linting (?)

Running locally, the integration test (tests/integration/ignite-new/new.test.js) never returns from this line:

await execa(IGNITE, ['new', APP_DIR, '--min'])
Seems like the --min switch doesn't work like it used to (?) I tried adding -b ignite-ir-boilerplate-bowser in various combinations in place of and as well as the --min. and also with the `--no-boilerplate` option, but none of these were good.

What should work here?

kevinvangelder:
Summoning @GantMan

juddey:
OK, so kicked the tyres :car: a bit on this and found that the integration tests fail on master. Rolled back to commit 892fa5a, found integration tests passed successfully at this point, as long as this [line](https://github.com/infinitered/ignite/blob/892fa5a7ff175e472eeb9286536403b3c53c07db/tests/integration/ignite-new/new.test.js#L33) was changed to this:
 `const sporkedFile = 'ignite/Spork/ignite-ir-boilerplate-andross/component.ejs'`

Rolled back into master, applied this change, however the tests still hang. Maybe because of [this](https://github.com/infinitered/ignite/blob/9a01449b6f7a8440c8e62514f3e0ce4f06cade6b/src/commands/new.js#L107)? introduced in b124c8. But a dig into that looks like it was meant to be reverted...

.... now I'm really :confused:

adrienthiery:
Hi @juddey , 

I was looking at that as well and the only way to fix it for now I saw, was to modify the test (https://github.com/infinitered/ignite/blob/master/tests/integration/ignite-new/new.test.js#L15) to be `await execa(IGNITE, ['new', APP_DIR, '--min', '-b', 'ignite-ir-boilerplate-bowser'])`

Then it works! (takes a while, but it does)

Problem is, it block the same way on `await execa(IGNITE, ['g', 'component', 'Test'])`([here](https://github.com/infinitered/ignite/blob/master/tests/integration/ignite-new/new.test.js#L28))

So I guess this is a problem for all prompts.

The only solution I could see to that problem is if you could pass the answer of a prompt in command line, but thats another job. 

The idea would be that if you have a prompt that expects to get an answer for the `foo` parameter, if you pass `parameters.options.foo`, this should take it's place and disable the prompt.

Actually, @kevinvangelder , would you be open to that ?

I could work on a PR.

kevinvangelder:
Yeah, that's the pattern we've been using in most places, specifically for supporting automated testing. Do @GantMan  or @skellock have any different thoughts on this before Adrian gets started?

juddey:
Hey @adrienthiery! 

Thanks for the thoughts, and the idea to pass a specific boilerplate. I hadn't tried the combination of that _and_ the solution for the test I [talked about here](https://github.com/infinitered/ignite/pull/1277#issuecomment-391169262), using the `andross` boilerplate.

The tests pass locally, and I've just pushed a commit to see what will happen on the `ci` side of things.

juddey:
:point_down: Well, it looks like the ci script has the [same sort of problem](https://semaphoreci.com/ir/ignite/branches/pull-request-1277/builds/2). Looking into it.

juddey:
And there we go. :tada:

skellock:
@juddey insert Bette Midler song here

skellock:
I thought we removed that codacy plugin long ago.  Standby.

juddey:
@skellock _You_ are the wind beneath my wings.

juddey:
:tada:

PR #1284, opened 2018-05-30
Upgraded crashlytics plugin and changed name

adrienthiery:
- [ ] ~Everything works on iOS/Android~
- [ ] ~`npm test` **ava** tests pass with new tests, if relevant~

## Describe your PR

I had made `ignite-fabric` a while back, but since Fabric has been bought by Google (Firebase), the instructions to get started have changed, and now Crashlytics is replacing Crash detection on Firebase, so I revamped the plugin with the new setup instructions and renamed it to `ignite-crashlytics` (because that's actually the only thing it sets up).

Right now, the other plugin is still working, though, so I marked it as `(to be deprecated)`.

Is that good for you guys?


skellock:
Our CI is borked, so I'm going to cancel that build (41 minutes and counting).

PR #1323, opened 2018-09-30
README update: The conference is over

jakeasmith:
Looks like the conference is over, seems like a good time to drop the note :)


GantMan:
Good catch @jakeasmith !!! thanks :)

PR #1331, opened 2018-10-28
enable using specific version of a boilerplate

ruddell:

## Describe your PR
This allows developers to generate an app using a specific version of a boilerplate, such as a past release or a beta release.

I plan on releasing some big breaking changes to my boilerplate and it would be nice to allow users to generate using an older version if they wish.

### DIFF
--- a/docs/quick-start/ignite-commands.md
+++ b/docs/quick-start/ignite-commands.md
@@ -89,6 +89,8 @@ You can (with most boilerplates) pass through a `--min` or `--max` flag to autom
 
 If the new app's folder already exists, you can pass through `--overwrite` to overwrite the directory. If you don't, Ignite CLI will ask you if you want to overwrite it.
 
+If you want to use a specific version of a boilerplate, you can add the version to the boilerplate name.  For example, use `ignite new BetaApp -b ir-boilerplate-bowser@1.0.0-beta.1` to use the `1.0.0-beta.1` release of the Bowser boilerplate.
+
 ### Plugin
 
 ```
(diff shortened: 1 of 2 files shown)


mlaco:
@ruddell thanks for this PR, and good idea! We'll review it soon.

ruddell:
@kevinvangelder The `packageName` and `packageVersion` are split [in detectInstall.js](https://github.com/infinitered/ignite/blob/c41ad06ffd7eb293b73cfbeae3b0be335aec1cb6/src/lib/detectInstall.js#L37-L39)

kevinvangelder:
Ah, that's the piece I was missing, thanks. I haven't worked on this code base in far too long!

jamonholmgren:
I'll look at this next week. Thank you!

jamonholmgren:
This looks good to me. Will merge and deploy a new version soon.

ruddell:
It looks like this breaks `ignite add` when you use `->` at the start of the version (for example, in [the Andross boilerplate](https://github.com/infinitered/ignite-ir-boilerplate-andross/blob/master/boilerplate.js#L212): `ignite add i18n@"~>1.0.0"`). It also looks like that version isn't applied anwhere, it installs v1.1.1 even though v1.0.0 is specified.

I don't understand the purpose of that format, maybe someone meant `~` and added an extra `>`?

jamonholmgren:
@ruddell I'm working on setting up a CI/CD pipeline, which is what is holding this up. Should have it done today.

jamonholmgren:
Better late than never!

PR #1363, opened 2019-01-04
Install library globally, run from global dir

mlaco:
- [ ] `npm test` **ava** tests pass with new tests, if relevant

Untested presently

## Describe your PR
@jamonholmgren and I discussed the possibility of making `ignite-library` a plugin and installing it globally. This PR allows that. Ignite will recognize `ignite-library` specifically and do so. It will then locate the appropriate package manager's global installation location and run the plugin from there.

### DIFF
--- a/src/commands/add.js
+++ b/src/commands/add.js
@@ -30,7 +30,7 @@ const removeIgnitePlugin = async (moduleName, context) => {
 
 module.exports = async function (context) {
   // grab a fist-full of features...
-  const { print, filesystem, prompt, ignite, parameters, strings } = context
+  const { print, filesystem, prompt, ignite, parameters, strings, system } = context
   const { log } = ignite
 
   const perfStart = (new Date()).getTime()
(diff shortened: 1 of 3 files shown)


src/commands/add.js
```diff
@@ -67,8 +67,24 @@ Examples:
 
   // find out the type of install
   const specs = detectInstall(context)
-  const { moduleName } = specs
-  const modulePath = `${process.cwd()}/node_modules/${moduleName}`
+  const { moduleName, global } = specs
+  let modulePath
+  if (global) {
```
mlaco: `global` is determined in `detectInstall`. Only `ignite-library` will have this flag set to true.

src/lib/detectInstall.js
```diff
@@ -38,6 +38,12 @@ function detectInstall (context) {
   let packageName = plugin.split('@')[0]
   let packageVersion = plugin.split('@')[1] || null
 
+  let global = false
```
jamonholmgren: We won't want to use `global`, since it [conflicts with a predefined Node.js variable](https://nodejs.org/api/globals.html#globals_global). Consider using `isGlobal` instead.

src/lib/detectInstall.js
```diff
@@ -60,7 +66,8 @@ function detectInstall (context) {
         directory: path,
         override: true,
         moduleName: filesystem.read(`${path}/package.json`, 'json').name,
-        type: 'directory'
+        type: 'directory',
+        global: global
```
jamonholmgren: When using `isGlobal: isGlobal`, you can just do `isGlobal` and it'll infer the `isGlobal: ` part.

src/lib/detectInstall.js
```diff
@@ -38,6 +38,12 @@ function detectInstall (context) {
   let packageName = plugin.split('@')[0]
   let packageVersion = plugin.split('@')[1] || null
 
+  let global = false
+  const basePackageName = plugin.split(path.sep).slice(-1)[0]
+  if (basePackageName=='ignite-library') {
```
jamonholmgren: You'll want to use `===`  instead of `==`, because, well, [javascript](https://codeburst.io/javascript-double-equals-vs-triple-equals-61d4ce5a121a).

Also doesn't look like you have Prettier on in your code editor. You'll want to install that and ensure it's running properly.

Issue #1325, opened 2018-10-11
Allow custom prefixes for ignite plugins because of reasons


monotv:
Dear Ignite team,

recently I built a boilerplate based on your awesome Ignite CLI. I did not use the command line to initiate it but rather forked a different project.

The main focus was custom generators to support a certain file structure.

To my surprise, after adding the plugin in a project, the command line showed the correct generators but with `missing` descriptions. Also the generators failed silently. Apparently some data from the plugin was picked up correctly but it did not work. What was happening?

Turns out, when generating a plugin, the CLI prefixes its name with `ignite-`, which is then the `package.json` name, as outlined [here](https://github.com/infinitered/ignite/blob/master/docs/advanced-guides/creating-plugins.md#generate-a-basic-plugin-structure). But his prefix is also a _requirement_ for the plugin to work at all, because when I added it on a whim to my plugin's name - it worked!

This in itself is not much of a problem, merely a small documentation improvement.

But where a problem arises is in the case of scoped npm packages:
```
Creating a New Scoped Package
To create a new scoped package for your Organization scope, create a directory with the name of the package you would like to create. Navigate into that directory and use this command:

npm init --scope=<org_scope>
To verify that this worked, open the generated package.json with an editor of your choice. You should see that the name of the package will be @org_scope/<pkg_name>.
```
So if I want to publish an npm package for my organization, wether it be public or private, it can not be done for Ignite plugins, because of the must-have `ignite-` name prefix requirement.

I would very much like to manage distribute my Ignite plugins for internal use via npm, that's why I would like to ask for a change in the CLI to make this possible.

jamonholmgren:
Absolutely, this is a valid concern. Thanks for bringing it up @monotv !

Do you have a sense for how we might accomplish this? Simplest is probably just to detect the @ at the beginning of the string and don't make any changes to the name.

mlaco:
I'll take a shot at this as time permits

mlaco:
@monotv thank you for your patience. Our team was away on our yearly retreat last week. I'll try to get this done when I can.

PR #1370, opened 2019-01-14
Ignite scoped packages

mlaco:

## Describe your PR
Enables scoped packages to be added and removed

Tested by adding and removing a scoped package from npm.

Fixes https://github.com/infinitered/ignite/issues/1325

### DIFF
--- a/src/commands/remove.js
+++ b/src/commands/remove.js
@@ -53,15 +53,20 @@ module.exports = async function (context) {
   // take the last parameter (because of https://github.com/infinitered/gluegun/issues/123)
   const moduleParam = parameters.array.pop()
 
+  const isScoped = /^@/.test(moduleParam)
+
   // Check if they used a directory path instead of a plugin name
-  if (moduleParam.includes(path.sep)) {
+  if (moduleParam.includes(path.sep) && !isScoped) {
     error(`💩 When removing a package, you shouldn't use a path.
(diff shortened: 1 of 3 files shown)


src/commands/remove.js
```diff
@@ -53,15 +53,20 @@ module.exports = async function (context) {
   // take the last parameter (because of https://github.com/infinitered/gluegun/issues/123)
   const moduleParam = parameters.array.pop()
 
+  const isScoped = /^@/.test(moduleParam)
```
jamonholmgren:
```suggestion
  const isScoped = moduleParam.startsWith('@')
```

I'd prefer the more understandable `startsWith` function here.

src/lib/detectInstall.js
```diff
@@ -33,15 +33,19 @@ function detectInstall (context) {
...
   let plugin = parameters.second
-
+  
   // extract the package name and (optionally) version
-  let packageName = plugin.split('@')[0]
-  let packageVersion = plugin.split('@')[1] || null
+  let packageName, packageVersion
+  const versionSepRe = /(?<!^|\/)@/
+  const scopedRe = /^@/
+  const isScoped = scopedRe.test(plugin)
```
jamonholmgren:
```suggestion
  const isScoped = plugin.startsWith('@')
```

jamonholmgren: We won't need line 40 then.

Issue #1371, opened 2019-01-15
Error when using Node 8.9.4


alisarac:
### Hey there! Ignite CLI seems to be having a problem, eh?

---
What's going on?

I installed ignite-cli and then tried to create a new project. The error below occurred after selecting boilerplate. 

SyntaxError: Invalid regular expression: /(?<|^|\/)@/: Invalid group at detectInstall 
It looks this is a "RegExp Lookbehind Assertion" and it is not supported in node.js version 8.9.4 and previous versions, maybe requires a flag.

[es2018 feature support lists](https://node.green/#ES2018-features--RegExp-Lookbehind-Assertions)

After I updated my node.js version to v10.15.0 everything works as expected.

---
Steps to reproduce

`# ignite new MyApp`

then select any boilerplate

```vvv -----[ DEBUG ]----- vvv
SyntaxError: Invalid regular expression: /(?<!^|\/)@/: Invalid group
    at detectInstall (/Users/alisarac/.config/yarn/global/node_modules/ignite-cli/src/lib/detectInstall.js:39:24)
    at Command.installBoilerplate [as run] (/Users/alisarac/.config/yarn/global/node_modules/ignite-cli/src/commands/boilerplate-install.js:39:25)
    at Runtime.run (/Users/alisarac/.config/yarn/global/node_modules/gluegun/src/domain/runtime.js:159:46)
    at run (/Users/alisarac/.config/yarn/global/node_modules/ignite-cli/src/cli/cli.js:65:29)
    at Object.<anonymous> (/Users/alisarac/.config/yarn/global/node_modules/ignite-cli/bin/ignite:9:31)
    at Module._compile (module.js:643:30)
    at Object.Module._extensions..js (module.js:654:10)
    at Module.load (module.js:556:32)
    at tryModuleLoad (module.js:499:12)
    at Function.Module._load (module.js:491:3)
^^^ -----[ DEBUG ]----- ^^^
```
---
`ignite doctor` results:
System
  platform           darwin                                                                                    
  arch               x64                                                                                       
  cpu                4 cores                               Intel(R) Core(TM) i5-4570R CPU @ 2.70GHz  
  directory          /Users/alisarac/projects/mobile/CRM                                                       

JavaScript
  node               8.9.4        /Users/alisarac/.nvm/versions/node/v8.9.4/bin/node  
  npm                5.6.0        /Users/alisarac/.nvm/versions/node/v8.9.4/bin/npm   
[...]

PR #1387, opened 2019-03-05
Use regex without lookbehind in detectInstall

mlaco:
Lookbehind might not always be supported

- [ ] `npm test` **ava** tests pass with new tests, if relevant


jamonholmgren:
Fixes #1371.

PR #1416, opened 2019-05-05
Removes requirement to have Ignite attached to add a plugin

jamonholmgren:
There's no compelling reason to error when you're not in an Ignited project. This removes that requirement. If plugins want to write to a non-existent Ignite config, it'll be created for them (as long as they use `setIgniteConfig`).

This also fixes an issue I'm having with Ignite CLI 3.1.0.

### DIFF
--- a/src/commands/add.ts
+++ b/src/commands/add.ts
@@ -3,7 +3,6 @@ import * as R from 'ramda'
 import detectedChanges from '../lib/detected-changes'
 import detectInstall from '../lib/detect-install'
 import importPlugin from '../lib/import-plugin'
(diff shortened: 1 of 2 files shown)


jamonholmgren:
@ryanlntn I merged this, but I'd still like your thoughts.

ryanlntn:
I’m guessing the original intent behind that was to prevent accidentally adding plugins when in a subdirectory. What would happen if you try to add from a subdirectory of an ignite project now?

jamonholmgren:
I'll check that, good question.

Issue #1418, opened 2019-05-11
Not able to create a new plugin


GuillaumeHemmen:
Hello, I recently updated my ignite-cli and tried to create a new plugin with no success. See details below:

---
What's going on?

I tried to create a new plugin following the official documentation: https://github.com/infinitered/ignite/blob/master/docs/advanced-guides/creating-plugins.md

The command `ignite plugin new radio-buttons` throw the following error:

```
➜  ignite plugin new radio-buttons
ignite plugin new ignite-foo

Plugin name is required
```

---
Steps to reproduce

- Try to create a new plugin using the command `ignite plugin new <plugin name>`.

---
`ignite doctor` results:

```
➜  test ignite doctor
System
  platform           darwin

  arch               x64
  cpu                12 cores                        Intel(R) Xeon(R) CPU E5-1650 v2 @ 3.50GHz
  directory          /Users/ghemmen/Downloads/test

JavaScript
  node               10.15.3      /Users/ghemmen/.nvm/versions/node/v10.15.3/bin/node
  npm                6.9.0        /Users/ghemmen/.nvm/versions/node/v10.15.3/bin/npm
  yarn               1.16.0       /usr/local/bin/yarn

React Native
  react-native-cli   2.0.1

Ignite
  ignite-cli         3.1.1        /Users/ghemmen/.nvm/versions/node/v10.15.3/bin/ignite

Android
  java               1.8.0_172    /Library/Java/JavaVirtualMachines/jdk1.8.0_172.jdk/Contents/Home/bin/java
  android home       -            /Users/ghemmen/Library/Android/sdk

iOS
  xcode              10.2.1
```

GuillaumeHemmen:
I would also add that I get the same issue on a freshly install mac VM.

ruddell:
https://github.com/infinitered/ignite/blob/3da165cd48950f0e46d6d7e9a02c7218be945b8e/src/commands/plugin.ts#L51

It looks like `parameters.second` contains the plugin name instead of `parameters.third`
```
{ plugin: 'ignite',
  command: 'plugin',
  array: [ 'new', 'TestPlugin' ],
  options: {},
  raw: 
   [ '/Users/ruddell/.nvm/versions/node/v8.14.0/bin/node',
     '/Users/ruddell/.nvm/versions/node/v8.14.0/bin/ignite',
     'plugin',
     'new',
     'TestPlugin' ],
  argv: 
[...]

PR #1421, opened 2019-05-15
Updates plugin generator and adds integration test

jamonholmgren:
Fixes issue #1418, updates the generated Ignite plugin to latest specifications, adds some integration tests to ensure we don't break this in the future.

image


src/templates/plugin/package.json.ejs
```diff
@@ -10,23 +10,24 @@
...
-    "babel-eslint": "^7.1.1",
-    "np": "^2.12.0",
-    "nyc": "^10.1.2",
-    "sinon": "^1.17.7",
-    "standard": "^8.6.0"
+    "ava": "^1.4.1",
+    "babel-eslint": "^10.0.1",
+    "np": "^5.0.1",
+    "sinon": "^7.3.2",
+    "standard": "^12.0.1",
```
robinheinze: Do we even still use standard? I thought we had taken it out everywhere.

robinheinze:
Nice work! 💯

src/templates/plugin/package.json.ejs
jamonholmgren: We use tslint-config-standard on our other projects. This isn't TypeScript though.

PR #1456, opened 2019-08-19
Added "naive" check for cocoapods.

nirre7:

## Describe your PR
Added a "naive" check for CocoaPods if user is on mac, since it is now a RN core "prerequisite".

### DIFF
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -111,9 +111,8 @@ module.exports = {
       info(colors.cyan('iOS'))
       table([[column1('xcode'), column2(xcodeVersion)]])
 
-      const cocoaPodsPath = which('pod')
-      const cocoaPodsVersion = await run('pod --version', { trim: true })
-
+      const cocoaPodsPath = which('pod') || ''
+      const cocoaPodsVersion = cocoaPodsPath ? await run('pod --version', { trim: true }) : 'Not installed'
       table([[column1('cocoapods'), column2(cocoaPodsVersion), column3(cocoaPodsPath)]])
     }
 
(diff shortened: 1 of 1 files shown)


jamonholmgren:
@nirre7 Thanks for doing this! What will this do if it's run on Linux or Windows?

nirre7:
@jamonholmgren No worries. 
The check is within the 
`if (isMac)` code block so it should be fine.
However it does not fail gracefully when you don't have cocoapods installed on a mac.

src/commands/doctor.ts
```diff
@@ -110,6 +110,11 @@ module.exports = {
       info('')
       info(colors.cyan('iOS'))
       table([[column1('xcode'), column2(xcodeVersion)]])
+
+      const cocoaPodsPath = which('pod')
```
jamonholmgren: Can you just check `if (cocoaPodsPath) {` for the next block so you can print out that it isn't available instead of crashing?

jamonholmgren:
```typescript
const cocoaPodsPath = which('pod')
const cocoaPodsVersion = cocoaPodsPath ? await run('pod --version', { trim: true }) : "Not installed"
table([[column1('cocoapods'), column2(cocoaPodsVersion), column3(cocoaPodsPath)]])
```

jamonholmgren: ^ That would work, actually

nirre7: Should be a little less naive now

jamonholmgren:
@nirre7 Thanks so much! This looks great. Releasing now.

PR #1471, opened 2019-10-04
Find locally installed plugins and display [Installed] if found

### LINKED ISSUE #1028: CLI Needs A list option
As far as I can tell, there is no way to see which ignite packages are currently installed. Now that my app is gearing up for release, I would like to remove the ignite packages that I didn't end up using, but I don't know what they are.

Should there be an `ignite list` command added?

mikaoelitiana:

## Describe your PR

Add [Installed] label if ignite plugin is installed when user launch `ignite list`

Fixes #1028

### DIFF
--- a/src/lib/show-plugin-directory.ts
+++ b/src/lib/show-plugin-directory.ts
@@ -13,6 +13,7 @@ export default async function(toolbox: IgniteToolbox) {
   const { print, parameters } = toolbox
   const { colors, newline, info, table, error } = print
   const directory = await fetchPluginRegistry(toolbox)
+  const plugins = installedPlugins(toolbox)
   const search = parameters.first
 
   const pluginTable = pipe(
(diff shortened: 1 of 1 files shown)


src/lib/show-plugin-directory.ts
```diff
@@ -13,6 +13,7 @@ export default async function(toolbox: IgniteToolbox) {
   const { print, parameters } = toolbox
   const { colors, newline, info, table, error } = print
   const directory = await fetchPluginRegistry(toolbox)
+  const plugins = installedPlugins(toolbox)
```
jamonholmgren: You should be able to get the plugins by doing something like:

```
const plugins = toolbox.plugins.map(p => p.name)
```

Not in a place where I can test this now, but it should be more accurate than scanning the package.json dependencies.

Give that a shot and let me know how it goes! This is cool!

mikaoelitiana: @jamonholmgren I found this list in `toolbox.runtime.plugins` and I updated the PR. Thanks

Issue #1474, opened 2019-10-11
ignite add <package> needs a better error when package not found


jamonholmgren:
When running `ignite add push-notifications` (should be singular, not plural), I got this scary error:

![image](https://user-images.githubusercontent.com/1479215/66678181-27105080-ec20-11e9-9490-a86970221cc2.png)

We need to protect against this and provide nicer output to the developer.

arvidboivie:
I can take a look at this

PR #1475, opened 2019-10-11
Do not print import-plugin error, log instead

arvidboivie:

- [ not sure if applicable ] Everything works on iOS/Android

## Describe your PR

In response to #1474 i stopped printing the error message, logging it instead. This will hide it from everyday output while still making it easily available via the --debug flag


jamonholmgren:
Thank you @arvidboivie, simple fix!

Issue #1487, opened 2019-10-29
Add `ignite-cli` binary command to allow running with NPX


jamonholmgren:
We need to add a new binary in `./bin` called `./bin/ignite-cli` to allow Ignite to be run with NPX.

It can just require the other `./ignite`, which should work to trigger the CLI.

```js
#!/usr/bin/env node

require('./ignite')
```

Make sure to `chmod +x ./bin/ignite-cli`, and test it with `./bin/ignite-cli doctor` to ensure parameters are passed.

The `files` property in package.json shouldn't need to change, as it already includes `./bin` in full.

anusha5695:
@jamonholmgren : I would like to give it a try

anusha5695:
@jamonholmgren : PR: https://github.com/infinitered/ignite/pull/1492. Is this what is required?

PR #1492, opened 2019-10-30
Add ignite-cli binary command

anusha5695:
**Issue:**
This PR is linked to the issue: https://github.com/infinitered/ignite/issues/1487

**Summary of changes:**
Added a new file for ignite-cli under bin that refers ignite

### DIFF
--- a/bin/ignite-cli
+++ b/bin/ignite-cli
@@ -1,0 +1,3 @@
+#!/usr/bin/env node
+
+require('./ignite')


jamonholmgren:
@anusha5695 Thanks for doing this! Did you happen to chmod the binary so it's executable? I also forgot to ask you to add it to the package.json: https://github.com/infinitered/ignite/blob/8c780339ffa0c7b2626b381d492bfd40dfc826d0/package.json#L5-L7

anusha5695:
Yes i did do chmod. And will add it to package.json 👍

jamonholmgren:
@anusha5695 This makes me think maybe we don't even need the other binary? Just this?

```
 "bin": { 
   "ignite": "bin/ignite",
   "ignite-cli": "bin/ignite" 
 }, 
```

anusha5695:
@jamonholmgren : If thats the case would we even need a second command. We could link ignite with npx as well?

jamonholmgren:
@anusha5695 I agree, we wouldn't need the second command. We do need to make sure `ignite` and `ignite-cli` point to the same binary.

ahmdsolmn:
@jamonholmgren @anusha5695 Any updates here ?

PR #1507, opened 2019-11-18
chore: add install cocoapods extension

Gorhom:

## Describe your PR
Hi there ✋,

While building a boilerplate, i notice that `ignite` won't run `pod install`  before the initialise git commit. 

So i added a new extension `installCocoapods` to `ignite tools`, it runs `pod install` on the iOS project folder. Currently it triggered on the `new` command before initialise git. However, we could let the boilerplate trigger it.

Thanks :)

### DIFF
--- a/src/commands/new.ts
+++ b/src/commands/new.ts
@@ -6,6 +6,8 @@ import addEmptyBoilerplate from '../lib/add-empty-boilerplate'
 import boilerplateInstall from '../lib/boilerplate-install'
 import { IgniteToolbox } from '../types'
 
+const isMac = process.platform === 'darwin'
+
 /**
  * Creates a new ignite project based on an optional boilerplate.
  */
@@ -15,7 +17,7 @@ module.exports = {
   run: async function command(toolbox: IgniteToolbox) {
     const { parameters, strings, print, filesystem, system, ignite, prompt, runtime, meta } = toolbox
     const { isBlank, upperFirst, camelCase } = strings
-    const { log } = ignite
+    const { log, installCocoapods } = ignite
 
     // grab the project name
     const projectName = (parameters.first || '').toString()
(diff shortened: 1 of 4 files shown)


src/commands/new.ts
```diff
@@ -203,6 +205,11 @@ module.exports = {
     await system.run(yarnOrNPM)
     spinner.succeed(`${yarnOrNPM} complete`)
 
+    // run install cocoapods
+    if (isMac === true) {
+      await installCocoapods()
+    }
```
jamonholmgren: Rather than running it here, I think it makes sense to provide it to the boilerplate and have the boilerplate itself run this command.

For example, in Bowser we already do something similar, but it would be good to include the functionality in Ignite CLI so each boilerplate doesn't have to reimplement it.

During install, we create a `postInstall` hook to run a bunch of stuff, including `pod install`: https://github.com/infinitered/ignite-bowser/blob/cfa6d3f58716977793075dd6789dc90f6ac8537b/src/boilerplate.ts#L254

Running yarn triggers pod install: https://github.com/infinitered/ignite-bowser/blob/cfa6d3f58716977793075dd6789dc90f6ac8537b/src/boilerplate.ts#L264-L267

postInstall hook: https://github.com/infinitered/ignite-bowser/blob/cfa6d3f58716977793075dd6789dc90f6ac8537b/boilerplate/bin/postInstall#L24

jamonholmgren: I'm not entirely sure what the right process might look like here. A postinstall hook for running `pod install` is a good general solution I think.

Gorhom: Thanks for taking a time to response :)

I think i have missed that, and also i did not put platform `web` in consideration. i'll look at `Bowser` boilerplate again

You could close this pr

PR #1526, opened 2020-01-14
Switch to using npx for Ignite and react-native commands

jamonholmgren:
This will use NPX to invoke external yarn/npm commands.


.circleci/config.yml
```diff
@@ -46,9 +46,6 @@ jobs:
       - run:
           name: Change Permissions
           command: sudo chown -R $(whoami) /usr/local
-      - run:
-          name: Install React Native CLI
-          command: npm i -g react-native-cli
```
jamonholmgren: No longer needed due to npx usage.

PR #1563, opened 2020-08-12
Fix shelljs warnings & test failures

bryanstearns:

## Describe your PR
When I run commands like `ignite generate` in a Bowser app, I was getting a screenful of warnings:
```
$ ignite g
(node:44309) Warning: Accessing non-existent property 'cat' of module exports inside circular dependency
(Use `node --trace-warnings ...` to show where the warning was created)
(node:44309) Warning: Accessing non-existent property 'cd' of module exports inside circular dependency
(node:44309) Warning: Accessing non-existent property 'chmod' of module exports inside circular dependency
(node:44309) Warning: Accessing non-existent property 'cp' of module exports inside circular dependency
[...]
```

Traced this to [this issue](https://github.com/nodejs/node/issues/32987#issuecomment-635726114) - turns out our problem was also a need to upgrade to a newer `shelljs`, which fixed the warnings.

The ignite new-bowser test was also failing because of (my) recent changes to ignite-bowser (5.3.0), so I fixed those tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "ramda": "0.26.1",
     "ramdasauce": "2.1.3",
     "semver": "6.3.0",
-    "shelljs": "0.8.3",
+    "shelljs": "0.8.4",
     "which": "1.3.1"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


jamonholmgren:
@bryanstearns We honestly could be using Gluegun's built-in `system` tools for everything shelljs is doing in Ignite. Same for `which`.

PR #1569, opened 2020-10-27
Fixes for `ignite new --expo` in Flame

bryanstearns:
Fix `ignite new AppName --expo` in Flame, which was failing because it tries to run `npm prepare` (in the context of the boilerplate's package.json) at some point before it's actually copied the boilerplate to the new app's directory... and since we have a "prepare" script in the boilerplate's package.json that runs `npm-run-all patch hack:*`, it tried to run that -- but since it hadn't installed dependencies yet, `npm-run-all` isn't available.

I fixed this by (in the --expo case) ~~copying the boilerplate to a temp directory and removing the "prepare" script from the copy's package.json before passing that temp path to `expo-cli`. I added the "prepare" script to the boilerplate's `package.expo.json` so it gets added back when we merge that in later.~~ removing the "prepare" script from the boilerplate's package.json, and manually adding it back after initial app creation.

Also:
- That prepare problem initially showed up as a failure when trying to merge the package.json files later in run -- because spawnProgress didn't notice that `expo-cli init` was failing - to make this more obvious in the future, I've modified it here to reject its promise (with the collected error output) when the command fails with non-zero status; gluegun's unhandled promise rejection handler will display that output, which is better than nothing, but we could make the `new.ts` `run` function be more proactive about it...

- To ensure that --expo app generation continues to work, I duplicated most of the existing app-generation test and made it exercise --expo app generation

- eslint (`yarn lint`) was complaining about an unused parameter in `generators.ts` - I shut it up.

### DIFF
--- a/src/commands/new.ts
+++ b/src/commands/new.ts
@@ -1,4 +1,3 @@
-import * as tempy from "tempy"
 import { GluegunToolbox } from "../types"
 import { spawnProgress } from "../tools/spawn"
 import { isAndroidInstalled } from "../tools/react-native"
@@ -31,20 +30,11 @@ export default {
     const cli = expo ? "expo-cli" : "react-native-cli"
     const bowserPath = path(`${meta.src}`, "..")
     const boilerplatePath = path(bowserPath, "boilerplate")
-    const boilerplateCopy = path(tempy.directory(), "boilerplate")
-    if (expo) {
-      filesystem.copy(boilerplatePath, boilerplateCopy)
-      const packageJsonPath = path(boilerplateCopy, "package.json")
-      const packageJson = filesystem.read(packageJsonPath, "json")
-      delete packageJson.scripts.prepare
-      filesystem.write(packageJsonPath, packageJson)
-    }
-
     const cliString = expo
-      ? `npx expo-cli init ${projectName} --template ${boilerplateCopy}`
+      ? `npx expo-cli init ${projectName} --template ${boilerplatePath}`
       : `npx react-native init ${projectName} --template ${bowserPath}`
 
-    log({ expo, cli, bowserPath, boilerplatePath, boilerplateCopy, cliString })
+    log({ expo, cli, bowserPath, boilerplatePath, cliString })
 
     // welcome everybody!
     p("\n")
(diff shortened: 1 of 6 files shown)


src/commands/new.ts
```diff
@@ -30,11 +31,20 @@ export default {
     const cli = expo ? "expo-cli" : "react-native-cli"
     const bowserPath = path(`${meta.src}`, "..")
     const boilerplatePath = path(bowserPath, "boilerplate")
+    const boilerplateCopy = path(tempy.directory(), "boilerplate")
+    if (expo) {
+      filesystem.copy(boilerplatePath, boilerplateCopy)
+      const packageJsonPath = path(boilerplateCopy, "package.json")
+      const packageJson = filesystem.read(packageJsonPath, "json")
+      delete packageJson.scripts.prepare
```
jamonholmgren, line 39: You won't need to do this based on our Slack conversation, but just FYI, there's a very cool way to handle this built-in to Gluegun!

```js
await toolbox.patching.update('package.json', pkgJson => {
  delete pkgJson.scripts.prepare
  return pkgJson
})
```

bryanstearns: Cool!

bryanstearns:
I ended up doing something slightly different than we discussed in Slack - instead of adding a new package.vanilla.json (which would have duplicated the "prepare" statement with a copy in package.expo.json), I made the package.json cleanup step unconditioned on --expo; a substep conditionally merges `package.expo.json` in.

(In doing this, I removed the code that parallelized a few of the cleanup steps, since the two directory removes and the package.json write are nearly instantaneous, and the logic to continue to parallelize them de-simplified the code a bit.)

src/commands/new.ts
```diff
@@ -72,23 +72,24 @@ export default {
       filesystem.copy(path(boilerplatePath, ".gitignore"), gitPath)
     }
 
-    // expo-specific changes
+    // Update package.json. Having a "prepare" script in package.json
+    // messes up expo-cli init above (it fails because npm-run-all hasn't been
+    // installed yet), so we're adding it here; we also need to merge in our
+    // extra expo stuff.
+    let packageJson = filesystem.read("package.json", "json")
+    packageJson.scripts.prepare = "npm-run-all patch hack:*"
```
jamonholmgren, line 80: Sorry to be nitpicky, but I think for future maintainability I'd prefer this to be in a separate `package.rncli.json` (or similar) so we know where this script line is coming from more intuitively.

I'm going to be pairing with a dev in a bit, so we might do that for you if that's OK.

jamonholmgren (approved):
Good to merge, although I will make a couple nitpicky changes after we merge it in. Thanks @bryanstearns !

PR #1573, opened 2020-12-02
Several Flame tweaks

bryanstearns:
- Expo needs Exponent.app downloaded for Detox; instead of doing this at app-generation time, we'll do if it hasn't been done, whenever we run Detox. This means we don't need to commit Exponent.app anymore, so it's now .gitignored. (Note: Detox still doesn't actually work on Expo - not sure why).

- There were a few `<%= ... %>` left in `package.json` (in the detox command lines); I replaced them with the boilerplate's app name there, and rename them to the real app name when we generate a new app. This let me try out the boilerplate "in place" - it works.

- Somewhere along the way, I screwed up the merge direction when merging expo stuff into `package.json` - fixed.

- When I tried the boilerplate in place, it updated some Podfile.lock versions... so I committed them.

- I replaced a few mentions of "bowser" to avoid confusion.


boilerplate/bin/downloadExpoApp.sh
```diff
@@ -1,18 +1,20 @@
 #!/bin/bash -e
 
-# taken from https://github.com/expo/with-detox-tests/blob/master/setup.sh
+# based on https://github.com/yaron1m/expo-detox-typescript-example/blob/master/scripts/dl_expo_bins
+APP_PATH=./bin/Exponent.app
```
jamonholmgren, line 4: Do we know that they'll be running this from the root folder? If not, we may want to add a check for that (see if `package.json` exists, for example).

boilerplate/bin/downloadExpoApp.sh
```diff
@@ -1,18 +1,20 @@
...
 
-# query expo.io to find most recent ipaUrl
-IPA_URL=`curl --silent --show-error https://expo.io/--/api/v2/versions |  python -c 'import sys, json; print json.load(sys.stdin)["iosUrl"]'`
+if [ ! -d $APP_PATH ]; then
```
jamonholmgren, line 6: Let's add an `echo` to indicate what we're doing, something like this:

```
echo "First time running Detox -- downloading Expo client app"
```

jamonholmgren (approved):
A couple minor changes and then we're golden

boilerplate/bin/downloadExpoApp.sh
bryanstearns: Good idea! Will do

boilerplate/bin/downloadExpoApp.sh
bryanstearns: 👍

boilerplate/bin/downloadExpoApp.sh
bryanstearns: TIL: Both `yarn` and `npm` execute this script in the project root even if the user runs `yarn test:e2e` or `npm run test:e2e` from a project subdirectory (but I'm still adding the check!).

bryanstearns: ✅

boilerplate/bin/downloadExpoApp.sh
bryanstearns: ✅

PR #1576, opened 2020-12-16
Unify typescript-eslint versions and config; fix eslint complaints; f…

bryanstearns:
Unify typescript-eslint versions and config; fix eslint complaints; format

Sync'd the eslint configs between ignite and the boilerplate

Added "root: true" to boilerplate's eslint config to suppress eslint
"Couldn't determine the plugin uniquely" error when linting the
boilerplate in place

When new-ing, we `yarn format` at the end to make sure ignite's updates
are formatted nicely.

When new-ing an Expo app, which uses `App.js` instead of `index.js`, we now
update that filename in `tsconfig.json`, to make lint on Expo happy

Prettier fights with eslint over a few things, so we always `yarn format`
after linting to keep things stable

Changed Ignite's Prettier line length to match the boilerplate (100)

Added lint/compile/format checks to Ignite's `new` tests

Automatic and manual Lint fixes:

- A few `var`s turned into `let`s

- `object` shouldn't be used as a type; found more precise types to use instead

- Used a regexp literal instead of calling the RegExp constructor

- Made a few `@ts-ignore`s unnecessary because eslint hates them

- Moved definitions and imports up in several places, to makes sure things
are defined before they're used


boilerplate/app/utils/validate.ts
```diff
@@ -70,7 +70,7 @@ export interface ValidationErrors {
  * @param rules The rules to apply.
  * @param data The object to validate.
  */
-export function validate(rules: ValidationRules, data: {}): ValidationErrors {
+export function validate(rules: ValidationRules, data: Record<string, unknown>): ValidationErrors {
```
jamonholmgren, line 73: pretty

bryanstearns: That's straight out of the lint error message, FYI:
```
Don't use `{}` as a type. `{}` actually means "any non-nullish value".
- If you want a type meaning "any object", you probably want `Record<string, unknown>` instead.
- If you want a type meaning "any value", you probably want `unknown` instead.
- If you want a type meaning "empty object", you probably want `Record<string, never>` instead  @typescript-eslint/ban-types
```

PR #1586, opened 2020-12-31
Show globally-installed packages in `ignite doctor`

bryanstearns:

## Describe your PR

@nirre7 had a problem that was caused by an older version of react-native-cli being installed globally (which is no longer recommended); he suggested having Ignite warn when this happens. https://infiniteredcommunity.slack.com/archives/CD8UA5RJ4/p1609336353033800?thread_ts=1609248567.028100&cid=CD8UA5RJ4

I think that's a great idea, and I thought it'd be useful as a preliminary step to see _all_ the globally-installed packages in `ignite doctor`'s output, so that's what this PR adds. Because `npm` and `yarn` store global packages in different places, it lists them separately -- the Javascript section of its output now looks like this (I just have one package in each group):

```
JavaScript (and globally-installed packages)
  node                14.15.1      /Users/stearns/.asdf/installs/nodejs/14.15.1/bin/node
  npm                 6.14.8       /Users/stearns/.asdf/installs/nodejs/14.15.1/bin/npm
    expo-cli          4.0.17
  yarn                1.22.5       /Users/stearns/.yarn/bin/yarn
    create-next-app   10.0.4
```

### DIFF
--- a/boilerplate/app/screens/demo/demo-screen.tsx
+++ b/boilerplate/app/screens/demo/demo-screen.tsx
@@ -78,7 +78,7 @@ const HINT: TextStyle = {
 }
 
 const platformCommand = Platform.select({
-  ios: "Cmd + D", 
+  ios: "Cmd + D",
   android: "Cmd/Ctrl + M",
 })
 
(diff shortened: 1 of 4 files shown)


jamonholmgren (approved):
Looks great. Tested on my machine and it works, including without Yarn installed.

PR #1600, opened 2021-01-22
chore: migrate to new react-native-async storage package

harrisrobin:

## Describe your PR
`react-native-async-storage` has moved away from `react-native-community` and into it's own organization, so we are currently pointing to a deprecated package. This PR updates all references to react-native-async-storage to use the new location at https://github.com/react-native-async-storage/async-storage. There are no breaking changes from my testing.


yulolimum:
@harrisrobin just a though but since we have `unimodules` integrates - we could just use: https://docs.expo.io/versions/v40.0.0/sdk/async-storage/

harrisrobin:
> @harrisrobin just a though but since we have `unimodules` integrated - we could just use: https://docs.expo.io/versions/v40.0.0/sdk/async-storage/

I like that idea. will look into it tomorrow

jamonholmgren (approved):
Nice!

jamonholmgren (changes requested):
Actually I like the idea of using the Expo unimodule, since that will work on Expo as well.

harrisrobin:
@yulolimum @jamonholmgren after looking at this again, seems like Expo simply points to `@react-native-async-storage/async-storage` so I believe there is nothing else to do here 😄

PR #1606, opened 2021-02-13
Use absolute dependency versions; bump Expo to SDK 40

bryanstearns:

- N/A `README.md` has been updated with your changes, if relevant 

## Describe your PR

I tried to spin up a new ignite --expo app, and it was broken because of the way our boilerplate specifies dependencies:
- We depend on mobx (“^6.0.4”, note the caret) and mobx-state-tree (“4.0.2", no caret).
- We got mobx 6.1.6 because the caret [“allows changes that do not modify the left-most non-zero element in the [major, minor, patch] tuple.”](https://github.com/npm/node-semver#caret-ranges-123-025-004)
- We got mobx-state-tree 4.0.2 because just specifying a plain version gets you that version.
- It seems that mobx 6.1.0 [broke something in mobx-state-tree 4.0.2](https://github.com/mobxjs/mobx-state-tree/issues/1653). This was [fixed by mobx-state-tree 5.0.1](https://github.com/mobxjs/mobx-state-tree/pull/1656)

It definitely seems wrong to use a range specifier for some but not all of our dependencies’ versions - but even if we’d allowed patch or minor updates to the mobx-state-tree dependency, we wouldn’t have gotten 5.0.1 because that’s a major version change.

This PR makes all the dependency versions absolute - though generated apps might not automatically get allegedly-safe updates this way, it ensures that generating an app months after we released the last Ignite version will still get the same dependency versions we tested with.

I've verified that a generated vanilla app's yarn.lock exactly matches boilerplate/yarn.lock. I can't do the same thing with a generated Expo app because we don't have a record of what exact versions we used when we tested 6.1.1 - so I bumped Expo to SDK 40 (after good experience upgrading one of our other projects) and it tested good.


jamonholmgren (approved):
Nice!

PR #1607, opened 2021-02-14
Update package.json to fix error related to mobx

preetb123:
Adding fix for https://github.com/mobxjs/mobx-state-tree/issues/1653

Screenshot_1613281315

`mobx-state-tree` version 4.0.2 gave above error.

### DIFF
--- a/boilerplate/package.json
+++ b/boilerplate/package.json
@@ -37,7 +37,7 @@
     "i18n-js": "3.8.0",
     "mobx": "6.0.4",
     "mobx-react-lite": "3.1.6",
-    "mobx-state-tree": "4.0.2",
+    "mobx-state-tree": "^5.0.1",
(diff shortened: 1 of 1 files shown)


bryanstearns:
Hi @preetb123 - thanks for the PR, but we decided to go "the other way" to fix this - instead of using a newer `mobx-state-tree` version, we realized that using "flexible" version specifications (using "^" or "~") in generated apps meant that people would generate apps using dependency versions that we hadn't tested (which leads to most of the issues filed here!).

So, as of ignite v6.2.0, we now use non-flexible versions for all dependencies. Yes, this means that some apps will start life with slightly older versions of some dependencies, but people can upgrade dependencies when they're ready to - when they do, they'll know that it was related to upgrading, and not think that Ignite is just broken. :-)

Issue #1605, opened 2021-02-10
Add example list screen with proper API fetch


jamonholmgren:
I received feedback via a private message that the documentation for making and handling API calls is inadequate.

Looking at the actual code, he has a point. This is pretty awful and about the only API code in the app.

https://github.com/infinitered/ignite/blob/465f3c15a5e9f7cc29376a1d2ede6cc57c260242/boilerplate/app/screens/demo/demo-screen.tsx#L118-L124

*I'd like to add another page that fetches info and displays it in a list.*

Recommendation would be to fetch the data in a static JSON file hosted on Github and display the data in a listview.

bryanstearns:
@jamonholmgren I agree that more examples would be useful (especially this one!), but we've already heard that people don't like having more stuff in generated apps that needs to be ripped out.

Expo uses [a separate repo for examples](https://github.com/expo/examples) - what would you think about us setting up a new ignite-examples repo and adding this there?

syahnur197:
maybe you guys can have an example repository, i.e., a todo app? That would be useful

jamonholmgren:
We've had example apps, but they're almost impossible to maintain properly. They end up stagnating.

Having it in the generated app forces us to confront how things are architected and makes us stay on top of them.

I have heard the feedback that people don't like having to delete stuff, but I think as long as they're well-defined and you don't have to go through a lot of files to pull them out, it's not too big of a hardship. I'm not talking about an entire kitchen sink, but certainly the most common use cases (login form, API call) make sense.

chrispecoraro:
@bryanstearns Minimally, the main issue here is _this_ line, which is problematic if you're trying to demonstrate a very, opinionated pattern. It should be changed and replaced with something else. I, as developer new to react native, understand, conceptually, what a _the store's API_ could represent, but _how_ is data stored and more importantly, how is it retrieved on another screen.
` // Don't do API like this, use store's API `.

Also, now that I have the metadata for a user with id 1, I want to "do some async storage stuff", so how do I store it using `await save`. 
```
demo.getUser("1") 
 // Let's do some async storage stuff 
 await save("Cool Name", "Boaty McBoatface") 

```
For context, I'm finishing up a book on Jamstack with Gatsby and the reviewers, for example, would ask me this. It's not clear if your audience is high-level React Native developers or developers from other tech stacks wanting to start into ReactJS.

vantuan88291:
[...]

PR #1608, opened 2021-02-16
Add a demo list screen to boilerplate

kateinkim:

## Describe your PR

Per https://github.com/infinitered/ignite/issues/1605 adds a demo list screen that fetches data from a public API. Not sure if this is the right place to put it per that discussion but I'm happy to move it wherever it should be.

Simulator Screen Shot - iPhone 11 - 2021-02-15 at 21 00 42


vantuan88291:
I don't think this demo is useful. You should fetch api and use model to store data from api. This is **mobx-state-tree**

chrispecoraro:
@kateinkim, yes, I agree with @vantuan88291. This demo is great and is badly needed, but what I'm interested in learning is how to use the MST with the API and then how to pass the data in the MST between screens using navigation, so actually, there should be two screens involved. I wish I could help, but I'm still learning.

vantuan88291:
> @kateinkim, yes, I agree with @vantuan88291. This demo is great and is badly needed, but what I'm interested in learning is how to use the MST with the API and then how to pass the data in the MST between screens using navigation, so actually, there should be two screens involved. I wish I could help, but I'm still learning.

i have some repo on my git using mobx state tree, you can refer it

kateinkim:
@chrispecoraro @vantuan88291 I hear you. I'm still learning myself. I was trying to balance between adding the example and cutting down on the need to delete across multiple files per the original issue discussion. I can definitely see how it would be more useful to include MST integration. Maybe @jamonholmgren and/or @bryanstearns can chime in with some guidance?

syahnur197:
@chrispecoraro @kateinkim I've added an example on how to use the demo app with MST and API Sauce as recommended by the boilerplate itself. From what I understand, you can just import the `CharacterStore` into anywhere in your screens and component, you would have access to the characters in the store

kateinkim:
@syahnur197 Thanks for the MST help. 😄 I made some additional updates at @jamonholmgren's suggestion.

jamonholmgren (approved):
This looks really clean. Nice job @kateinkim !

leonskim (approved):
Nicely done! This is a neat example that would really help who just stepped in RN, Ignite and MST. Thanks for your contributions! 🥇

Issue #1610, opened 2021-02-18
Add newly generated component to components/index.ts exports


fvonhoven:
What's going on?

When generating a new component with Ignite, the component is not added to the exports in `components/index.ts`

Steps to reproduce

1. Ignite a new project
2. `npx ignite-cli g component NewlyGenerated`
3. Notice that the screen is not added to the exports in `components/index.ts`

---

`ignite doctor` results:
N/A

jamonholmgren:
Fixed with #1617.

fabtrompet:
I have the same problem, when creating model

![image](https://user-images.githubusercontent.com/6043140/111240723-85dab600-85da-11eb-8061-f9601bb2e3f4.png)

In 6.5.0 Version

Issue #1609, opened 2021-02-18
Add newly generated screen to screens/index.ts exports


fvonhoven:
What's going on?

When generating a new screen with Ignite, the screen is not added to the exports in `screens/index.ts`

Steps to reproduce

1. Ignite a new project
2. `npx ignite-cli g screen NewlyGenerated`
3. Notice that the screen is not added to the exports in `screens/index.ts`

---

`ignite doctor` results:
N/A

jamonholmgren:
Fixed with #1617

PR #1617, opened 2021-02-24
Update generators to include barrel exports automatically

kateinkim:
Addresses: #1609 and #1610 

Updates `generateFromTemplate` to append an index file if it exists and the file is one likely to want exporting 😉 

Assumptions made:
* Index file is ts, tsx, js, or jsx and should include exports
* Files including `.test` or `.story` shouldn't be included in exports (could there be others?)


jamonholmgren:
This is awesome.

3 things:

1. Could we have an option for skipping this? `--skip-index-file` comes to mind, but other ideas welcome
2. Any documentation need to be updated for this?
3. Any tests could be updated for this? [Here's a good candidate](https://github.com/infinitered/ignite/blob/8cbcfd6e58f918c3ff477d2e6bec147725b1b645/test/ignite-new.test.ts#L95-L102)

kateinkim:
Thanks Jamon!

1. I added the skip option. I liked your name.
2. I just updated one line in the docs. Not sure if there's something else I should add.
3. I updated that test to check for the appended export but the test seems to hang so I can't verify if it's passing. Any ideas?

jamonholmgren:
Current CI failure is related to https://github.com/infinitered/solidarity/issues/263.

jamonholmgren:
I'm not entirely sure what the latest test failure is, tbh.

kateinkim:
@jamonholmgren I figured it out. Just needed to configure the user for git before the test runs. Apparently circle doesn't have a git user setup by default. Tests are passing now!

High five

jamonholmgren:
Yay!! Thanks @kateinkim . I'll release this soon.

Issue #1611, opened 2021-02-18
Generating a new navigator creates a sibling folder to `navigation/`


fvonhoven:
What's going on?

*  On first time generating a navigator creates a `navigators/newly-generated/navigator` folder and sub folder as a sibling to `/navigation` , is this desired?  
* Additionally, there is no index.ts in the newly created `navigators/` folder. 
* Perhaps make `navigators/` a child of `navigation/`? 

Seems a little confusing to have similarly named sibling folders.

Steps to reproduce

1. `npx ignite-cli g navigator MyNewNavigator`
2. Observe sibling folders `navigation/` & `navigators/`

Screenshot of folders:
![image](https://user-images.githubusercontent.com/10098988/108294041-b30e8280-715a-11eb-854a-f908d6317069.png)

---

`ignite doctor` results:
N/A

jamonholmgren:
I think @kateinkim is working on something that might address this.

PR #1629, opened 2021-03-13
Rename the navigation folder to navigators.

mcarcaso:
This will allow it to play nicer with the ignite cli to generate navigators.

## Please verify the following:

- [x] `yarn ci:test` **jest** tests pass with new tests, if relevant
- [x] `README.md` has been updated with your changes, if relevant

## Describe your PR

When running `npx ignite-cli g navigator Foo`, it creates a new `app/navigators` directory and generates the files in there but the ignite template generates the initial set of navigators in the `app/navigation` directory. This PR renames the `app/navigation` to `app/navigators` so it plays nicer with that generator out of the box.

PR #1630, opened 2021-03-15
Add front matter support to templates

kateinkim:
Per #1611 templates currently only support copying their contents to app directories with a pluralized name of the template directory. Since ignite puts navigators in the `app/navigation` directory we need a way to customize this. 

By using front matter we can specify a `destinationDir` at the top of the template, customizing per template file. This also opens up the possibility for more powerful generator customization in the future.


jamonholmgren:
Thanks for this PR, @kateinkim! It's super well done.

So, #1629 addresses the current problem, which is the `navigation` / `navigators` naming issue brought up in #1611.

I like the idea of frontmatter in general. However, I'm putting this PR "on hold" for now until we see another situation where it will be directly useful. I'll keep it open.

_One note: I don't want to use `gray-matter`. If we can use a different frontmatter library that would be good. I don't want Ignite associated with the author of that library._

kateinkim:
Thanks @jamonholmgren! I'll get the library updated.

jamonholmgren:
@kateinkim We already have the `yaml` library in our dependencies. I'm sorta considering us rolling our own frontmatter ... basically, see if the file contains two lines that are just `---`, and if so, parse what's between as yaml, and remove that chunk for rendering purposes. Just thinking out loud, though -- if you find a good (small) frontmatter library that might be all we need.

kateinkim:
@jamonholmgren I updated to remove the gray-matter dependency and added support for patching from templates. I added patches to the model template to automatically import and add stores to the root store.

kateinkim:
Hey @jamonholmgren! I updated to support append, prepend, and replace as well and refactored the existing barrel export handling with it (still supports the skipping option). I also moved the patching handling out into it's own function. Let me know what you think.

gavinharris-dev:
Would this PR also allow us to create a custom generator for `api`? The folder structure would be to place the generated 'api' files within the `services/api` sub-directory of `app`?

kateinkim:
@gavinharris-dev Yes! You’d be able to target any path in the project. 🙂

kateinkim:
> @kateinkim When you get a chance, can you resolve conflicts on this? No rush.

Yep! I will take a look at this later today!

kateinkim:
Whoa, it took long time to remember what and how I did this thing after coming back a year later. It's all tested and working now! 😄 

Screen Shot 2022-04-25 at 7 57 23 PM

jamonholmgren:
Since this is working, let's target v8: Maverick for this @kateinkim. Can you resolve conflicts and re-target this at the `maverick` branch?

kateinkim:
Well, It wasn't easy to fix the conflicts but it's working as it is supposed to be still. @jamonholmgren

jamonholmgren (approved):
Looks good!

PR #1640, opened 2021-03-26
Add tx types for autocomplete

kateinkim:
I noticed that none of the tx props were autocompleting. I updated the the tx prop types to use a recursive type built from the default locales json. Now translation keys will autocomplete in VSCode:

Screen Shot 2021-03-25 at 7 44 13 PM


jamonholmgren (approved):
Now that's a brilliant PR. Nice job @kateinkim.

PR #1644, opened 2021-03-29
Add keyboardShouldPersistTaps to screen

kateinkim:
Exposes ScrollView's keyboardShouldPersistTaps prop on Screen component. Sets "handled" as a reasonable default.


jamonholmgren:
Not sure why CI failed, as it shouldn't make much difference to this. I'll re-run it.

kateinkim:
I couldn't understand why it was failing but could not re-run the test. I'm glad it passed finally. Thanks!

jamonholmgren (approved):

PR #1650, opened 2021-04-05
Fix keyboard avoid view behaviour

dani-z:

## Describe your PR
Fix the values passed to KeyboardAvoidingView `behaviour` prop.

As per React Native documentation `behaviour` is an enum ([link](https://reactnative.dev/docs/keyboardavoidingview#behavior)) which means one of the values must be provided. In case there is no need to use behaviour, the correct value should be `undefined`.

Please see screenshot below for the TS warning:
image


jamonholmgren (approved):
Nice, thanks!

PR #1656, opened 2021-04-18
Fix style prop types

kateinkim:
I discovered that I could not use false in array styles while working on a top secret project:

Screen Shot 2021-04-17 at 7 34 29 PM

By using react native's `StyleProp` generic we can support array styles just like base components do:

Screen Shot 2021-04-17 at 7 37 53 PM

This also meant a bunch of `mergeAll`s could be removed which I think cleans up the components a little bit.


jamonholmgren (approved):
Good catch! I like this, should improve the DX. TY!

PR #1657, opened 2021-04-19
Add groups nesting navigation, upgrade to react native screens v3

vantuan88291:

## Describe your PR
- Add groups nesting navigation to object
- Add Middleware for routes, handle routes before navigate
- Add dispatch action to navigation-utilities
- Upgrade to react native screens v3, remove enableScreens(), its will enable by default, reference: https://github.com/software-mansion/react-native-screens/releases/tag/3.0.0


vantuan88291:
> @vantuan88291 Thanks for taking the time to do this PR! I have a couple notes.
> 
> 1. I like the react-native-screens update; I'd love it if you'd submit that as a separate PR
> 2. The routes change doesn't appear to add enough value to justify the additional abstraction. Can you give me an example where this would make more sense?
> 
> (BTW, we have automated version number bumping, so you don't need to do that manually.)

Hi, the first one i will make an other PR. About the second, do you want to make an example inside template or example right here? Because i think make so many examples inside template will make it complicated

harrisrobin:
Now that we merged #1758 , I think we can close this PR. To achieve something similar to the groups suggested here, we can now use Stack.Groups since we upgraded to React Navigation V6.

Issue #1619, opened 2021-02-25
Feature Request: Loading indicator for ignite cli


MichaelDimmitt:
What's going on?

When the cli is running in a terminal it is working on the problem but there is no loading indicator to tell the developer to wait. This could lead the developer to assume that the process is in an infinate loop of some sort and will never finish causing them to kill the process and start again.

---

Steps to reproduce

1. npx ignite-cli new MyApp --expo

---

I made an initial attempt cloning the project locally and which was to run the cli with a dot printing for every second of inactivity. The final result looked like the following:

![run-with-dots-part1](https://user-images.githubusercontent.com/11463275/109172652-02ebdb80-7751-11eb-8d3c-30fddd0cdfc9.png)

MichaelDimmitt:
Feel free to close, I just wanted to see if there was interest to have some form of loading spinner added to the project.

jamonholmgren:
@MichaelDimmitt I'd love to have one, just haven't put the time in to add it! I like your idea!

There is the `print.spinner()` command available in Gluegun. I haven't looked at whether it makes sense to incorporate it

MichaelDimmitt:
Thanks for the advice @jamonholmgren , 
I will take a look at print.spin and give draft pr's of both implementations.
If that sounds good.

kateinkim:
Hey @MichaelDimmitt! Are you working on this or would it be okay if I picked it up? Thanks!

MichaelDimmitt:
Sure @kateinkim 

eager to see what you come up with 🙂

mazenchami:
@MichaelDimmitt just tested with the latest version (v6.12.2), Igniting a new app will show a loading spinner thanks to PR #1660 (thanks @kateinkim).

MichaelDimmitt:
yes, looks good ! 

thanks @kateinkim

PR #1660, opened 2021-04-23
Add spinner to new steps

kateinkim:
Addresses https://github.com/infinitered/ignite/issues/1619

Adds loading spinner to ignite new command steps (with emojis).

Demo @ 60x speed:

https://user-images.githubusercontent.com/53795920/115897204-688ada00-a419-11eb-8605-8b8327ff3a79.mov

### DIFF
--- a/src/tools/pretty.ts
+++ b/src/tools/pretty.ts
@@ -1,12 +1,36 @@
-import { print } from "gluegun/print"
+import { print } from "gluegun"
 
 const { cyan, gray, white, bold, red, yellow } = print.colors
 const { underline } = print.colors
 
+type Spinner = ReturnType<typeof print.spin>
+const spinners: { [key: string]: Spinner } = {}
+
 export const p = (m = "") => print.info(gray(`   ${m}`))
+
+export const startSpinner = (m = "") => {
+  let spinner = spinners[m]
+  if (!spinner) {
+    spinner = print.spin({ prefixText: "   ", text: gray(m) })
+    spinners[m] = spinner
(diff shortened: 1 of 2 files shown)


src/tools/pretty.ts
```diff
@@ -1,12 +1,36 @@
-import { print } from "gluegun/print"
+import { print } from "gluegun"
```
jamonholmgren, line 1: Was this not working with `gluegun/print`? If so, that might be a bug with Gluegun.

src/tools/pretty.ts
```diff
@@ -1,12 +1,36 @@
...
 
 const { cyan, gray, white, bold, red, yellow } = print.colors
 const { underline } = print.colors
 
+type Spinner = ReturnType<typeof print.spin>
+const spinners: { [key: string]: Spinner } = {}
```
jamonholmgren, line 7: I like the idea of having multiple spinners at once and tracking them.

jamonholmgren (approved):
This looks great to me. Nice job Kate! Excited to finally have nice spinners in the new version of Ignite!

src/tools/pretty.ts
kateinkim: It works but is untyped. I can switch it back but I thought it was nice to have the type safety.

jamonholmgren: Ooh, didn't realize it was untyped!! That's something we'll want to fix eventually on Gluegun. Thanks for pointing it out!

PR #1680, opened 2021-05-12
fix(boilerplate): make demo flatlist re-render on mst store data change

ilovepku:

## Describe your PR

Converts ObservableArray to Array with a spread to make demo FlatList re-render on mst store data change.


jamonholmgren (approved):
Cool, thanks @ilovepku ! This will work for now.

The canonical way to do this would be to wrap each renderItem in `observer` to catch changes to existing elements and also provide an `extraData={characters.length}` to catch adding or removing elements.

I've even thought about making an `MSTFlatList` that does this for you. But that's beyond the scope of what we're doing here. Thank you!

Issue #1632, opened 2021-03-15
Detect when `ignite new` would overwrite a folder and fail unless `--overwrite` is provided


jamonholmgren:
When running `ignite new Foo`, if there's already a `Foo` folder, Ignite will either fail (in the case of `--expo`) or just skip past the `react-native init` process (when not using Expo).

We should error with something like:

```
Error: There's already a folder with the name "Foo". To force overwriting that folder, run with `--overwrite`.
```

If they run it with `--overwrite`, we should delete any existing folder with that name.

mazenchami:
Closing this issue as since v6.11.0 (currently on v6.12.2), this feature is available. Example: `npx ignite-cli new PizzaApp --overwrite)

PR #1682, opened 2021-05-14
Add overwrite option to new

kateinkim:
Addresses: https://github.com/infinitered/ignite/issues/1632

Adds a check for an existing directory when running new. Throws if directory exists unless an overwrite flag is passed in which case it will remove the existing directory before continuing.


jamonholmgren (approved):
Looks great!

This reminds me that we need CLI documentation in this folder:

https://github.com/infinitered/ignite/tree/master/docs

PR #1683, opened 2021-05-17
Use internal StyleSheet.flatten instead ramda

pronevich:


jamonholmgren (approved):
Nice! This will work well. Thanks @pronevich !

pronevich:
Yeah, I made PR early today) #1685

pronevich:
Also, before PR I have checked that RN styles will flatten for any nesting level.

jamonholmgren:
That's awesome! Thanks @pronevich !

PR #1684, opened 2021-05-18
Cut out ramda

pronevich:
Using ramda depends on team/project and can be avoided from using in boilerplate.
In all cases where imported ramda utils, we can use simple plain js:
1. As `preset` is runtime type of string `!preset` - will cover all from ramda isNil - undefined and null, plus !preset.length. Same for !presets[preset]
2. ramda `contains` is deprecated and not needed.


jamonholmgren:
I've been wanting to drop Ramda for a while, so this is very appreciated @pronevich !

PR #1685, opened 2021-05-19
Drop redundant flatten. RN flatten styles by default

pronevich:
Thanks, @satya164 for a hint.


markrickert:
This is awesome! We definitely don't need to be flattening those when we could either just send an array to the style prop or splat them into place.

boilerplate/app/components/switch/switch.tsx
```diff
@@ -91,20 +85,24 @@ export function Switch(props: SwitchProps) {
     outputRange: [OFF_POSITION, ON_POSITION],
   })
 
-  const style = enhance({}, props.style)
```
jamonholmgren, line 94: This was kinda weird anyway!

jamonholmgren (approved):
Excellent. :100:

PR #1688, opened 2021-05-25
Discuss: Apply insetStyle to outer view

pronevich:
I think `insetStyle` should be applied to outer view in _ScreenWithoutScrolling_ same as in _ScreenWithScrolling_.

Then styles defined in _inner preset_ like `paddingBottom` specific for app design won't overrided by `insetStyle`. Because for example in most cases we want our design paddingBottom: 20 + safe paddingBottom: 34 = 54  for iPhone 12 and 20 + 0 for iPhone SE and so on.

### DIFF
--- a/boilerplate/app/components/screen/screen.tsx
+++ b/boilerplate/app/components/screen/screen.tsx
@@ -15,12 +15,12 @@ function ScreenWithoutScrolling(props: ScreenProps) {
 
   return (
     <KeyboardAvoidingView
-      style={[preset.outer, backgroundStyle]}
+      style={[preset.outer, backgroundStyle, insetStyle]}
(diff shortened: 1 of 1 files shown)


pronevich:
But after this, it's harder to style when we need some content under StatusBar for example, because safe paddingTop applied to the outer view and shift our inner content view

pronevich:
Now I think it's a possibly wrong idea. And also find out that when `behavior={isIos ? "padding" : undefined}` outer `paddingBotom` not applied, for behavior="height" - `paddingBottom` applied.

rashd640 (approved):

jamonholmgren:
@pronevich After a month, do you still think this is the wrong approach? I could get others to weigh in if you think it's worthwhile.

PR #1691, opened 2021-05-30
Update Generators.md documentation

jsflor:
Change all "ignite" commands to "ignite-cli"


nirre7:
Thank you for helping us keeping the docs up to date!
I think it is even better if we add the `npx` i.e `npx ignite-cli generate component MyNewComponent`.
That way we might get less environment related issues 🤞

jamonholmgren (approved):
Ooh, good catch.

I agree with @nirre7 about the npx additions, but those can be a separate PR.

jsflor:
Absolutely @nirre7! @jamonholmgren let me handle this! 😄

PR #1663, opened 2021-04-24
fix(windows): Fix issue #1639 -- remove "file://" from template address

aliassar:
## Please verify the following:

- [ ] `yarn ci:test` **jest** tests pass with new tests, if relevant
- [ ] `README.md` has been updated with your changes, if relevant

## Describe your PR

Fix problems for windows caused by 'file://' syntax that should only be used in linux

Issue #1639, opened 2021-03-23
Error in create new app.


araymoises:
Error in create new app.
[
![Sin título](https://user-images.githubusercontent.com/22068894/112232838-0ec7a200-8c18-11eb-8782-7f9ed02beae3.png)
](url)
---
I'm new with Ignite, i want to create a new project and it gives me this error. Any solution?

Steps to reproduce

1. npx ignite-cli new APP

---

jascination:
Might be the wrong issue, but do you have `yarn` installed? I had similar errors when my environment wasn't set up properly (I had wonky versions of Node via both NVM and Homebrew installed and it was confusing the system when I used `npx`).

Can you try making an expo project (like `npx ignite-cli new PizzaApp --expo`) and see if it still fails?

andrewlorenz:
I have exactly the same issue - actually had it 2 weeks ago and gave up with ignite, but trying again now in some spare time.
```
npx ignite-cli new ignited
   · · · · · · · · · · · · · · · · · · 🔥 Ignite 🔥 · · · · · · · · · · · · · · · · · ·

    █ Creating ignited using Ignite 6.6.0
    █ Powered by Infinite Red - https://infinite.red
    █ Using react-native-cli
    ────────────────────────────────────────────────

   🔥 Igniting app
   🖨  3D-printing a new React Native app
C:\Users\user\AppData\Roaming\npm-cache\_npx\3460\node_modules\ignite-cli\node_modules\gluegun\build\index.js:13
    throw up;
    ^

TypeError: Cannot read property 'replace' of undefined
    at C:\Users\user\AppData\Roaming\npm-cache\_npx\3460\node_modules\ignite-cli\build\commands\new.js:137:26
    at step (C:\Users\user\AppData\Roaming\npm-cache\_npx\3460\node_modules\ignite-cli\build\commands\new.js:44:23)
    at Object.next (C:\Users\user\AppData\Roaming\npm-cache\_npx\3460\node_modules\ignite-cli\build\commands\new.js:25:53)
    at fulfilled (C:\Users\user\AppData\Roaming\npm-cache\_npx\3460\node_modules\ignite-cli\build\commands\new.js:16:58)    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```
Yes I have yarn:
```
> yarn -v
1.22.5
```
Does the same work with --expo ?  yes:
```
> npx ignite-cli new ignited --expo
[...]

PR #1696, opened 2021-06-02
Use @react-native-community/cli instead of react-native-cli

jamonholmgren:
First, this switches from `react-native-cli` to `@react-native-community/cli`.

Secondly, this removes the `file:///` prefix schema that seems to have caused problems in Windows (#1639 and #1663). I'm not entirely sure why it was necessary in the first place.

And lastly, this refactors the `new.ts` command a bit to use a CLI string builder, and also adds some tests. It should make it easier to add options going forward (which I'll need in order to support other platforms).

It shouldn't affect much, so it can be a patch-level release, assuming everything works of course.

### DIFF
--- a/src/tools/react-native.ts
+++ b/src/tools/react-native.ts
@@ -1,4 +1,4 @@
-import { GluegunParameters, GluegunToolbox, filesystem } from "gluegun"
+import { GluegunToolbox, filesystem } from "gluegun"
 
 // export const isWindows = process.platform === "win32"
 
--- a/src/tools/react-native.ts
+++ b/src/tools/react-native.ts
@@ -1,4 +1,6 @@
-import { GluegunToolbox } from "gluegun"
+import { GluegunParameters, GluegunToolbox, filesystem } from "gluegun"
+
+// export const isWindows = process.platform === "win32"
 
 export const isAndroidInstalled = (toolbox: GluegunToolbox): boolean => {
   const androidHome = process.env.ANDROID_HOME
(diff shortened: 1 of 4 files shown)


src/tools/react-native.ts
```diff
@@ -1,4 +1,6 @@
-import { GluegunToolbox } from "gluegun"
+import { GluegunToolbox, filesystem } from "gluegun"
+
+// export const isWindows = process.platform === "win32"
```
markrickert, line 3: Can we completely remove this?

src/tools/react-native.ts
```diff
@@ -7,3 +9,45 @@ export const isAndroidInstalled = (toolbox: GluegunToolbox): boolean => {
 
   return Boolean(hasAndroid)
 }
+
+type Params = {
```
markrickert, line 13: is `Params` too generic of a name?

markrickert (approved):
Nice! :shipit:

src/tools/react-native.ts
jamonholmgren: Turns out, no. We need to keep the `file:///` with the new CLI. Working on it!

src/tools/react-native.ts
jamonholmgren: I'll be honest, I was trying to keep it short so it wouldn't word-wrap the function definition. :laughing:

jamonholmgren:
I'm really running into issues with this. We might have to rethink our strategy. I don't understand what changed and why this is now failing. Even `master` is failing.

jamonholmgren:
Closing in favor of #1697.

PR #1697, opened 2021-06-03
Remove react-native-cli and copy over files ourselves

jamonholmgren:
This was spurred on by some (apparently local?) problems I was having with the react-native CLI. However, I think it's worth a look even if react-native-cli works fine.

This PR removes our dependency on `@react-native-community/cli` aka `react-native-cli` and instead undertakes copying over the boilerplate files ourselves (when not using Expo).

Since the boilerplate is essentially already a runnable project, this is a pretty straightforward copy operation. We still do some cleanup by removing any stray Expo files that make it in and also run `npx react-native-rename` on it -- since your app is probably not named `HelloWorld`.

It seems that most of the time spinning up a new Ignite app is still when it's `yarn install`ing. We probably need to spend time stripping our dependencies down a bit more.

This *shouldn't* make any breaking changes to the boilerplate. This would also supplant #1696 if we merge it.

### DIFF
--- a/src/commands/new.ts
+++ b/src/commands/new.ts
@@ -1,6 +1,6 @@
 import { GluegunToolbox } from "../types"
 import { spawnProgress } from "../tools/spawn"
-import { isAndroidInstalled } from "../tools/react-native"
+import { isAndroidInstalled, copyBoilerplate } from "../tools/react-native"
 import { packager } from "../tools/packager"
 import {
   command,
@@ -10,9 +10,8 @@ import {
   p,
   startSpinner,
   stopSpinner,
+  clearSpinners,
 } from "../tools/pretty"
-
-const isWindows = process.platform === "win32"
 
 export default {
   run: async (toolbox: GluegunToolbox) => {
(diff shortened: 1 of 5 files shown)


src/commands/new.ts
```diff
@@ -62,63 +61,68 @@ export default {
...
+      startSpinner("Igniting app")
+      remove(path(boilerplatePath, "ios", "Pods"))
+      remove(path(boilerplatePath, "node_modules"))
+      stopSpinner("Igniting app", "🔥")
+
+      startSpinner(" 3D-printing a new React Native app")
+      await copyBoilerplate(toolbox, {
+        boilerplatePath,
+        projectName,
+        excluded: ["node_modules", "yarn.lock", /.?\.expo\..?/],
```
markrickert, line 123: nice! Excluding all expo files with a regex :D

markrickert (approved):
Looks good. I like this method of making an ignite boilerplate project better than the old way anyways.

jamonholmgren:
Pushing up a fix for the test failure.

src/commands/new.ts
jamonholmgren: Took me a bit to remember if the wildcard should be `?.` or `.?`. haha

bryanstearns (approved):
Looks good! 🎉

PR #1700, opened 2021-06-04
Proper error message when folder already exists

jamonholmgren:
Currently, Ignite provides the wrong name ("Foo") on the overwrite error.

image

This fixes that.

yo dawg

### DIFF
--- a/src/commands/new.ts
+++ b/src/commands/new.ts
@@ -33,7 +33,7 @@ export default {
     if (parameters.options.overwrite) {
       remove(projectName)
     } else if (exists(projectName)) {
-      const alreadyExists = `Error: There's already a folder with the name ${projectName}. To force overwriting that folder, run with --overwrite.`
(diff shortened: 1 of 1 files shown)


derekgreenberg (approved):

src/commands/new.ts
```diff
@@ -33,12 +33,9 @@ export default {
     if (parameters.options.overwrite) {
       remove(projectName)
     } else if (exists(projectName)) {
+      const alreadyExists = `Error: There's already a folder with the name ${projectName}. To force overwriting that folder, run with --overwrite.`
```
kateinkim, line 36: Do we still want double quotes around the name or not?

kateinkim (approved):

src/commands/new.ts
GantMan: Agreed.  Do quotes.

PR #1717, opened 2021-07-07
Typescript: prop types added

Vercjames:

## Describe your PR
`StoryScreenProps` is an unused variable since it is not used in the file.
Typescript adjustment to include the aforementioned `StoryScreenProps` as the defined interface.
as a result, the eslint for `react/prop-types` on the props.children element is now satisfied.


jamonholmgren (approved):
Thank you @Vercjames , good catch!

Issue #1725, opened 2021-07-11
Error generating new project with Expo


phhoef:
What's going on?

When creating a new project with expo support, I'll get the following error:

```
➜  ~ npx ignite-cli new myProject --expo


   · · · · · · · · · · · · · · · · · · 🔥 Ignite 🔥 · · · · · · · · · · · · · · · · · ·

    █ Creating myProject using Ignite 7.0.6
    █ Powered by Infinite Red - https://infinite.red
    █ Using expo-cli
    ────────────────────────────────────────────────

    🔥 Igniting app
    🪔 Summoning Expo CLI
    🎫 Cleaning up Expo install
    🧶 Unboxing NPM dependencies
    ⠋  Writing your app name in the sand
/Users/ph/.npm/_npx/e31027f3785124a8/node_modules/gluegun/build/index.js:13
    throw up;
    ^
mv: no such file or directory: /Users/ph/myProject/ios/myProject
mv: no such file or directory: /Users/ph/myProject/ios/myProject-tvOS
mv: no such file or directory: /Users/ph/myProject/ios/myProject-tvOSTests
mv: no such file or directory: /Users/ph/myProject/ios/myProject.xcodeproj
mv: no such file or directory: /Users/ph/myProject/ios/myProject.xcodeproj/xcshareddata/xcschemes/myProject-tvOS.xcscheme
mv: no such file or directory: /Users/ph/myProject/ios/myProject.xcodeproj/xcshareddata/xcschemes/myProject.xcscheme
mv: no such file or directory: /Users/ph/myProject/ios/myProjectTests
mv: no such file or directory: /Users/ph/myProject/ios/myProjectTests/myProjectTests.m
mv: no such file or directory: /Users/ph/myProject/ios/myProject.xcworkspace
mv: no such file or directory: /Users/ph/myProject/ios/myProject/myProject.entitlements
mv: no such file or directory: /Users/ph/myProject/ios/myProject-Bridging-Header.h
node:internal/process/promises:246
          triggerUncaughtException(err, true /* fromPromise */);
          ^

[Error: ENOENT: no such file or directory, open '/Users/ph/myProject/android/app/src/main/AndroidManifest.xml'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/ph/myProject/android/app/src/main/AndroidManifest.xml'
}
```

Creating the project without expo (vanilla react native) is working.

---

Steps to reproduce

[...]

PR #1729, opened 2021-07-13
Fixes Expo error on ignite new -- fixes issue #1725

jamonholmgren:
In issue #1725, Expo builds were failing on `react-native-rename`, which isn't necessary since we are using `expo-cli` anyway.

This avoids calling the rename functionality when using Expo.

Additionally, we discovered that using some older versions of our CLI dependencies was causing issues, so we are locking major versions.

After implementing these changes, running `npx ignite-cli new MyApp --expo` with Node 16.4.2 succeeded.

cc @derekgreenberg

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ lerna-debug.log
 node_modules
 .vscode/*
 package-lock.json
+boilerplate/yarn.lock
 
 # TS build
 /build
--- a/src/commands/new.ts
+++ b/src/commands/new.ts
@@ -13,6 +13,13 @@ import {
   clearSpinners,
 } from "../tools/pretty"
 
+// CLI tool versions we support
+const cliDependencyVersions: { [k: string]: string } = {
+  expo: "4",
+  podInstall: "0.1",
+  rnRename: "2",
+}
+
 export default {
   run: async (toolbox: GluegunToolbox) => {
     const { print, filesystem, system, meta, parameters, strings } = toolbox
@@ -72,7 +79,7 @@ export default {
     p(` ────────────────────────────────────────────────\n`)
 
     if (expo) {
-      const expoCLIString = `npx expo-cli init ${projectName} --template ${boilerplatePath} --non-interactive`
+      const expoCLIString = `npx expo-cli@${cliDependencyVersions.expo} init ${projectName} --template ${boilerplatePath} --non-interactive`
       log({ expoCLIString })
 
       // generate the project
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -9,6 +9,7 @@ lerna-debug.log
 node_modules
 .vscode/*
 package-lock.json
+boilerplate/yarn.lock
```
jamonholmgren, line 12: The reason we added this to the CLI's .gitignore instead of the boilerplate's .gitignore is that we _only_ want to ignore it as a boilerplate, not when a new app is generated.

src/commands/new.ts
```diff
@@ -13,6 +13,13 @@ import {
   clearSpinners,
 } from "../tools/pretty"
 
+// CLI tool versions we support
+const cliDependencyVersions: { [k: string]: string } = {
```
robinheinze, line 17: Should we put this somewhere other than `new.ts`? Maybe in `/commands`? Seems like these could be used elsewhere in the CLI.

robinheinze (approved):

src/commands/new.ts
jamonholmgren: Good question. Perhaps in here: https://github.com/infinitered/ignite/tree/master/src/tools

jamonholmgren: I do tend to be partial to "co-locate until you can't" though, tbh.

wilmanh:
pretty @lindboe,  review pretty please! XD thanks everyone!

wilmanh:
hi @jamonholmgren any workaround to install create the new project without the error?

twerth:
> pretty @lindboe, review pretty please! XD thanks everyone!

@wilmanh it is unprofessional to comment on other people's appearance, please do not do that.

jamonholmgren:
Additionally, please review our [Code of Conduct](https://github.com/infinitered/ignite/blob/master/CODE_OF_CONDUCT.md).

wilmanh:
> > pretty @lindboe, review pretty please! XD thanks everyone!
> 
> @wilmanh it is unprofessional to comment on other people's appearance, please do not do that.

Hi! @twerth and @lindboe i am really sorry if i create any discomfort to anyone, english is not my first language, and i am still struggle with it, nevertheless, i was not referring to the appearance of @lindboe, i just forgot to put a word after pretty, 
the message was "pretty please" in both cases. in a way to try to beg (now i see that that was still an issue tho)

my regards

PR #1739, opened 2021-07-22
Update NAME-navigator.tsx.ejs

anhtuank7c:

## Describe your PR
- Current navigator template was obsolete since we're using react-navigation v5, this PR will use stack navigator from new dependency `@react-navigation/stack`


jamonholmgren (approved):
Thanks for this, @anhtuank7c . This is way better.

Issue #1738, opened 2021-07-21
When I make a `mst async action` in my first screen, something werid happend, async funciton not be called.


Arnoldnuo:
What's going on?

When I make a `mst async action` in my first screen, something werid happend, sometime async function not be called, sometimes I got this error `ExceptionsManager.js:180 Unhandled promise rejection RangeError: Maximum call stack size exceeded when consume api`.

---

Steps to reproduce

1. create a new project
2. remove all screens but the demo-list-screen, we need the first screen to make a async call.
3. use `flow(generator)` to make async call, for example

 ```tsx
.actions((self) => ({
    getCharacters: flow(function* (){
      const characterApi = new CharacterApi(self.environment.api)
      // yield delay(1);
      const result = yield characterApi.getCharacters()
      console.log('action end---');

      if (result.kind === "ok") {
        self.saveCharacters(result.characters)
      } else {
        __DEV__ && console.tron.log(result.kind)
      }
      return 1;
    }),
  }))
```

---

`npx ignite-cli doctor` results:

```
System
  platform           darwin
  arch               x64
  cpu                8 cores      Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
  directory          app          /Users/nuoli/Documents/projects/fapaifang/app

JavaScript (and globally-installed packages)
  node               14.17.3      /Users/nuoli/.nvm/versions/node/v14.17.3/bin/node
  npm                7.20.0       /Users/nuoli/.nvm/versions/node/v14.17.3/bin/npm
    @nestjs/cli      8.0.2
    ignite-cli       7.0.8
    npm              7.20.0
    react-devtools   4.14.0
    react-native     0.64.2
    react            17.0.1
    serverless       2.51.2
    tldr             3.3.7
    ts-node          10.1.0
    typeorm          0.2.34
    typescript       4.3.5
  yarn               -            not installed

Ignite
  ignite-cli         7.0.8        /Users/nuoli/.nvm/versions/node/v14.17.3/bin/ignite
  ignite src         build        /Users/nuoli/.nvm/versions/node/v14.17.3/lib/node_modules/ignite-cli/build

Android
  java               1.8.0_291    /usr/bin/java
  android home       -            /Users/nuoli/Library/Android/sdk

iOS
  xcode              12.5.1
  cocoapods          1.10.1       /usr/local/bin/pod
```

kateinkim:
[...]

PR #1740, opened 2021-07-22
Add transform for async to generator

kateinkim:
Some weird behavior was noticed when mixing async/await and flow actions: https://github.com/infinitered/ignite/issues/1738

Apparently including an async to generator babel transform fixes the issue (including a 1ms timeout also fixes the issue for some weird reason 🤷🏼‍♀️ ).


kateinkim:
This doesn't apply to ignite with expo.

PR #1759, opened 2021-08-07
adding gradient background component

JayPerfetto:

## Describe your PR
Adding a component which can be used as a LinearGradient Background wallpaper. I have replaced the use of the <Wallpaper> component in the demo screens, since it was not loading quickly enough in development builds and causing the UI to "flash" white when performing a navigation event. The component takes in an array of hex strings as props, so a user can customize the appearance of the gradient. 

It's worth noting that this component required adding the dependency `expo-linear-gradient` and raising the minimum iOS deployment target from 10.0 to 11.0 as a result. I tried to look into a lower level implementation of this, but all roads seemed to point to either this or the library "react-native-linear-gradient".

This is my first PR for this project, but I did my best to conform to existing code style + conventions! Feedback is welcome 😅


JayPerfetto:
I'm gonna need a pointer resolving this failed test - The failure is complaining about seemingly normal import syntax, and I don't want to go changing babel configs without discussion first

carlinisaacson:
Hi @JayPerfetto, thanks for the PR. I haven't looked through the changes yet, but I did pull it down and look at the failing tests. To fix them, you just need to add the expo-linear-gradient package to the `transformIgnorePatterns` in the boilerplate package.json. You can find documentation for that property in the Jest docs [here](https://jestjs.io/docs/configuration#transformignorepatterns-arraystring).

jamonholmgren:
This looks great! Just adding that Jest transformIgnorePatterns property and then we should be good to ship. Thanks @JayPerfetto !

carlinisaacson:
Had some time to scan through the code. Looks good!

JayPerfetto:
Great! I just checked in that change and it looks like everything is passing now - thanks for the quick pointer there

jamonholmgren (approved):

jamonholmgren:
@JayPerfetto This is great, thanks. Now you're forever one of us. 💜

Issue #1765, opened 2021-08-11
Android files have wrong package name: package com.helloworld;


derekgreenberg:
Ignite a new project with a name such as "funapp" and observe that generated Android java files include this: `package com.helloworld;`

<img width="377" alt="Screen Shot 2021-08-11 at 2 24 45 PM" src="https://user-images.githubusercontent.com/144869/129105482-5d19875e-bda0-4a3e-975d-2a9ef27e911e.png">

PR #1767, opened 2021-08-13
Ensures that bundle ID is set and customizable

jamonholmgren:
In #1765, the issue is that on Android (vanilla apps), the bundle identifier is not being set.

This PR fixes that (thanks @carlinisaacson) and also provides the ability to pass in `--bundle` to your `ignite-cli new` command to customize that.

Bundle ID is set screenshot


carlinisaacson (approved):
Thanks for the credit, boss! LGTM

PR #1787, opened 2021-09-02
Add handle exception screen

vantuan88291:

## Describe your PR
As we know, when the app has a JS exception, it will throw White screen instead Red screen in release mode. So i added error screen let user can reload app when some exception was happened. Its look like bellow:
Simulator Screen Shot - iPhone SE (2nd generation) - 2021-09-02 at 15 55 53


boilerplate/app/app.tsx
```diff
@@ -59,10 +60,12 @@ function App() {
...
-          <AppNavigator
-            initialState={initialNavigationState}
-            onStateChange={onNavigationStateChange}
-          />
+          <ErrorBoundary catchErrors={"always"}>
+            <AppNavigator
+              initialState={initialNavigationState}
+              onStateChange={onNavigationStateChange}
+            />
+          </ErrorBoundary>
```
lindboe, line 68: Suggestion, not required for this PR: I find that it's generally good practice to also wrap screens within a navigator with an error boundary, so that navigation can continue to work if a component only nested within a single screen fails. I think it's helpful to be able to recommend that in a best-practices template.

However, this is mostly useful in apps that have navigation menus/bars outside the screen components (making sure those continue to work), so maybe it won't be super clear why that's helpful in this particular layout. Curious for your thoughts @jamonholmgren. I could see an argument for a separate PR that introduces an example navigation menu and then wraps individual screens.

jamonholmgren: Good idea! I like this.

jamonholmgren (approved):

boilerplate/app/app.tsx
jamonholmgren: @lindboe I think we should get a fresh new design for Ignite's boilerplate and add either a drawer or tab bar, after which we could incorporate your suggestion.

PR #1831, opened 2021-11-17
Adding React Native Colo Loco to Ignite

jamonholmgren:
This adds [React Native Colo Loco](https://github.com/jamonholmgren/react-native-colo-loco) support to Ignite out of the box.

It also enables Swift and Kotlin as well, out of the box.

You can skip this by passing in `--colo-loco=false`. It also won't work with `--expo` (yet).

Just need reviews & it should be good to go!


leonskim (approved):
Looks good to me! I was wondering if anyone prefers the OG way developing native modules in the native directories only, but I thought it again, it won't matter because it will work both way. Nice work!

carlinisaacson (approved):
> Looks good to me! I was wondering if anyone prefers the OG way developing native modules in the native directories only, but I thought it again, it won't matter because it will work even if coco-loco is enabled. Nice work!

I wondered this too. More specifically how this might affect newcomers learning RN. Learning the OG way is a good skill for an RN developer to have. But this doesn't force anybody to use it, and it could be nice for those that are tired of writing native module boilerplate.

jamonholmgren:
> Looks good to me! I was wondering if anyone prefers the OG way developing native modules in the native directories only, but I thought it again, it won't matter because it will work even if coco-loco is enabled. Nice work!

Yeah, exactly -- this will _add_ colocated native modules, but you can still add your own the old way. In fact, this makes it easier because you don't have to make your own Package file in Android, we include it for you.

PR #1833, opened 2021-11-21
Add  bullet-item component story

oguzturker8:
This PR adds missing story for bullet-item component.


jamonholmgren (approved):
Good catch!

Issue #1807, opened 2021-10-14
not seeing the "Toggle Storybook" option in the developer menu


chancesmith:
What's going on?
I'm not seeing the "Toggle Storybook" option in the developer menu.

![image](https://user-images.githubusercontent.com/4466585/137400023-a385349d-923a-450a-8dd2-dfc4edfc7870.png)

---

Steps to reproduce

1. `npx ignite-cli new PizzaApp --expo --bundle=com.infinitered.pizzaapp`
2. `yarn storybook`
3. `yarn start`
4. open developer menu (CMD + D)
5. expect to see the option "Toggle Storybook"

---

`npx ignite-cli doctor` results:
```
*[master][~/wip/pizzaapp]$ npx ignite-cli doctor
System
  platform           darwin                                                  
  arch               x64                                                     
  cpu                16 cores      Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz 
  directory          pizzaapp   /Users/chance/wip/pizzaapp             

JavaScript (and globally-installed packages)
  node                  14.18.1      /Users/chance/.nvm/versions/node/v14.18.1/bin/node 
  npm                   6.14.15      /Users/chance/.nvm/versions/node/v14.18.1/bin/npm  
    npm                 6.14.15                                                         
  yarn                  1.22.10      /usr/local/bin/yarn                                
    create-react-app    4.0.1                                                           
    ios-deploy          1.11.1                                                          
    npm-check-updates   7.0.2                                                           
    serve               11.3.2                                                          

Ignite
  ignite-cli         7.5.0        /Users/chance/.npm/_npx/822/bin/ignite                        
  ignite src         build        /Users/chance/.npm/_npx/822/lib/node_modules/ignite-cli/build 

Android
  java               13.0.2       /usr/bin/java 
  android home       -            undefined     

iOS
  xcode              12.4       
  cocoapods          1.10.1       /usr/local/bin/pod 
```

chancesmith:
@jamonholmgren any tip or direction to get storybook showing? I'm happy to dig in.

bayswaterpc:
To add to this I'm getting a slightly different issue where storybook is showing a constant "loading story" on the storybook page.

![StorybookEternalLoading](https://user-images.githubusercontent.com/78702667/140648667-d3df5698-4237-4d30-9d9b-f9a25a397bf6.png)

-  `yarn android`
-  `yarn storybook`

I'm using windows 10 and node 16.

[...]

PR #1839, opened 2021-11-23
Fix Storybook not triggering from Reactotron

leonskim:

- [✅] `yarn ci:test` **jest** tests pass with new tests, if relevant
- [✅] `README.md` has been updated with your changes, if relevant

## Describe your PR
- Fixes Storybook on/off toggle on Reactotron not switching the Ignite app and Storybook.
- `Reactotron.storybookSwitcher()` makes the app to toggle between app and Storybook but it requires the app to be wrapped in a HOC. However, there are cases that wouldn't work the best.
- This fix directly connects and listens commands from Reactotron, and catch the Storybook toggle commands.
- There's another possible way: If `reactotron-react-native` provides a listener wrapped Mitt instance as a callback, the app could possibly utilize by planting a listener for this type of events. (cc: @rmevans9)
- For the https://github.com/infinitered/ignite/issues/1807, it is impossible to customize the dev menu on managed Expo app, so like @jamonholmgren mentioned, using Reactotron to toggle would be the most convenient.

### Bare React-Native
storybook-bare

### Expo
storybook-expo


jamonholmgren (approved):
Looks great.

PR #1846, opened 2021-12-10
Fixes new project init when no yarn is installed

jamonholmgren:
When not using yarn, Ignite is supposed to fall back to NPM. However, with recent versions of NPM, it was failing with an error about peer dependencies.

This PR upgrades some dependencies to remove those errors and also removes the package `react-powerplug` which is not maintained anymore.


jamonholmgren:
Going to fold this PR into #1847 and do it all at once.

PR #1847, opened 2021-12-10
Support pnpm in addition to yarn

jamonholmgren:
This is a PR for supporting pnpm in addition to yarn/npm.

pnpm is a much faster alternative to yarn and npm, but uses symlinks. So this PR will add support for symlinks to Ignite, which has the added benefit of working better with monorepos!

How this works: if Ignite detects `pnpm`, it will use that packager first. If it doesn't, it'll then check for `yarn`. Lastly, it'll fall back to `npm` if nothing else is available.

Note that if you're using Expo, it doesn't support `pnpm`, so it'll fall back to yarn or npm.

You can also pass in a packager by using `--packager=<packager>` -- options are `pnpm`, `yarn`, or `npm`.

Other changes:

* react-powerplug is replaced with a local version as it is no longer maintained
* Added metro.config.js
* Added expo-modules-core, @react-native-community/cli-platform-ios, @rnx-kit/metro-config, and @rnx-kit/metro-resolver-symlinks to support symlinks
* Updated to latest version of Detox


boilerplate/package.json
```diff
@@ -78,36 +80,26 @@
     "eslint-plugin-promise": "4.2.1",
     "eslint-plugin-react": "7.21.5",
     "eslint-plugin-react-native": "3.10.0",
+    "expo-modules-autolinking": "^0.3.4",
```
leonskim, line 83: Just curious: AFAIK  `expo` script does autolinking, what does this dependency do?

jamonholmgren: It's a dependency of `expo`, but for some reason I had to make certain packages explicit or pnpm wouldn't set them up properly. Weird, I nkow.

leonskim (approved):
LGTM! I'd like to see some feedbacks from monorepo projects.

carlinisaacson (approved):

boilerplate/metro.config.js
```diff
@@ -0,0 +1,29 @@
...
+    resolver: {
+      resolveRequest: MetroSymlinksResolver(),
+    },
+  })
+
+  return {
+    resolver: {
+      assetExts: [...defaultConfig.resolver.assetExts, "bin"],
+    },
+    ...symlinkSupport,
```
tido64, line 27: Wouldn't `resolver` be overwritten by spreading `symlinkSupport` here? Any reason it couldn't be moved to the `resolver` section above, together with `resolveRequest`?

jamonholmgren: You're right @tido64. I'll fix that.

Issue #1800, opened 2021-10-05
Issue in starting new app


aawaheed:
What's going on?
It was talking a lot of time on unboxing NPM dependencies so I quit. Found a way to enable `--debug` for CLI. Turns out that it was failed as there is no github account configured globally.

![ignite-error](https://user-images.githubusercontent.com/6452873/135960211-fce53ae4-25dc-4c86-a37c-d06a4e2b6e6a.png)

---

Steps to reproduce

1. run `npx ignite-cli new pizzapp --debug`

---

`npx ignite-cli doctor` results

![ignite-doctor](https://user-images.githubusercontent.com/6452873/135960161-91ea503f-b738-4b86-af8d-b323fcf4851f.png)

---
However the application seems to run if I go ahead and do 
`npx react-native run-android`

aawaheed:
I tried `npx ignite-cli new RN1 --git=false --debug ` but that also resulted the same issue.

PS: I made sure this time there is no git config set :)

PR #1853, opened 2021-12-28
Continue "new" process when there's no git config

leonskim:

- [☑️] `yarn ci:test` **jest** tests pass with new tests, if relevant
- [☑️] `README.md` has been updated with your changes, if relevant

## Describe your PR
- Ref: https://github.com/infinitered/ignite/issues/1800
- When a user creates a new Ignite app without `git config` setup in the global environment(`user.name` and `user.email` to be precise) it fails on committing the initial changes after setting up the app.
- I wasn't able to recreate it when deleting `user.name` and `user.email` even deleting the items Keychain Access, but then I was when setting the items with empty(`""`) to trigger the error.
- This workaround fixes it and let the `ignite new` process complete - And let the user commit it after setting up the git config.
- Note: Didn't update the tests for when git config isn't setup, since there's not much value testing/developing without it. 

lil'info

### DIFF
--- a/src/commands/new.ts
+++ b/src/commands/new.ts
@@ -276,14 +276,18 @@ export default {
     // commit any changes
     if (parameters.options.git !== false) {
       startSpinner(" Backing everything up in source control")
-      await system.run(
-        log(`
-          \\rm -rf ./.git
-          git init;
-          git add -A;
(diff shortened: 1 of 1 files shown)


jamonholmgren (approved):
Nice ... I like it.

PR #1855, opened 2021-12-29
fix: update component name in navigator template

### LINKED ISSUE #1854: Navigator Naming Convention
What's going on?

When generating a navigator (e.g. "test") the generated file will be named test-navigator.tsx.
However, the generated component's name is Test and not TestNavigator, which would be consistent with the AppNavigator component. 

 I wonder why this is the default case and if there is a reason why generated, custom navigators should not be called navigators.

---

Steps to reproduce

npx ignite-cli generate navigator test

simon-lund:

## Describe your PR

Append "Navigator" to name of navigator component in navigator template. resolves #1854

**Note**: I tried to run tests (used both  `yarn test` and `yarn ci:test`) right after cloning the repo but it got stuck at *test/ignite-new.test.ts*, so this PR is not tested :cry:

### DIFF
--- a/boilerplate/ignite/templates/navigator/NAME-navigator.tsx.ejs
+++ b/boilerplate/ignite/templates/navigator/NAME-navigator.tsx.ejs
@@ -9,7 +9,7 @@ export type <%= props.pascalCaseName %>ParamList = {
 }
 
 const Stack = createStackNavigator<<%= props.pascalCaseName %>ParamList>()
-export const <%= props.pascalCaseName %> = () => {
+export const <%= props.pascalCaseName %>Navigator = () => {
   return (
     <Stack.Navigator screenOptions={{ cardStyle: { backgroundColor: "transparent" }, headerShown: false, }}>
       <Stack.Screen name="demo" component={DemoScreen} />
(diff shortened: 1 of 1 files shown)


boilerplate/ignite/templates/navigator/NAME-navigator.tsx.ejs
```diff
@@ -9,7 +9,7 @@ export type <%= props.pascalCaseName %>ParamList = {
 }
 
 const Stack = createStackNavigator<<%= props.pascalCaseName %>ParamList>()
```
jamonholmgren, line 11: Let's change the ParamList to *NavigatorParamList too.

```suggestion
const Stack = createStackNavigator<<%= props.pascalCaseName %>NavigatorParamList>()
```

jamonholmgren: (And on line 7 as well...)

jamonholmgren (changes requested):
Just a couple small changes!

boilerplate/ignite/templates/navigator/NAME-navigator.tsx.ejs
simon-lund: good point

jamonholmgren (approved):

Issue #1826, opened 2021-11-10
ToggleStorybook vs react-native-camera


maqeelqureshi:
react-native-camera library doesn't work inside ToggleStorybook tag.

I am trying to use camera library. Camera component works fine in simple RN project but when I try to use the same in ignite project, it shows black screen. 

Only if I remove the ToggleStorybook parent element in app.tsx than it works fine.

Any suggestion?

jamonholmgren:
We probably need to move the `getStorybookUI` function call into a function and only call it when storybook is activated.

https://github.com/infinitered/ignite/blob/ef5810aac792794ee9102e4e563862145d098ea0/boilerplate/storybook/storybook.tsx#L11-L16

According to @dannyhw from React Native Storybook, it polyfills some things which can be problematic and possibly causing this issue. (Potentially resolved in Storybook 6.x)

dannyhw:
somewhat related issue: https://github.com/storybookjs/react-native/issues/20

dannyhw:
@maqeelqureshi it would be worth checking this on the latest version of ignite since the change in #1906 would probably fix this issue.

Issue #1769, opened 2021-08-16
first api call in the first screen has no result callback


Magrider:
What's going on?
First api call in the first screen has no result callback 
---

Steps to reproduce

1. switch the stack screen order in 'main-navigator.tsx' to be 
   ` <Stack.Screen name="demoList" component={DemoListScreen} />
      <Stack.Screen name="welcome" component={WelcomeScreen} />`

then start the program, you will find out it will never get the result 
`      const result = await characterApi.getCharacters()
` in character-store.ts file

It only happen in the initial screen.

---

`npx ignite-cli doctor` results:
System
  platform           darwin
  arch               x64
  cpu                12 cores       Intel(R) Core(TM) i7-8700 CPU @ 3.20GHz
  directory          canbeDeleted   /Users/maci/Documents/git/canbeDeleted

JavaScript (and globally-installed packages)
  node               16.3.0       /usr/local/bin/node
  npm                7.17.0       /usr/local/bin/npm
    npm              7.17.0
    yarn             1.22.10
  yarn               1.22.10      /usr/local/bin/yarn
    code-push-cli    2.1.9
    ignite-cli       7.0.3

Ignite
  ignite-cli         7.0.3        /usr/local/bin/ignite
  ignite src         build        /Users/maci/.config/yarn/global/node_modules/ignite-cli/build

Android
  java               1.8.0_261    /usr/bin/java
  android home       -            /Users/maci/Library/Android/sdk

iOS
  xcode              12.5.1
  cocoapods          1.10.1       /usr/local/bin/pod

Magrider:
I have to add a delay to solve it. I have no idea why.

kateinkim:
Hi @Magrider! We have noticed this issue in other cases and what you did by adding a delay does solve the issue. This is a workaround that we are currently recommending. We are trying to figure out why this is happening and we would like to solve it as quickly as possible. Thanks 😃

m9hmood:
I had the same issue, 
I called the store action in ``useEffect`` but it wasn't working
after adding a delay to the actions it worked fine
but after checking the ``useEffect`` I found that I had an infinite loop
so, after I fixed it everything works normally now

momolafrooo:
Me too i got the same issue, add i fixed it by adding a delay of 1ms. i know that's not the best solution, but it's the only one i found. Hope this bug will be fixed soon.

abhisekz:
[...]

PR #1863, opened 2022-01-15
Use ES6 imports in StorybookToggle and remove component from state

### LINKED ISSUE #1860: App stops executing next lines of code after network requests
## What's going on?

I'm waiting for network requests to finish before change the state of the app using mobx actions, I'm using generator functions as described in [here](https://mobx-state-tree.js.org/concepts/async-actions)

Here's a screenshot of logs from Reactotron, there should be a log after the API response that shows the state change
reactotron

### Action:
```typescript
.actions((self) => ({
  // ...
  signIn: flow(function* (form: CustomerSigninForm) {
    const api = self.environment.apis.guest
    const signinResult = yield api.signinAsCustomer(form) // <= code stops executing here

    // state change (saving token, fetching user details, etc

  }),
  // ...
}))
```

### API method:
```typescript
export class GuestApi extends Api {
  // client config, other methods
  async signinAsCustomer(form: CustomerSigninForm): Promise<SigninCustomerResult> {
    const url = this.base + "/integration/customer/token"
    const params = mapToCustomerSigninRequest(form)
    const response = await this.apisauce.post<string>(url, params) // <-- Code lines after this line don't execute
    if (!response.ok) {
      return getGeneralApiProblem(response) || { kind: "bad-data" }
    }
    return { kind: "ok", token: response.data }
  }
}
```

I'm experiencing this issue with both Android and iOS, however on Android, if I turn off __DEV__ from `cmd + M > Settings` and reload the app it works perfectly fine

I tried running it with minification enabled and the issue still persists, am I doing something wrong?

---

## Steps to reproduce

Will create a reproduction repo if no one has faced this issue before

---

## `npx ignite-cli doctor` results:

doctor

carlosa54:

## Describe your PR
Refactor `storybook-toggle.ts` to use ES6 imports when importing `StorybookUIRoot` and remove component from state this is causing issues across the app regarding networking and random blank pages, in my case I was getting a blank page on Android and I fixed it with this

Here's a list of issues related:
Closes #1860, #1826?, #1769

### DIFF
--- a/boilerplate/storybook/toggle-storybook.tsx
+++ b/boilerplate/storybook/toggle-storybook.tsx
@@ -2,6 +2,7 @@ import React, { useState, useEffect, useRef } from "react"
 import { DevSettings } from "react-native"
 import { loadString, saveString } from "../app/utils/storage"
 import { DEFAULT_REACTOTRON_WS_URI } from "../app/services/reactotron/reactotron-config"
+import { StorybookUIRoot } from "./storybook"
 
 /**
  * Toggle Storybook mode, in __DEV__ mode only.
@@ -15,7 +16,6 @@ import { DEFAULT_REACTOTRON_WS_URI } from "../app/services/reactotron/reactotron
  */
 export function ToggleStorybook(props) {
   const [showStorybook, setShowStorybook] = useState(false)
-  const [StorybookUIRoot, setStorybookUIRoot] = useState(null)
   const ws = useRef(new WebSocket(DEFAULT_REACTOTRON_WS_URI))
 
   useEffect(() => {
@@ -44,9 +44,6 @@ export function ToggleStorybook(props) {
         })
       }
 
-      // Load the storybook UI once
-      setStorybookUIRoot(() => require("./storybook").StorybookUIRoot)
-
       // Behave as Reactotron.storybookSwitcher(), not a HOC way.
       ws.current.onmessage = (e) => {
         const data = JSON.parse(e.data)
@@ -64,7 +61,7 @@ export function ToggleStorybook(props) {
   }, [])
 
   if (showStorybook) {
-    return StorybookUIRoot ? <StorybookUIRoot /> : null
+    return <StorybookUIRoot />
   } else {
     return props.children
   }


dannyhw:
@carlosa54 I think that this code isn't exactly equivalent, the inline require is used to avoid loading storybook when it's not being used. You could argue that by setting inlineRequires in metro to be true then it might be equivalent however I'm not sure about that.

I think you could fix those bugs by not loading storybook on line 47:49 and instead calling the require just before it gets rendered on line 67. 

An additional note is that the problem with the Promises is resolved completely in the v6 alpha of react native storybook by using the modern version of storybook. This is done by adding sbmodern in resolverMainFields.

jamonholmgren:
Thanks @dannyhw!

@carlosa54, would you be able to update your PR with those changes?

carlosa54:
Thanks @dannyhw for the context.

@jamonholmgren sure, I'll add the changes to account for the inline require when I get a chance

carlosa54:
I implemented the suggested code and started to get the blank page again. I also noticed that the promises started hanging E.g. `loadString("devStorybook")` don't resolve. Not sure why I can't replicate this when using ES6 imports might need to dig deeper into this when I have more time.

dannyhw:
@jamonholmgren @carlosa54 How do you feel about using patch package on these types of things?

In this [PR](https://github.com/HurrellT/formikStorybookTest/pull/1) I demonstrate how you can use patch package to fix the promises issue.

Basically just this patch for @storybook/addons which remove the corejs promise polyfill (the cause of the bug)

```
diff --git a/node_modules/@storybook/addons/dist/index.js b/node_modules/@storybook/addons/dist/index.js
index 589d611..81bc9e7 100644
--- a/node_modules/@storybook/addons/dist/index.js
+++ b/node_modules/@storybook/addons/dist/index.js
@@ -8,8 +8,6 @@ require("core-js/modules/es.object.to-string");
 
 require("core-js/modules/es.object.values");
 
-require("core-js/modules/es.promise");
-
 require("core-js/modules/web.dom-collections.for-each");
 
 Object.defineProperty(exports, "__esModule", {
 
```

jamonholmgren:
@dannyhw I'm all for patch-package here! Ignite is all about sanding off these rough edges that you run into, and if patch-package can get us there, that would be fantastic.

dannyhw:
@carlosa54 I hope you don't mind but I created a new PR #1906 with a fix

In the PR I include the patch as mentioned in my previous comment. I think that should solve the problem.

carlosa54:
Thanks @dannyhw @jamonholmgren

I can confirm that when applying the patch from https://github.com/infinitered/ignite/pull/1906 all the promises resolves and I'm not getting the blank page I used to get.

jamonholmgren:
Fixed in 7.10.6 -- closing -- thank you @carlosa54!

Issue #1857, opened 2021-12-30
TS Issue on generating new screens


lvlrSajjad:
What's going on?

When I generate a new screen using command below, it generates screen without NavigatorParamList more info in following paragraphs.

This is the demo screen:
![image](https://user-images.githubusercontent.com/42815758/147771254-440a3acf-76b5-468c-9f9c-4b00361708cb.png)

This is my generated screen:
![image](https://user-images.githubusercontent.com/42815758/147771306-a10f2038-aee9-4fea-aea2-532798e49d7c.png)

This is the TS issue:
![image](https://user-images.githubusercontent.com/42815758/147771331-f9f89e6b-9a90-426d-88ed-9898c5fad413.png)

Although it is pretty easy to fix but it'd be less confusing if we generated screens with NavigatorParamList ...

Also I had to add this to fix all TS issues:
![image](https://user-images.githubusercontent.com/42815758/147771692-1d9872ab-ea5d-4a3b-82e1-72c8b74bfb3e.png)

---

Steps to reproduce

1. 
```  npx ignite-cli g navigator Splash ```

---

`npx ignite-cli doctor` results:
``` Not needed for this one :-D ```

robinheinze:
Hey @lvlrSajjad, taking a look at this!

PR #1874, opened 2022-01-27
Add better typings and instructions to screen generator

robinheinze:

 DOES NOT PASS, looks like a `colo-loco` error? 
- [N/A] `README.md` has been updated with your changes, if relevant

## Describe your PR

- Addresses #1857 by adding in the requested types, along with some instructions for adding the screen to navigation config. 

Here's what the user sees in the generated screen: 
Screen Shot 2022-01-27 at 3 56 30 PM

And here's what they see in `app-navigator.tsx` before adding their own config (arrows show what was changed):
Screen Shot 2022-01-27 at 3 51 37 PM

- I'm thinking about the value of establishing a system where users can look for 🔥 in the code as a sign that they may need to add their own customization various places? This is an idea I stole from Kent C. Dodds courses.

### DIFF
--- a/boilerplate/app/navigators/app-navigator.tsx
+++ b/boilerplate/app/navigators/app-navigator.tsx
@@ -27,6 +27,7 @@ export type NavigatorParamList = {
   welcome: undefined
   demo: undefined
   demoList: undefined
+  // 🔥 Your screens go here
 }
 
 // Documentation: https://reactnavigation.org/docs/stack-navigator/
@@ -43,6 +44,7 @@ const AppStack = () => {
       <Stack.Screen name="welcome" component={WelcomeScreen} />
       <Stack.Screen name="demo" component={DemoScreen} />
       <Stack.Screen name="demoList" component={DemoListScreen} />
+      {/** 🔥 Your screens go here */}
     </Stack.Navigator>
   )
 }
--- a/boilerplate/ignite/templates/screen/NAME-screen.tsx.ejs
+++ b/boilerplate/ignite/templates/screen/NAME-screen.tsx.ejs
@@ -1,6 +1,8 @@
-import React from "react"
+import React, { FC } from "react"
 import { observer } from "mobx-react-lite"
 import { ViewStyle } from "react-native"
+import { StackScreenProps } from "@react-navigation/stack"
+import { NavigatorParamList } from "../../navigators"
 import { Screen, Text } from "../../components"
 // import { useNavigation } from "@react-navigation/native"
 // import { useStores } from "../../models"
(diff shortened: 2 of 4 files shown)


robinheinze:
Troubleshooting the tests now...

boilerplate/app/navigators/app-navigator.tsx
```diff
@@ -27,6 +27,7 @@ export type NavigatorParamList = {
   welcome: undefined
   demo: undefined
   demoList: undefined
+  // 🔥 Your screens go here
```
lindboe, line 30: _Love_ the fire emoji idea, that's brilliant

jamonholmgren:
I restarted the tests, which appear to have timed out.

boilerplate/app/navigators/app-navigator.tsx
kateinkim: Agreed~

lindboe: Thinking about it more, I wonder if we should use different emoji for different generators? "Look for the 🔥 " is less useful if we reuse it a lot. (This might be the plan, just wasn't sure.)

kateinkim (approved):

boilerplate/app/navigators/app-navigator.tsx
robinheinze: I hadn't put all that much thought into it beyond establishing it as an easy way to direct the users eyes to the lines they need to change. I think there's definitely room to iterate on it in the future!

robinheinze:
Ok, so the tests are failing because of the TS error, which was knowingly introduced with this change (because we are requiring the user to add the correct typings rather than letting the generator modify the navigator files). 

Couple of different options here:
- Put in an `any` after the screen key, which is erroring
- Add a `@ts-ignore` line and instruct the user to remove it?
- ???

Any thoughts @jamonholmgren @lindboe @kateinkim ?

lindboe:
Thinking about the TS failure:

This seems maybe more like a paradigm shift that should be reflected in the tests instead of being ignored here. It seems like we're changing the expectation to "Now, after you generate a new file, don't expect the project to immediately pass type-checking." We expect people to immediately fix it, so ts-ignore doesn't feel quite right to me.

But we do still want it to pass after generating a new project.

What do you think of changing the tests so that we run typechecking and other lint rules after making the new project, but only check for non-type lint errors after running generate commands?

robinheinze:
I agree that ignore does feel *wrong* somehow. But also, in general, I think we do want the generated screens/components/etc. to not have any avoidable TS errors, so how do we guard against developer oversight when making changes, but allow for *intentional* ones like this? I haven't been able to come up with anything better yet.

lindboe:
> so how do we guard against developer oversight when making changes, but allow for intentional ones like this

That makes sense -- I was thinking they'd show up during development, but that might not work with template files?

lindboe (approved):

robinheinze:
Exactly. If there is a TS error in a generated file, it wouldn't be obvious from the template. So in case someone made a change without actually running the generator (which... they shouldn't but... humans :D) we want CI to still catch that.

PR #1886, opened 2022-02-11
using `baseUrl` to avoid `../../../`

### LINKED ISSUE #1770: Use baseUrl in tsconfig for absolute paths
We should discuss/consider using `baseUrl` in our tsconfig so we can change imports like this:

```tsx
import { Thing } from "../../components/thing"
```

to this

```tsx
import { Thing } from "components/thing"
```

Docs: https://www.typescriptlang.org/tsconfig#baseUrl

flexbox:

## Describe your PR

It the first attempt to closes #1770 

I installed the tooling with https://github.com/infinitered/ignite/pull/1886/commits/2da5b74a07812c1e941cd4a4384110a676d39102

@majidln
Before migrating the entire `/boilerplate` folder I need some inputs if you think it's in a good direction
https://github.com/infinitered/ignite/pull/1886/commits/d4f364bee4eab2a90d1c11335160cf610c25972d

Let me know in the comments 😉 

---

Nex Step:

- [ ] Fix model errors to avoid

```
 FAIL  app/models/character-store/character-store.test.ts
  ● Test suite failed to run

    [mobx-state-tree] expected mobx-state-tree type as argument 1, got undefined instead

       7 | // prettier-ignore
       8 | export const RootStoreModel = types.model("RootStore").props({
    >  9 |   characterStore: types.optional(CharacterStoreModel, {} as any),
         |                         ^
```

### DIFF
--- a/boilerplate/README.md
+++ b/boilerplate/README.md
@@ -65,7 +65,6 @@ ignite-project
 │   └── IgniteProjectTests
 ├── .env
 └── package.json
-
 ```
 
 ### ./app directory
@@ -113,15 +112,15 @@ This is a great place to put miscellaneous helpers and utilities. Things like da
 
 **app.tsx** This is the entry point to your app. This is where you will find the main App component which renders the rest of the application.
 
-### ./ignite directory
+### `./ignite` directory
 
 The `ignite` directory stores all things Ignite, including CLI and boilerplate items. Here you will find generators, plugins and examples to help you get started with React Native.
 
-### ./storybook directory
+### `./storybook` directory
 
 This is where your stories will be registered and where the Storybook configs will live.
 
-### ./test directory
+### `./test` directory
 
 This directory will hold your Jest configs and mocks, as well as your [storyshots](https://github.com/storybooks/storybook/tree/master/addons/storyshots) test file. This is a file that contains the snapshots of all your component storybooks.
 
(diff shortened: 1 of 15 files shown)


jamonholmgren:
@flexbox On-stream I trialled another option. Here's a link to the commit (not fully done, but has the gist of it). So, this works, pretty well actually, but the only slight disadvantage is that storybook would be `~/../storybook`. I'm okay with that, and you can watch today's stream (first section) to see why. But interested in getting your thoughts, too.

https://github.com/jamonholmgren/FastingApp/commit/ff1c35d6eaef425e880dc9bb77ee442a6adf9163

flexbox:
Just tried the `tsconfig.json` strategy only on another project.
https://github.com/flexbox/react-native-bootcamp/pull/77/commits/1d0df58a531583558ee82035fe61328053bd374b

It works fine, and avoid having a dependency with `babel-plugin-root-import`.
This is a better solution to reduce errors due to the plumbing 😅 

Next step:
I will investigate the storybook issue on `ignite-cli`

flexbox:
Just noticed with the `package.json` approach, during expo build, we are running into an error

image


Next step:

- [ ] try to build the project
- [ ] investigate the storybook issue on ignite-cli

flexbox:
Closing that for now because

I don't have the bandwidth right now
let the team focus on https://github.com/infinitered/ignite/milestone/19

PR #1895, opened 2022-02-16
Faster CI

jamonholmgren:
This branch implements some experiments to speed up CI.

* Remove --runInBand so tests run in parallel
* Split Expo and vanilla tests on CircleCI (cuts off 3 minutes)
* Run in a ramdisk so disk writes are faster (not a ton faster TBH)


jamonholmgren:
Latest update runs Expo and "vanilla" tests in parallel on CircleCI, which is nice.

image

PR #1900, opened 2022-02-22
Auto scroll screen preset

martincpt:

## Describe your PR
Giving ability to automatically set screen scroll behaviour depending on content size. With this PR, there is an additional screen preset named 'auto' which behaves as fixed screen if content fits the screen and changes itself to scrollable if it does not.


boilerplate/app/components/screen/screen.tsx
```diff
@@ -32,6 +32,16 @@ function ScreenWithScrolling(props: ScreenProps) {
...
   const insetStyle = { paddingTop: props.unsafe ? 0 : insets.top }
 
+  // The followings for <Screen preset='auto'/>
+  // This will automatically disables scrolling if content fits the screen.
+  const { height } = Dimensions.get('window');
+  const [screenHeight, setScreenHeight] = React.useState(0);
+  const scrollEnabled = props.preset === 'scroll' || props.preset === 'auto' && screenHeight > height * presets.auto.offset.percent - presets.auto.offset.point;
+
+  const onContentSizeChange = (width, height) => {
+    setScreenHeight(height);
```
jamonholmgren, line 42: Will this cause a double-render?

In other words, does `onContentSizeChange` come with a re-render anyway? If so, we could just hold the screenheight in a useRef and calculate when the re-render happens.

It's possible this simply gets batched with the onContentSizeChange render too. Would need to test this, I guess!

jamonholmgren:
I like this idea, @martincpt -- just have a question above.

boilerplate/app/components/screen/screen.tsx
martincpt: Thank you for your response!

Yeah, you are right about double-rendering.

I've made some changes and pushed a new commit. Can you check?

With this fix, it's not rendering twice by default, only if preset '**auto**' is used and content fits the screen, so it's necessary.

There is one setback though, metro's soft reload is not updating **scrollEnabled** state properly when constantly changing between preset '**auto**' and '**scroll**' because **onContentSizeChange** not called either on a file change (at least if content stays the same). One may need reload screen manually to catch the updated state.

It's something I can live with. To fix this issue, it needs significantly more code which probably makes code readability worse.

martincpt: > There is one setback though, metro's soft reload is not updating scrollEnabled state properly when constantly changing between preset 'auto' and 'scroll' because onContentSizeChange not called either on a file change (at least if content stays the same). One may need reload screen manually to catch the updated state.

I pushed one more commit to fix this and works well but it's more code as I said.

martincpt:
Thank you @jamonholmgren, I've pushed updates. Please review.

boilerplate/app/components/screen/screen.tsx
```diff
@@ -32,6 +32,37 @@ function ScreenWithScrolling(props: ScreenProps) {
...
+  const updateScrollState = () => {
+    if (props.preset === 'auto'){
+      // check whether if content fits the screen
+      const contentFitsScreen = screenHeight.current > height * presets.auto.offset.percent - presets.auto.offset.point
+        
+      // then toggle scroll state according to it, make sure it's not rendering twice
+      if (scrollEnabled && !contentFitsScreen) setScrollEnabled(false)
+      else if (!scrollEnabled && contentFitsScreen) setScrollEnabled(true)
+    }
+    else if (!scrollEnabled) setScrollEnabled(true) // set back initial state in case it's locked
```
jamonholmgren, line 50: Nit: I prefer either one-liners or using braces, rather than 2-liners.

```suggestion
    } else if (!scrollEnabled) {
      // set back initial state in case it's locked
      setScrollEnabled(true)
    }
```

boilerplate/app/components/screen/screen.tsx
```diff
@@ -32,6 +32,37 @@ function ScreenWithScrolling(props: ScreenProps) {
...
+    screenHeight.current = contentHeight
+
+    // then update scroll state
+    updateScrollState()
+  }
+
+  // update scroll state on every render 
+  // when screenHeight isn't null
+  if (screenHeight.current !== null) 
+    updateScrollState()
```
jamonholmgren, line 64:
```suggestion
  if (screenHeight.current !== null) updateScrollState()
```

boilerplate/app/components/screen/screen.tsx
```diff
@@ -32,6 +32,37 @@ function ScreenWithScrolling(props: ScreenProps) {
...
+  // The followings for <Screen preset='auto'/>
+  // This will automatically disables scrolling if content fits the screen.
+  const { height } = Dimensions.get('window')
+  const screenHeight = React.useRef(null)
+  const [scrollEnabled, setScrollEnabled] = React.useState(true)
+
+  const updateScrollState = () => {
+    if (props.preset === 'auto'){
+      // check whether if content fits the screen
+      const contentFitsScreen = screenHeight.current > height * presets.auto.offset.percent - presets.auto.offset.point
```
jamonholmgren, line 44: I could be misreading this, but doesn't this variable indicate that the content _doesn't_ fit the screen anymore, and thus we need to enable scroll? Shouldn't it be something like `contentOverflowsScreen`?

boilerplate/app/components/screen/screen.tsx
```diff
@@ -32,6 +32,37 @@ function ScreenWithScrolling(props: ScreenProps) {
...
+        
+      // then toggle scroll state according to it, make sure it's not rendering twice
+      if (scrollEnabled && !contentFitsScreen) setScrollEnabled(false)
+      else if (!scrollEnabled && contentFitsScreen) setScrollEnabled(true)
```
jamonholmgren, line 48:
```suggestion
      // content is less than the size of the screen, so we can disable scrolling
      if (scrollEnabled && !contentFitsScreen) setScrollEnabled(false)
      
      // content is greater than the size of the screen, so let's enable scrolling
      if (!scrollEnabled && contentFitsScreen) setScrollEnabled(true)
```

jamonholmgren (changes requested):
We're close! Sorry, just a few more comments to address here.

boilerplate/app/components/screen/screen.tsx
martincpt: Nope, it indicates exactly the opposite. More precisely: Content is smaller than screen size.

How about using **ContentFitsWithinScreen**? Does it describe better in English?

boilerplate/app/components/screen/screen.tsx
martincpt: I'm fine with it. Any reason you removed that **else**? Much better to read for sure, but unnecessary computing.

boilerplate/app/components/screen/screen.tsx
jamonholmgren: I guess what I'm confused by is the `height` variable comes from `Dimensions.get("window")`, so the logic seems to be:

contentFitsScreen = the screen's height is _larger_ than the window's height

In my mind, it should be the opposite, right?

`height` indicates the window height (see line 37).

In fact, I don't see anything about the content height at all.

Can you walk me through the logic here? Sorry this is confusing to me!

boilerplate/app/components/screen/screen.tsx
jamonholmgren: I went through it in [this stream](https://youtu.be/bvNZ5fSrAfE?t=2408) and honestly this one change took me many retries until I settled on this. Haha

martincpt: Oh, I've just watched it. You spend a lot of time on this one. Makes sense. No more headache on this.

boilerplate/app/components/screen/screen.tsx
martincpt: All right, I've reviewed the code again. You are right. The variable name is wrong. Sorry about that!

Even `screenHeight` name is kinda wrong. You even mentioned that at one point in your stream. It's confusing because it can either mean the height of the device screen and the height of the component too.

So we better go with `scrollViewHeight` name there which is more clear and even looks better that all scroll related variables are grouped together. What do you think?

For `contentFitsScreen` the declaring really mean the opposite, I just did not catch this until I reviewed again. I'm gonna change the angle bracket there, and fix those two ifs according to it.

martincpt:
Wrong and misleading variable names have been fixed. Please review, @jamonholmgren.

boilerplate/app/components/screen/screen.tsx
```diff
@@ -44,6 +80,8 @@ function ScreenWithScrolling(props: ScreenProps) {
           style={[preset.outer, backgroundStyle]}
           contentContainerStyle={[preset.inner, style]}
           keyboardShouldPersistTaps={props.keyboardShouldPersistTaps || "handled"}
+          onContentSizeChange={onContentSizeChange}
```
jamonholmgren, line 83: Let's guard against running the onContentSizeChange unnecessarily.

```suggestion
          onContentSizeChange={props.preset === 'auto' ? onContentSizeChange : undefined}
```

jamonholmgren (approved):
Just one small change, and I think we're good to go!

martincpt:
Hey @jamonholmgren,

I've committed your suggestion but one of the CircleCI test still fails.

I told you this on one of your stream last week and you have reset the tests for but now expo tests fails. I think this is out of my authority. Source code must be ok as it was fine through several weeks.

Can you please look into it?

I want to close this PR already, hopefully with successful merge.

Thank you!

jamonholmgren:
@martincpt I think updating the branch should help with that, trying that now.

martincpt:
@jamonholmgren Thanks! Looks like that helped. Ready to merge?

Issue #1912, opened 2022-03-04
npx ignite-cli new MyApp failed


jiyuan12354:
What's going on?
ignite-cli cannot init/new project
---

Steps to reproduce

1.  just run `npx ignite-cli new MyApp`
`

 dengjianyuan@Jerrys-MacBook-Pro ~ npx ignite-cli new MyApp                                                                                                                                                                                                                                                                                    
   · · · · · · · · · · · · · · · · · · 🔥 Ignite 🔥 · · · · · · · · · · · · · · · · · ·

    █ Creating MyApp using Ignite 7.10.6
    █ Powered by Infinite Red - https://infinite.red
    █ Using ignite-cli with yarn
    █ Bundle identifier: com.MyApp
    ────────────────────────────────────────────────

    🔥 Igniting app
    🖨  3D-printing a new React Native app
    ⠴ Unboxing npm dependencies
/Users/dengjianyuan/.npm/_npx/58859/lib/node_modules/ignite-cli/node_modules/gluegun/build/index.js:13
    throw up;
    ^
warning expo > uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
warning expo > expo-constants > uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
warning expo > expo-file-system > uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
warning expo > fbemitter > fbjs > core-js@1.2.7: core-js@<3.4 is no longer maintained and not recommended for usage due to the number of issues. Because of the V8 engine whims, feature detection in old core-js versions could cause a slowdown up to 100x even if nothing is polyfilled. Please, upgrade your dependencies to the actual version of core-js.
warning react-native > @react-native-community/cli-platform-ios > xcode > uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
warning react-native > react-native-codegen > jscodeshift > micromatch > snapdragon > source-map-resolve@0.5.3: See https://github.com/lydell/source-map-resolve#deprecated
[...]

PR #1926, opened 2022-04-15
Removes npm-run-all which causes certain failures

jamonholmgren:
While spinning up a new Ignite app, I ran into some irritating issues where Ignite would fail with:

```shell
$ npm-run-all format:* ''
ERROR: Task not found: ""
```

For some reason, that following `''` is being appended.

I'm having trouble figuring out where, so instead of having `npm-run-all` at all, I'm just calling the format scripts directly and removing the `npm-run-all` package entirely. This is more explicit, removes a dependency, and solves this and IIRC potentially another issue where Windows doesn't work properly expand the `*` if you don't escape it properly.


robinheinze:
I wonder if this could this be related to #1912

jamonholmgren:
Looks like we're getting a TypeScript error on CI. I replicated it -- something to do with our ErrorBoundary class, which is taken mostly from RN docs.

```
    app/app.tsx:61:12 - error TS2786: 'ErrorBoundary' cannot be used as a JSX component.
      Its instance type 'ErrorBoundary' is not a valid JSX element.
        The types returned by 'render()' are incompatible between these types.
          Type 'import(\"/tmp/ignite-49f777cf87c17424f01054c8db113762/Foo/node_modules/@types/react/index\").ReactNode' is not assignable to type 'React.ReactNode'.
    
    61           <ErrorBoundary catchErrors={\"always\"}>
                  ~~~~~~~~~~~~
```

To debug this, spin up a new app and then run `yarn tsc -p .`.

I don't have time right now, unfortunately, but maybe someone else can carry this forward.

jamonholmgren:
> I wonder if this could this be related to #1912

Could be, @robinheinze!

dannyhw:
@jamonholmgren 

just stumbled onto this and I think I might know the issue with the types

If you check the react native template its including this resolution here: 

```
 "resolutions": {
    "@types/react": "^17"
  },
```

see https://github.com/react-native-community/react-native-template-typescript/blob/main/template/package.json

That's because the types for children changed in react 18 and somehow the types get pulled in via dependencies.

If I add this and resolution and yarn install then the typescript issues are fixed for me

PR #1950, opened 2022-05-23
Upgrades to React Native 0.68.2

jamonholmgren:
This PR upgrades to the latest version of React Native (0.68.2), as well as multiple other dependencies.

It's been tested locally with yarn and pnpm, and also is passing on CI.

Notable boilerplate changes:

* Upgrades to React Native 0.68.2, allowing for the (optional) new architecture
* .npmrc file added to avoid pnpm peer dependency issues
* Fixes recursive type issue in i18n files
* Fixes some remaining symlink issues

Notable other changes:

* Adds new `ignite-cli rename` feature and removes dependency on `react-native-rename`. More below!
* Added check for leftover "HelloWorld" files in CI tests to protect against rename issues
* CI has been updated to a newer image

#### Rename

We've been relying on the package `react-native-rename` for a few versions now, but it doesn't seem to be keeping pace with the changes in the React Native world. So, I've built my own version of this. It runs automatically when you spin up a vanilla Ignite project (non-Expo), but also can be run on any React Native project!

Just run:

```
npx ignite-cli rename
```

It'll ask for your new app name and bundle identifier and make the required changes.

If you run into problems, please report them!


jamonholmgren:
This is ready to test fly! :tada: 

flying

src/tools/react-native.ts
```diff
@@ -23,18 +23,15 @@ export async function copyBoilerplate(toolbox: GluegunToolbox, options: CopyBoil
   const { filesystem } = toolbox
   const { copyAsync, path } = filesystem
 
-  // ensure the desitnation folder exists
+  // ensure the destination folder exists
```
flexbox, line 26: 💯 🦅 👁️

flexbox (approved):

boilerplate/app/i18n/en.json
```diff
@@ -10,7 +10,12 @@
   "welcomeScreen": {
     "poweredBy": "POWERED BY IGNITE",
     "readyForLaunch": "Ready for launch.",
-    "continue": "CONTINUE"
+    "continue": "CONTINUE",
+    "more": {
+      "things": {
+        "here": "Here"
+      }
+    }
```
mazenchami, line 18: do you still need these?

src/commands/help.ts
```diff
@@ -46,6 +46,10 @@ module.exports = {
       ["ignite doctor"],
     )
     p()
+    command("rename      ", "Renames your React Native project (experimental)", [
+      "ignite rename NewName com.mycompany.newname",
+    ])
```
mazenchami, line 51: I know its experimental, but would be good to add visibility to the README file of this command

mazenchami (approved):
Looks good to me, great work! 🚀

src/tools/react-native.ts
```diff
@@ -45,3 +42,155 @@ export async function copyBoilerplate(toolbox: GluegunToolbox, options: CopyBoil
...
+
+      log(`Patching ${file} - ${oldName} to ${newName} and variants`)
+
+      // replace all instances of the old name and all its variants
+      const newContent = content
+        .replace(new RegExp(oldBundleIdentifier, "g"), newBundleIdentifier)
+        .replace(new RegExp(oldnamekebab, "g"), newnamekebab)
+        .replace(new RegExp(oldnamesnake, "g"), newnamesnake)
+        .replace(new RegExp(oldName, "g"), newName)
+        .replace(new RegExp(oldnamelower, "g"), newnamelower)
```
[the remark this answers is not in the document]
jamonholmgren: That's a good question, @kbrandwijk. I think that I'll rely on the developer to catch those, as they should be carefully examining the diff for any issues. I think I'll add that to the output (when running from the CLI) at the end, something like:

```
Caveats:
  * Ignite's rename feature is not 100% perfect in all cases. Carefully examine the diff it produces
    with `git diff` before committing.
  * Additionally, you'll want to re-run `pod install` and rebuild your app and caches, as we don't
    make any changes to those files or folders.
```

boilerplate/app/i18n/en.json
jamonholmgren: That was a TypeScript test, good catch!

good catch

src/commands/help.ts
jamonholmgren: I'm working on another branch with a bunch of new documentation and will definitely add it there. It doesn't really have a place to live right now, unfortunately. Good catch though!

src/tools/react-native.ts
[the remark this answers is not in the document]
jamonholmgren: Added that info at the end of running the rename.

image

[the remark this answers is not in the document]
kbrandwijk: Looks like a fair compromise to me! 👍🏻

jamonholmgren:
Having some CI issues that are preventing publication. Expo build is failing. I'm looking into it.

Issue #1963, opened 2022-06-20
Error after cloning fresh boilerplate. Unable to run.


Kirk-Str:
What's going on?

Unable to run a fresh project after cloning the repo. Fighting with it for the last one week. What's going on? 

I used to remove `expo-linear-gradient` fix the issue last week. But this new release throws entirely different issue now.

Steps to reproduce

1.  npx ignite-cli new PizzaApp --bundle=com.infinitered.pizzaapp
2. npx react-native run-android
3. getting the following error now.

*Also the android package name does not get changed.

The error output.

Try the new cross-platform PowerShell https://aka.ms/pscore6

PS D:\Workspace\Heilo\Cinnamon> npx react-native run-android
npm WARN config global `--global`, `--local` are deprecated. Use `--location=global` instead.
warn Package expo-modules-core has been ignored because it contains invalid configuration. Reason: "dependency.platforms.ios" must be of type 
object
info Running jetifier to migrate libraries to AndroidX. You can disable it using "--no-jetifier" flag.
Jetifier found 1578 file(s) to forward-jetify. Using 8 workers...
info Starting JS server...
info Installing the app...

> Configure project :app
Unable to detect AGP versions for included builds. All projects in the build should use the same AGP version. Class name for the included build object: org.gradle.composite.internal.DefaultIncludedBuild$IncludedBuildImpl_Decorated.

> Configure project :expo

Using expo modules
  - expo-application (4.1.0)
  - expo-constants (13.1.1)
  - expo-error-recovery (3.1.0)
  - expo-file-system (14.0.0)
  - expo-font (10.1.0)
  - expo-keep-awake (10.1.1)
  - expo-linear-gradient (11.3.0)
  - expo-localization (13.0.0)
  - expo-modules-core (0.9.2)

> Task :expo-constants:createDebugExpoConfig
Execution optimizations have been disabled for task ':expo-constants:createDebugExpoConfig' to ensure correctness due to the following reasons:
  - Gradle detected a problem with the following location: 'D:\Workspace\Heilo\Cinnamon\android'. Reason: Task ':expo-constants:createDebugExpoConfig' uses this output of task ':app:checkDebugAarMetadata' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.3.3/userguide/validation_problems.html#implicit_dependency for more details about this problem.
[...]

Issue #1954, opened 2022-05-30
Project installs but does not run


danilodorgam:
What's going on?
Error when trying to start a newly created project.
I've tried it on both Linux Mint and Windows 10 following the suggested course
---

Steps to reproduce

1. npx ignite-cli new PizzaApp
2. cd PizzaApp/
3. yarn android

---

`npx ignite-cli doctor` results:
System
  platform           linux                                                                 
  arch               x64                                                                   
  cpu                16 cores     Intel(R) Core(TM) i7-10700K CPU @ 3.80GHz                
  directory          PizzaApp     /home/danilodorgam/Documentos/projetos/PizzaApp 

JavaScript (and globally-installed packages)
  node               16.15.0      /home/danilodorgam/.nvm/versions/node/v16.15.0/bin/node 
  npm                8.5.5        /home/danilodorgam/.nvm/versions/node/v16.15.0/bin/npm  
    corepack         0.10.0                                                               
    npm              8.5.5                                                                
  yarn               1.22.19      /usr/bin/yarn                                           
  pnpm               -            not installed                                           

Ignite
  ignite-cli         7.10.10      /home/danilodorgam/.npm/_npx/e31027f3785124a8/node_modules/.bin/ignite      
  ignite src         build        /home/danilodorgam/.npm/_npx/e31027f3785124a8/node_modules/ignite-cli/build 

Android
  java               11.0.15      /usr/bin/java                  
  android home       -            /home/danilodorgam/Android/Sdk

BUILD FAILED in 6s

    at makeError (/home/danilodorgam/Documentos/projetos/PizzaApp/node_modules/execa/index.js:174:9)
    at /home/danilodorgam/Documentos/projetos/PizzaApp/node_modules/execa/index.js:278:16
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async runOnAllDevices (/home/danilodorgam/Documentos/projetos/PizzaApp/node_modules/@react-native-community/cli-platform-android/build/commands/runAndroid/runOnAllDevices.js:109:5)
    at async Command.handleAction (/home/danilodorgam/Documentos/projetos/PizzaApp/node_modules/@react-native-community/cli/build/index.js:192:9)
info Run CLI with --verbose flag for more details.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

HAFDIAHMED:
@jamonholmgren 

please help us many projects are generated by ignite these days but they cannot be runned using ignite ???

[...]

PR #1930, opened 2022-04-22
Default App/Launcher Icons; App Icon Generator

yulolimum:
## Please verify the following:

- [x] `yarn ci:test` **jest** tests pass with new tests, if relevant
- [x] `README.md` has been updated with your changes, if relevant

## Describe your PR

❗❗❗ This PR depends on #1964 ❗❗❗ 

### Notes:

- adds default Ignite iOS app icons and registers them via Content.json
- adds default Ignite Android launcher icons - both legacy and adaptive
- adds default Ignite Expo launcher icons and references them in app.json
- creates a new Ignite generator called `app-icon`
- adds icon template files to `ignite/templates/app-icon`

##### new icons on Expo
yulolimum-capture-2022-04-21--13-35-09

##### new icons on Vanilla
yulolimum-capture-2022-04-21--14-27-56

##### generator CLI validations
yulolimum-capture-2022-04-21--14-15-22
yulolimum-capture-2022-04-21--14-18-26

##### generator successful output
yulolimum-capture-2022-04-21--14-21-02

##### android adaptive icon demo
yulolimum-capture-2022-04-21--14-32-37

PR #1964, opened 2022-06-22
Downgrades Some Deps for Expo-Flavor.

yulolimum:
There are some open issues mentioning Ignite not running and reference `expo-linear-gradient` (e.g. #1963 , #1954). Those aren't 100% related (since they are specific to vanilla-flavor), but I did confirm that a freshly ignited app (using expo-flavor) does not build. 

| iOS Output  | Android Output |
| ------------- | ------------- |
| yulolimum-capture-2022-06-21--19-39-36@2x | yulolimum-capture-2022-06-21--19-41-51@2x |

While starting the Expo dev server, I noticed this:

yulolimum-capture-2022-06-21--19-42-31@2x

Recently, there have been some new Ignite releases (upgrading to 0.68.x). The expo core dep got updated in vanilla-flavor, but not in expo-flavor (possibly due to expo-cli [prerequisite issue](https://github.com/expo/expo-cli/issues/4423)?). In any case, it's a good idea to always use recommended expo library versions. Upgrading core without libraries (or vice versa) can cause things to break. The real fix to this is to upgrade the expo-flavor boilerplate to latest and make sure we continue overriding the expo-flavor deps/libs according to the recommended. 

I downgraded the packages (in the expo-flavor package.json) that have jumped major versions compared to the recommended. The app now builds: 

yulolimum-capture-2022-06-21--19-51-02@2x

PS: this won't fix the tagged issues unless they re-ignite their apps which isn't practical.

### DIFF
--- a/boilerplate/package.expo.json
+++ b/boilerplate/package.expo.json
@@ -12,10 +12,14 @@
   },
   "dependencies": {
     "@expo/webpack-config": "^0.12.71",
-    "expo": "43.0.0",
+    "expo-linear-gradient": "~10.0.3",
+    "expo-localization": "~11.0.0",
+    "expo-modules-core": "~0.4.10",
     "expo-status-bar": "~1.1.0",
-    "react": "17.0.1",
-    "react-native": "https://github.com/expo/react-native/archive/sdk-43.0.0.tar.gz"
+    "expo": "43.0.0",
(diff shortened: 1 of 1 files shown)


mazenchami:
This PR looks good, closing in favor of #1930 (which includes this fix also)

Issue #1927, opened 2022-04-15
Solidarity failures cause confusing error on Ignite new


robinheinze:
While debugging #1912, I noticed that any solidarity failures (like incorrect cocoapods version for example) cause Ignite new to fail very vaguely. 

Steps to reproduce

1. Set cocoapods version to 1.10
2. Run `npx ignite-cli new Pizza App`

GantMan:
You can loosen the solidarity rules, or even remove solidarity completely imo.

robinheinze:
It's still nice to have to ensure folks have the right environment, I just need to figure out why the actual error isn't bubbling up through Gluegun.

yulolimum:
@jamonholmgren since Solidarity is removed, this issue could be closed?

PR #1968, opened 2022-06-28
Removes Solidarity, simplifies postInstall script

jamonholmgren:
After polling the Infinite Red internal React Native team, [Solidarity](https://github.com/infinitered/solidarity) is being used on a few of the projects, but there isn't quite enough support to justify its place as an Ignite default. Some issues have been reported using Solidarity (#1927 most recently), and it has failed in random places on CI and EAS Build.

Given that, I'm removing it from both the CLI and the boilerplate. I also took the opportunity to simplify and de-cleverize the ./bin/postInstall script.


mazenchami (approved):
🚀  Looks good to me. With this merged, we might be able to close some issues!

Issue #1945, opened 2022-05-16
Boilerplate Enhancement: Crash Reporting


jamonholmgren:
Crash reporting is one of the first things we add to new Ignite projects. It would be great to have a defined place to add this.

My current idea is to add a `crash-reporting.ts` file to the boilerplate with some comments:

```ts
/**
 * This is where you put your crash reporting service initialization code.
 * If you're using Crashlytics: <link to Crashlytics docs>
 * If you're using Sentry: <link to Sentry docs>
 * If you're using Bugsnag: <link to Bugsnag docs>
 */
```

We could potentially add a generic interface that works with all three of those systems, so it's easier to swap them in and out:

```ts
export function reportCrash(error) {
  // Sentry.captureException(error)
  // crashlytics().recordError(error)
  // Bugsnag.notify(error)
}
```

Open to additional feedback on this. I would recommend putting this file in [`app/config/crash-reporting.ts`](https://github.com/infinitered/ignite/tree/master/boilerplate/app/config)

frankcalise:
Implemented in #1996

Will be included in Ignite v8: Maverick #1987

PR #1996, opened 2022-08-08
Adds Crash Reporting boilerplate

frankcalise:

## Describe your PR

Implements #1945 by adding boilerplate `utils/crash-reporting.ts`

- Provides user with setup guidance for several crash reporting services as Sentry, crashlytics and Bugsnag.
- Adds a normalized `reportCrash` utility function to send debug output in console, Reactotron and to crash service of choice.


jamonholmgren (approved):
Looks good. We'll probably want to put a note in the docs somewhere about this.

boilerplate/app/utils/crash-reporting.ts
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Manually report a handled error.
+ */
+export const reportCrash = (error: any, type: ErrorType = ErrorType.FATAL) => {
+  const message = error.message || "Unknown"
+  console.error(error)
+  console.log(message, type)
+
+  __DEV__ && console.tron.log(error)
```
jamonholmgren, line 57: Minor nit -- if you use the && format, Metro won't strip it out of the final build (although it will still work as expected). If you change it to an if statement, Metro will strip it out of the prod build.

```suggestion
  if (__DEV__) {
    console.tron.log(error)
  }
```

jamonholmgren: We should probably also early exit here. We don't want to report bugs in dev.

Issue #1982, opened 2022-07-22
New Ignite walkthrough


jamonholmgren:
I’ve been rethinking the whole “Ignite CLI runs without asking any questions” thing we have going on. There are some benefits to it, but there’s also some benefit of asking the questions in a walkthrough where it can feel exciting / inspire more confidence.

```
$ npx ignite-cli new

🔥 About to Ignite your new app! 🔥

 ↣ What do you want to call it? PizzaApp
 ↣ What bundle identifier? com.infinitered.pizzaapp
 ↣ Want to use Expo? Y/n
 ↣ Want us to install dependencies for you? (adds 50-100 seconds) Y/n

In the future, if you'd like to skip the questions, you can run Ignite with these options:

  npx ignite-cli new PizzaApp --bundle=com.infinitered.pizzaapp --expo --install-deps

<Ignite continues>
```

Note that we give them the command for skipping the questions next time, if they want.

Primarily, this surfaces the Expo and bundle identifier options, which are somewhat hidden otherwise.

I think this makes sense to include in Maverick.

joshuayoes:
I was thinking of running a similar behavior of `npm init`, where you have a few options:
- By default, all `npm init` will prompt you to go through all the options, with yes or no, and give a text field if the option requires input
- If you pass in `npm init -y`, you are saying "yes" to all the question prompts

For `npx ignite-cli new`, we should have the following behavior

| Command  | Question Prompts? | Description
| ------------ | --------------------- | -------------
| `npx ignite-cli new`  | Yes | Go through each option as a question prompt, with a command outputted with flags for selected options. Example: `npx ignite-cli new PizzaApp --bundle=com.infinitered.pizzaapp --expo --install-deps`
| `npx ignite-cli new PizzaApp`  | Yes | Go through each option as a question prompt, with a command outputted with flags for selected options. Example: `npx ignite-cli new PizzaApp --bundle=com.infinitered.pizzaapp --expo --install-deps`. Since name being the first parameter is already the existing behavior, I think still prompting a question is important since most users will go through this flow. 
| `npx ignite-cli new PizzaApp --bundle=com.infinitered.pizzaapp --expo --install-deps`  | No |  Go straight to initializing
| `npx ignite-cli new PizzaApp --install-deps`  | No |  If only partial options are passed, assume existing defaults for not included flag options
| `npx ignite-cli new PizzaApp --y`  | No |  Current CLI behavior, use all default options and skip straight to initializing

[...]

PR #1999, opened 2022-08-10
Add prompts and flags for `ignite new` command options 

joshuayoes:

## Describe your PR
Implements new `npx ignite new` behavior described in https://github.com/infinitered/ignite/issues/1982


src/commands/new.ts
```diff
@@ -12,40 +12,123 @@ import {
...
+   *
+   * Input Source: `prompt.confirm` | `parameter.option`
+   * @default false
+   */
+  coloLoco?: boolean
+}
+
 export default {
   run: async (toolbox: GluegunToolbox) => {
+    // #region Toolbox
```
jamonholmgren, line 98: Let's remove the #region marks.

jamonholmgren (commented):
A couple questions / changes ... thanks!

src/commands/new.ts
```diff
@@ -158,15 +255,22 @@ export default {
...
     // await packager.add(`react-native-safe-area-context`, packagerOptions)
+    // #endregion
 
+    // #region Run Packager Install
     // pnpm/yarn/npm install it
-    startSpinner("Unboxing npm dependencies")
-    await packager.install({ ...packagerOptions, onProgress: log })
-    stopSpinner("Unboxing npm dependencies", "🧶")
+
+    if (normalize.boolean(options.installDeps) !== false) {
```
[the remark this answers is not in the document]
joshuayoes: This value is the actual string value, but I thought it would be nice to have them match the options interface type

src/commands/new.ts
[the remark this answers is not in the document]
joshuayoes: How would you feel about me wrapping these regions in functions? Being able to fold large chunks of code like this is really nice for me to go about a codebase

joshuayoes: See: https://github.com/infinitered/ignite/pull/1999#discussion_r948481135

test/_test-helpers.ts
```diff
@@ -16,8 +16,8 @@ export async function runIgnite(cmd: string, options: RunOptions = {}): Promise<
 }
 
 export async function run(cmd: string, options: RunOptions = {}): Promise<string> {
-  const pre = options.pre ? `${options.pre} ; ` : ""
-  const post = options.post ? ` ; ${options.post}` : ""
+  const pre = options.pre ? `${options.pre} && ` : ""
+  const post = options.post ? ` && ${options.post}` : ""
```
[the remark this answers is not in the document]
joshuayoes: This change will also be included after this MR is merged: https://github.com/infinitered/ignite/pull/2009

src/tools/normalize.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Windows uses `\` for slash, Unix uses `/` :'(
+ *
+ * Normalize path to use `/` as path separator.
+ */
+export const normalizePath = (input: string) => input.replace(/\\/g, "/")
+
+export const normalize = {
+  boolean: normalizeBoolean,
+  path: normalizePath,
```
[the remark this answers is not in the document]
joshuayoes: `normalizePath` was part of some code after I was trying to get the test suites to pass in Windows. Many of the asserts weren't working because directory slashes are different on Windows vs Unix. 

Eventually I abandoned doing this on windows, in favor of WSL Ubuntu and now MacOS. I can remove this utility

src/commands/new.ts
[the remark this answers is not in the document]
joshuayoes: When you have a 400 line file, it's nice to be able to fold logic into discreet chunks

Screen Shot 2022-08-17 at 3 40 46 PM

src/commands/new.ts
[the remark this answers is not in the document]
jamonholmgren: I checked and the value is indeed a boolean, so we don't need this unless I'm missing something.

image


image

src/commands/new.ts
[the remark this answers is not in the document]
jamonholmgren: That does look pretty cool. You win this one, Josh!

you win on a technicality

src/commands/new.ts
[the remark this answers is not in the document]
joshuayoes: I'll chock this up a windows machine shenanigans 😆

jamonholmgren (approved):

src/commands/new.ts
```diff
@@ -12,84 +12,291 @@ import {
...
+    // #endregion
+
+    // #region Project Name
+    heading("🔥 About to Ignite your new app! 🔥")
 
     // retrieve project name from toolbox
-    const { validateProjectName, validateBundleIdentifier } = require("../tools/validations")
-    const projectName = validateProjectName(toolbox)
+    const { validateProjectName } = require("../tools/validations") as ValidationsExports
+    const projectName = await validateProjectName(toolbox)
```
joshuayoes, line 143: Added a `prompt.ask` inside the validator if the project name is empty: https://github.com/infinitered/ignite/pull/1999/files#diff-7920917860492dfa617b4e6b53368ad95f26db81b3edb99a38336c578d56971bR12-R16

src/commands/new.ts
```diff
@@ -12,84 +12,291 @@ import {
   stopSpinner,
   clearSpinners,
 } from "../tools/pretty"
+import type { ValidationsExports } from "../tools/validations"
```
joshuayoes, line 15: Nice thing about `import type` is that Typescript can remove it from the bundle, so we don't need to worry about these type definition imports slowing down start up time

src/commands/new.ts
```diff
@@ -12,84 +12,291 @@ import {
...
+        name: "bundleIdentifier",
+        message: "What bundle identifier?",
+        initial: defaultBundleIdentifier,
+      })
+
+      bundleIdentifier = bundleIdentifierResponse.bundleIdentifier
+    }
+
+    const { validateBundleIdentifier } = require("../tools/validations") as ValidationsExports
+    validateBundleIdentifier(toolbox, bundleIdentifier)
```
joshuayoes, line 175: Usage of this has slightly changed. `validateBundleIdentifier` returns a boolean, but we only care about calling it will log a warning and call `process.exit(1)` if there is an error.

src/commands/new.ts
```diff
@@ -12,84 +12,291 @@ import {
...
+    }
+    // #endregion
 
+    // #region Packager
     // check if a packager is provided, or detect one
     // we pass in expo because we can't use pnpm if we're using expo
-    const packagerName = parameters.options.packager || packager.detectPackager()
+
+    const availablePackagers = packager.availablePackagers()
+    const defaultPackagerName = availablePackagers.includes("yarn") ? "yarn" : "npm"
```
joshuayoes, line 233: According to this twitter poll, it seems that people overwhelming prefer `yarn` for React Native development: https://twitter.com/JoshuaYoes/status/1557809021905563649

Previously, we preferred pnpm -> yarn -> npm. Anecdotally, I had a lot of issues with using `pnpm`, so I think it would be best to prefer yarn in this case

src/commands/new.ts
```diff
@@ -12,84 +12,291 @@ import {
...
+
+    if (packagerName === undefined) {
+      const initial = availablePackagers.findIndex((p) => p === defaultPackagerName)
+      const NOT_FOUND = -1
+
+      if (initial === NOT_FOUND) {
+        p()
+        p(yellow(`Error: Default packager "${defaultPackagerName}" was not available on system`))
+        process.exit(1)
+      }
```
joshuayoes, line 259: This should never happen, but I felt it was important to document since it wasn't intuitive to me that `findIndex` returns -1 if not found

src/commands/new.ts
```diff
@@ -180,19 +399,27 @@ export default {
...
-      onProgress: log,
-    })
-    stopSpinner("Baking CocoaPods", "☕️")
+    if (installDeps === true) {
+      startSpinner("Baking CocoaPods")
+      await spawnProgress(`npx pod-install@${deps.podInstall}`, {
+        onProgress: log,
+      })
+      stopSpinner("Baking CocoaPods", "☕️")
+    }
```
joshuayoes, line 412: This throws an error if we don't run packager install before, so we don't call this unless we have installed packager dependencies

src/tools/packager.ts
```diff
@@ -38,17 +39,29 @@ function pnpmAvailable() {
...
-function detectPackager(): "npm" | "yarn" | "pnpm" {
-  // Expo doesn't support pnpm, so we'll use yarn or npm
-  if (pnpmAvailable()) {
-    return "pnpm"
-  } else if (yarnAvailable()) {
+function detectPackager(): PackagerName {
+  if (yarnAvailable()) {
     return "yarn"
+  } else if (pnpmAvailable()) {
+    return "pnpm"
```
joshuayoes, line 46: [See this comment](https://github.com/infinitered/ignite/pull/1999/files#r950597888) for justification of preferring yarn over pnpm:

test/vanilla/ignite-new.test.ts
```diff
@@ -16,20 +16,15 @@ afterEach(() => {
 })
 
 describe("Checking for ignite. 🪔", () => {
-  test(`ignite new (no name)`, async () => {
-    const result = await runError(`new`)
-    expect((result as any).stdout).toContain(`Project name is required`)
-  })
```
joshuayoes, line 22: [It's not possible for project name to be blank anymore](https://github.com/infinitered/ignite/pull/1999/files#diff-7920917860492dfa617b4e6b53368ad95f26db81b3edb99a38336c578d56971bR12-R17), so this test will hang forever and isn't a valid test case anymore

README.md
```diff
@@ -38,13 +38,12 @@ Prerequisites:
...
-# for vanilla React Native
+# Get walked through the prompts for the different options to start your new app
 npx ignite-cli new PizzaApp
-# to provide a custom bundle identifier (Android only):
-npx ignite-cli new PizzaApp --bundle=com.infinitered.pizzaapp
-```
 
-Ignite will walk you through the rest.
+# Accept all the recommended defaults and get straight to coding!
+npx ignite-cli new PizzaApp --y
```
jamonholmgren, line 45: The `--y` vs `-y` confuses me a bit here (we chatted about this).

I looked around and this seems to be the place that the standard of `-y` vs `--yes` comes:

[https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html)

So let's do `--yes` and `-y` but not `--y`.

src/commands/new.ts
jamonholmgren: That makes sense, and I think my enthusiasm for pnpm has been tempered since I made this change.

src/commands/new.ts
jamonholmgren: Fair enough! JS is weird sometimes.

src/commands/new.ts
jamonholmgren: 💯  I definitely include this in "deps", so this is appropriate.

jamonholmgren (approved):
One small thing to update (the `--y` thing) and then we're good to merge!

README.md
frankcalise: I agree, the convention is the double dash is the more verbose name with the single dash for the shorthand notation

src/commands/new.ts
joshuayoes: Honestly, pnpm has been an amazing package manager for me for traditional JS projects. Install times are pretty unbeatable. But I think their strategy of system links seems to be more buggy with native dependencies for react native

README.md
joshuayoes: Good catch, I forgot to update this. Commit is up now with a `-y` and `--yes` flag

PR #2005, opened 2022-08-15
Adds observer to navigator, other navigator cleanup

### LINKED ISSUE #1918: app-navigator should be observer component to detect MST
What's going on?
app-navigator does not listen to MST changes  
---

Steps to reproduce
1. use any store inside app-navigator
2. inside app-navigator UI use store prop to show UI based on conditions like 
 {authenticationStore.isAuthenticated ?
                `<AppStack/>`
                :
               ` <AuthStack />`
            }
3- observe no change happened when store prop value changed.

jamonholmgren:
Fixes #1918 by adding `observer` to the navigator stack and container.

Also cleans up the back button handler code a bit.


boilerplate/app/navigators/app-navigator.tsx
```diff
@@ -74,18 +83,4 @@ export function AppNavigator(props: NavigationProps) {
       <AppStack />
     </NavigationContainer>
   )
-}
-
-AppNavigator.displayName = "AppNavigator"
```
jamonholmgren, line 79: No longer needed if we use the `const X = observer(function X() { ... })` trick

boilerplate/app/navigators/navigation-utilities.tsx
```diff
@@ -39,6 +39,9 @@ export function getActiveRouteName(state: NavigationState | PartialState<Navigat
  * the navigation or allows exiting the app.
  */
 export function useBackButtonHandler(canExit: (routeName: string) => boolean) {
+  // ignore if iOS ... no back button!
+  if (Platform.OS === "ios") return
```
jamonholmgren, line 43: Breaks a hooks rule, but since this will never change on the fly somehow, shouldn't be an issue.

silasjmatson (approved):
👍

PR #2003, opened 2022-08-12
Framework Upgrades, RN 0.69.4, Expo 46

frankcalise:
## Please verify the following:

- [x] `yarn ci:test` **jest** tests pass with new tests, if relevant
- [x] `README.md` has been updated with your changes, if relevant

## Describe your PR

Implementation for #1979 

- Expo 46
- React Native 0.69.4
- React 18
- Removed validate.js and utils/validate.ts
- `yarn expo:web` has issues running via Expo's new local CLI (can still run via `expo-cli expo start --web`

PR #2009, opened 2022-08-15
Fix screen and component generator output in `ignite new` test case

joshuayoes:

### Problem
`ignite-new.test.ts` was silently failing when running `yarn compile` command on test generator output, but not throwing an error.

This has been a blocker for changes in [another PR I am working on](https://github.com/infinitered/ignite/pull/1999). I wanted to ensure that the existing test cases pass before making changes 

### Changes
- Update `ignite-new.test.ts` to use static `temp` folder instead of random one on system
- Fix generator `.ejs` files producing files with typescript errors
- Change test in `ignite-new.test.ts` to use `yarn` instead of `npm`


boilerplate/ignite/templates/component/NAME.tsx.ejs
```diff
@@ -19,7 +19,7 @@ const CONTAINER: ViewStyle = {
 const TEXT: TextStyle = {
```
yulolimum, line 19: let's update these to be like `$camelCase` and moved to bottom of file.

boilerplate/ignite/templates/component/NAME.tsx.ejs
```diff
@@ -19,7 +19,7 @@ const CONTAINER: ViewStyle = {
...
   fontFamily: typography.primary.normal,
   fontSize: 14,
-  color: color.primary,
+  color: colors.palette.primary100,
```
yulolimum, line 22: let's use something a bit darker here. maybe a 500 level

boilerplate/ignite/templates/screen/NAME-screen.tsx.ejs
```diff
@@ -8,34 +8,33 @@ import React, { FC } from "react"
...
 import { ViewStyle } from "react-native"
 import { StackScreenProps } from "@react-navigation/stack"
-import { NavigatorParamList } from "../../navigators"
+import { AppStackParamList } from "../../navigators"
 import { Screen, Text } from "../../components"
 // import { useNavigation } from "@react-navigation/native"
 // import { useStores } from "../../models"
-import { color } from "../../theme"
 
 const ROOT: ViewStyle = {
```
yulolimum, line 16: same here

boilerplate/ignite/templates/screen/NAME-screen.tsx.ejs
```diff
@@ -8,34 +8,33 @@ import React, { FC } from "react"
 import { observer } from "mobx-react-lite"
```
yulolimum, line 8: we probably want the screen name to be like `ExampleScreen`, not `example-screen`. Not sure exactly where this needs to change.

yulolimum (approved):
Couple notes, otherwise good to go!

boilerplate/ignite/templates/screen/NAME-screen.tsx.ejs
joshuayoes: I think that is a good idea, it would match the contention we already have.

We'd need to add some logic to the [`src/tools/generators.ts`](https://github.com/infinitered/ignite/blob/master/src/tools/generators.ts#L197), which I feel is a little out of scope for this PR. I mainly wanted to get the generator tests passing again in this PR, while I feel this is more of a feature request.

yulolimum: Works for me! I was going off of the PR title and assumed haha :)

joshuayoes: Fair enough, thanks for taking a look!

test/_test-helpers.ts
```diff
@@ -16,8 +16,8 @@ export async function runIgnite(cmd: string, options: RunOptions = {}): Promise<
 }
 
 export async function run(cmd: string, options: RunOptions = {}): Promise<string> {
-  const pre = options.pre ? `${options.pre} ; ` : ""
-  const post = options.post ? ` ; ${options.post}` : ""
+  const pre = options.pre ? `${options.pre} && ` : ""
+  const post = options.post ? ` && ${options.post}` : ""
```
joshuayoes, line 20: Changed `;` output to `&&` because I was having issues on windows with the `cd` pre and post command failing due to an invalid directory, then the main run `cmd` still continuing on, which lead to indirect error messages.

This way, `run` will fail now if the pre or post command fails.

test/vanilla/ignite-new.test.ts
```diff
@@ -28,8 +17,16 @@ describe("Checking for ignite. 🪔", () => {
 })
 
 describe("Igniting new app! 🔥\nGo get a coffee or something. This is gonna take a while.", () => {
+  const originalDir = process.cwd()
+  const tempDir = path.join(originalDir, "temp")
+  const appDir = path.join(tempDir, APP_NAME)
+
+  beforeEach(() => {
+    filesystem.remove(appDir)
+  })
```
joshuayoes, line 26: `tempy` was having some strange behavior on my windows machine, sometimes putting the file into inaccessible file locations and causing later tests to fail after not being able to `cd` back and forth from `tempDir` to `originalDir`. This way, it is always deterministic. 

I also changed the test to only clean the directory before each tests so I could inspect the `temp/Foo` output on failing tests.

Might be worth moving this into the `test/` directory to make it more clear of it's purpose.

test/vanilla/ignite-new.test.ts
```diff
@@ -87,7 +84,7 @@ async function checkForLeftoverHelloWorld(filePath: string) {
 
   // check to make sure there are no instances of helloworld or HelloWorld or hello-world
   // anywhere in the app -- including folder and filenames.
-  const appFiles = filesystem.list(filePath)
+  const appFiles = filesystem.list(filePath) ?? []
```
joshuayoes, line 87: This can return undefined, so I added fallback so the for loop later doesn't fail

test/vanilla/ignite-new.test.ts
```diff
@@ -1,20 +1,9 @@
 import { filesystem } from "gluegun"
-import * as tempy from "tempy"
+import * as path from "path"
```
jamonholmgren, line 2: I think I'd rather keep the `tempy` solution. We used to have a local gitignored folder for this back in Ignite 1.0 days, but went with the tempy solution instead. I forget the exact reasoning, but it had to do with running tests in parallel and not getting conflicts between integration tests.

When I'm debugging the integration tests, I comment out the `filesystem.remove(tempDir)` in the `afterEach` block and instead do `console.error(tempDir)`. It'll leave the folder as-is and tell me where it is, so I can go examine it.

test/_test-helpers.ts
```diff
@@ -44,7 +44,7 @@ export async function testSpunUpApp(appPath: string, originalDir: string) {
   // run typescript
   let resultTS: string
   try {
-    resultTS = await run(`npm run compile`, runOpts)
+    resultTS = await run(`yarn compile`, runOpts)
```
jamonholmgren, line 47: I think generally speaking we should keep `npm run` anywhere that it will work, as it's sort of the "safe" choice.

test/_test-helpers.ts
```diff
@@ -282,8 +282,8 @@ export async function testSpunUpApp(appPath: string, originalDir: string) {
   await run(`git commit -m "generated test components & assets"`, runOpts)
 
   // run the tests; if they fail, run will raise and this test will fail
-  await run(`npm run test`, runOpts)
-  await run(`npm run lint`, runOpts)
-  await run(`npm run compile`, runOpts)
+  await run(`yarn test`, runOpts)
+  await run(`yarn lint`, runOpts)
+  await run(`yarn compile`, runOpts)
```
jamonholmgren, line 287: Same as above, stick with `npm run` unless it really does not work.

test/vanilla/ignite-new.test.ts
jamonholmgren: Is this a known issue with tempy on Windows? I am reluctant to change this for erratic Windows behavior unless we have to.

jamonholmgren (commented):
Some comments added!

test/vanilla/ignite-new.test.ts
joshuayoes: Right now, we have one giant test for the new command, so there isn't any conflicts. If we add more tests, what I've done in past projects is add a unique ID to the `temp` folder for each artifact, either with a UUID or a timestamp

joshuayoes: I'm not a huge fan of needed to comment and uncomment code if I can help it, there are a lot of times I've accidentally committed debugging code.

Originally I was trying to get this work in Windows, that led me to the `temp` solution. I have a small SSD drive for my OS, and a data drive, so that may be part of the `cd` command is failing. However, I have switched a WSL Ubuntu environment, which may not make `tempy` as much of an issue, where all of WSL is encapsulated on one drive. 

I'll explore this more to see if `tempy` is an issue on WSL

joshuayoes: `tempy` appears to work fine with WSL. I'll revert these changes then

test/vanilla/ignite-new.test.ts
joshuayoes: Upon some further testing, the test suite fails on windows for a variety of reasons. 

`tempy` may be an issue on windows, but I've switched to WS.

See more here: L https://github.com/infinitered/ignite/pull/2009#discussion_r946208419

boilerplate/package.json
```diff
@@ -110,7 +110,7 @@
     "prettier": "2.6.2",
     "query-string": "^7.0.1",
     "react-devtools-core": "4.24.7",
-    "react-dom": "^18.0.0",
+    "react-dom": "18.0.0",
```
joshuayoes, line 113: Pinned this version because `npm install` was throwing a peer dependency warning.

Previously:
`"react": "18.0.0"` 
`"react-dom": "^18.0.0"` -> `"react-dom":"18.0.2"`

`"react-dom":"18.0.2"` has a peer dependency of `"react": "18.0.2"`, but only 18.0.0 was getting installed. 

Now: 
`"react": "18.0.0"` 
`"react-dom": "18.0.0"`

jamonholmgren: Shouldn't we upgrade React to 18.0.2?

joshuayoes: That would also likely work, this change was made from another PR

joshuayoes: https://github.com/infinitered/ignite/pull/2003/files#diff-594f75b3ec0c76210a39b8c7ce238eea438454a93d830bd968a282f89e30a8b0

yulolimum (approved):

boilerplate/package.json
joshuayoes: Updating React to `18.0.2` causes more peerDependency errors, which would involve updating `react-native`. I think sticking with `18.0.0` for this PR would be the more stable change

Screen Shot 2022-08-17 at 3 27 39 PM

PR #2014, opened 2022-08-15
Update docs for AutoImage and Icon components

silasjmatson:

## Describe your PR

* Update docs for AutoImage
* Update docs for Icon
* Update `IconProps` to type the `onPress` prop
* Update `Icon` component to only pass `color` if present

### DIFF
--- a/boilerplate/app/components/Icon.tsx
+++ b/boilerplate/app/components/Icon.tsx
@@ -37,6 +37,11 @@ interface IconProps extends TouchableOpacityProps {
    * Style overrides for the icon container
    */
   containerStyle?: StyleProp<ViewStyle>
+
+  /**
+   * An optional function to be called when the icon is pressed
+  */
+  onPress?: () => void
 }
 
 /**
@@ -63,7 +68,7 @@ export function Icon(props: IconProps) {
       <Image
         style={[
           $imageStyle,
-          { tintColor: color },
+          color && { tintColor: color },
           size && { width: size, height: size },
           $imageStyleOverride,
         ]}
--- a/docs/Components-AutoImage.md
+++ b/docs/Components-AutoImage.md
@@ -1,15 +1,11 @@
 # AutoImage Component
 
-MAVERICKTODO: update documentation with notes that this component is really only useful when the image source is a remote URI or a data uri scheme. for local images (those loaded with require()), this isn't needed. as per RN docs: https://reactnative.dev/docs/image
-
 [Back to all components](./Components.md)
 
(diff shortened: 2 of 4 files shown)


boilerplate/app/components/Icon.tsx
```diff
@@ -37,6 +37,11 @@ interface IconProps extends TouchableOpacityProps {
    * Style overrides for the icon container
    */
   containerStyle?: StyleProp<ViewStyle>
+
+  /**
+   * An optional function to be called when the icon is pressed
+  */
+  onPress?: () => void
```
yulolimum, line 44: do we need this? it's already inherited from TouchableOpacityProps above. or you're just adding it so we get the custom ts doc comment?

boilerplate/app/components/Icon.tsx
```diff
@@ -63,7 +68,7 @@ export function Icon(props: IconProps) {
       <Image
         style={[
           $imageStyle,
-          { tintColor: color },
+          color && { tintColor: color },
```
yulolimum, line 71: i don't think we necessarily needs this. if color is undefined, well it's undefined. i'm being nitpicky though :)

yulolimum (approved):

boilerplate/app/components/Icon.tsx
jamonholmgren: Will this overwrite `$imageStyle`'s tintColor though?

silasjmatson: I mostly did it so it would match how `size` was handled one line down. Also, if someone defined a default `tintColor` in the `$imageStyle`, not having this guard would cause a bug.

yulolimum: Yes it will. But $imageStyle doesn't have a tintColor (the default tintColor is the exported icon color since sometimes icons will have multiple colors: e.g. ladybug).

boilerplate/app/components/Icon.tsx
silasjmatson: Adding this removes a red squiggly on `onPress` in the de-structuring of the props. 😄

yulolimum: I think the red squiggly was there because you haven't run `yarn` in the boilerplate thus extending a blank type. once you install the deps, the squiggly disappears.

yulolimum: before/after `yarn`

yulolimum-capture-2022-08-15--16-23-41@2x

silasjmatson: Ah, I didn't even think of that. This might be sunk cost fallacy at work, but I think I'd prefer to keep `onPress` specifically called out here, because we use that to switch between `TouchableOpacity` and `View` for the icon wrapper. Thoughts? Low GAFO here, I can revert this change if you feel strongly about it.

yulolimum: I'm cool with keeping it bro! In which case, let's type it as `onPress?: TouchableOpacityProps["onPress"]` since the function actually has arguments.

PR #2018, opened 2022-08-16
switch to section list and add container around items per design

kateinkim:

## Describe your PR

1. To make drawer menu to be able to skip to the right items or sections by clicking, switched `FlatList` on the component screen to `SectionList` 

2. Added container around items per design

## Photos and Videos

|Container per design|Container in app|Scroll to section|
|-|-|-|
|CleanShot 2022-08-16 at 08 56 34@2x|CleanShot 2022-08-16 at 08 57 08@2x|CleanShot 2022-08-16 at 08 46 52|


kateinkim:
Removing extra space using `viewOffset` when scrolling to `title` by clicking drawer menu. While keeping that margin on the screen layout per design.

|before|after|action|
|-|-|-|
|CleanShot 2022-08-16 at 09 35 27@2x|CleanShot 2022-08-16 at 09 35 07@2x|CleanShot 2022-08-16 at 09 37 27|

boilerplate/app/screens/DemoComponentsScreen/DemoItem.tsx
```diff
@@ -1,21 +1,15 @@
...
     <View style={$demoItemContainer}>
       <Text preset="heading" style={$title}>
-        {item.name}
+        {name}
       </Text>
-      <Text style={$description}>{item.description}</Text>
-
-      {item.useCases.map((i) => (
-        <Fragment key={i.props.name}>{i}</Fragment>
-      ))}
```
yulolimum, line 18: If we're not wrapping usecases in DemoItem anymore, it might make sense to just move the name/description texts into the sectionListHeader and delete this component so we don't have to manage an extra component here.

boilerplate/app/screens/DemoComponentsScreen/DemoItem.tsx
```diff
@@ -25,9 +19,13 @@ const $demoItemContainer: ViewStyle = {
 }
 
 const $title: TextStyle = {
-  marginVertical: 8,
+  marginTop: 32,
+  paddingTop: 24,
```
yulolimum, line 23: I'm thinking part of the reason why you have to manage the negative scrollToLocation offset is because of these values. would it make sense to remove this and just add a bottom margin to the DemoItem wrapper?

PR #2022, opened 2022-08-18
Reactotron & RootStore initialization changes

jamonholmgren:
I refactored how Reactotron is initialized and how the RootStore is created and rehydrated. This follows the conventions I've been using for the past year.

I removed the Environment object as well. I don't really like that style of dependency injection, although if you have any feedback on that, happy to reflect on it.

The RootStoreProvider is still there, but only really useful for mocking the global RootStore for tests. I made notes to that effect.

Tested manually & works, as far as I can tell.

<table><tr>
<td>
image
</td>
<td>
image
</td>
<td>
image
</td>
</tr>
</table>

---

image


boilerplate/app/app.tsx
```diff
@@ -35,26 +47,32 @@ interface AppProps {
...
+  // Kick off initial async loading actions, like loading fonts and rehydrating RootStore
   useEffect(() => {
-    setupRootStore().then((rootStore) => {
-      setRootStore(rootStore)
+    ;(async () => {
+      // set up the RootStore (returns the state restored from AsyncStorage)
+      const { restoredState } = await setupRootStore(rootStore)
+
+      // reactotron integration with the MST root store (DEV only)
+      setReactotronRootStore(rootStore, restoredState)
```
jamonholmgren, line 68: I think this is a better place to integrate Reactotron and MST than in the setupRootStore function.

boilerplate/app/app.tsx
```diff
@@ -12,17 +12,29 @@
 import "./i18n"
 import "./utils/ignore-warnings"
 import { useFonts } from "expo-font"
-import React, { useEffect, useState } from "react"
+import React from "react"
 import { initialWindowMetrics, SafeAreaProvider } from "react-native-safe-area-context"
-import { RootStore, RootStoreProvider, setupRootStore } from "./models"
+import { useInitialRootStore } from "./models"
```
jamonholmgren, line 17: This is a new custom hook that moves the root store setup out of the app.tsx. I'm open to other naming ideas here. `useInitialRootStore` is okay, but I'm not sure if it really captures what's happening.

boilerplate/app/app.tsx
```diff
@@ -12,17 +12,29 @@
...
+  clearOnLoad: true,
+  // generally going to be localhost
+  host: "localhost",
+  // Reactotron can moniter AsyncStorage for you
+  useAsyncStorage: true,
+  // log the initial restored state from AsyncStorage
+  logInitialState: true,
+  // log out any snapshots as they happen (this is useful for debugging but slow)
+  logSnapshots: false,
+})
```
jamonholmgren, line 37: Expose the Reactotron config right here in app.tsx, making it easier to configure. Removed the previous note which applied to useScreens which is now gone.

boilerplate/app/app.tsx
```diff
@@ -44,39 +55,34 @@ function App(props: AppProps) {
...
-      // Note: (vanilla iOS) You might notice the splash-screen logo change size. This happens in debug/development mode. Try building the app for release.
-      setTimeout(hideSplashScreen, 500)
-    })
-  }, [])
+    // If your initialization scripts run very fast, it's good to show the splash screen for just a bit longer to prevent flicker.
+    // Slightly delaying splash screen hiding for better UX; can be customized or removed as needed,
+    // Note: (vanilla Android) The splash-screen will not appear if you launch your app via the terminal or Android Studio. Kill the app and launch it normally by tapping on the launcher icon. https://stackoverflow.com/a/69831106
+    // Note: (vanilla iOS) You might notice the splash-screen logo change size. This happens in debug/development mode. Try building the app for release.
+    setTimeout(hideSplashScreen, 500)
+  })
```
jamonholmgren, line 66: I like this pattern (as I mentioned above, the name could probably be improved). Open to feedback, however.

boilerplate/app/app.tsx
```diff
@@ -44,39 +55,34 @@ function App(props: AppProps) {
...
-    <RootStoreProvider value={rootStore}>
-      <SafeAreaProvider initialMetrics={initialWindowMetrics}>
-        <ErrorBoundary catchErrors="always">
-          <AppNavigator
-            initialState={initialNavigationState}
-            onStateChange={onNavigationStateChange}
-          />
-        </ErrorBoundary>
-      </SafeAreaProvider>
-    </RootStoreProvider>
```
jamonholmgren, line 79: No RootStoreProvider necessary in here anymore, as the initialized RootStore is now provided to the context by default.

boilerplate/app/models/helpers/setup-root-store.ts
```diff
@@ -1,55 +1,51 @@
...
- *
- * The environment includes other functions that will be picked from some
- * of the models that get created later. This is how we loosly couple things
- * like events between models.
- */
-export async function createEnvironment() {
-  const env = new Environment()
-  await env.setup()
-  return env
-}
```
jamonholmgren, line 22: Killed the Environment with fire. Which seems ... appropriate, somehow.

boilerplate/app/models/helpers/setup-root-store.ts
```diff
@@ -1,55 +1,51 @@
...
 
-  // prepare the environment that will be associated with the RootStore.
-  const env = await createEnvironment()
   try {
-    // load data from storage
-    data = (await storage.load(ROOT_STATE_STORAGE_KEY)) || {}
-    rootStore = RootStoreModel.create(data, env)
+    // load the last known state from AsyncStorage
+    restoredState = (await storage.load(ROOT_STATE_STORAGE_KEY)) || {}
+    applySnapshot(rootStore, restoredState)
```
jamonholmgren, line 31: Instead of creating a RootStoreModel, we just apply a snapshot to the existing one.

boilerplate/app/models/helpers/setup-root-store.ts
```diff
@@ -1,55 +1,51 @@
...
 
-  // reactotron logging
-  if (__DEV__) {
-    env.reactotron.setRootStore(rootStore, data)
-  }
+  // stop tracking state changes if we've already setup
+  if (_disposer) _disposer()
+
+  // track changes & save to AsyncStorage
+  _disposer = onSnapshot(rootStore, (snapshot) => storage.save(ROOT_STATE_STORAGE_KEY, snapshot))
```
jamonholmgren, line 43: We now have a way to unsubscribe from the onSnapshot listener.

boilerplate/app/models/helpers/use-stores.ts
```diff
@@ -0,0 +1,81 @@
...
+    })()
+
+    return () => {
+      // cleanup
+      if (_unsubscribe) _unsubscribe()
+    }
+  }, [])
+
+  return { rootStore, rehydrated }
+}
```
jamonholmgren, line 81: I like this better than the previous way (putting everything in setupRootStore), because it allows us to use hooks to let the App know when everything has been rehydrated.

boilerplate/app/navigators/navigation-utilities.ts
```diff
@@ -104,7 +104,9 @@ export function useNavigationPersistence(storage: any, persistenceKey: string) {
 
     if (previousRouteName !== currentRouteName) {
       // track screens.
-      __DEV__ && console.tron.log(currentRouteName)
+      if (__DEV__) {
+        console.tron.log(currentRouteName)
+      }
```
jamonholmgren, line 109: This form of `if (__DEV__) {` is better because Metro will literally strip the whole thing out on compilation (unlike the `__DEV__ && ...` method).

boilerplate/app/services/reactotron/reactotron-client.web.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * This file is loaded in web and exports the React.js version
+ * of Reactotron's client.
+ *
+ * React Native is loaded from reactotron-client.ts.
+ */
+import Reactotron from "reactotron-react-js"
+export { Reactotron }
```
jamonholmgren, line 8: These are just renamed from `tron.web.ts` and `tron.ts`, which I didn't think was very good.

boilerplate/app/services/reactotron/reactotron-config.ts
```diff
@@ -7,13 +7,10 @@ export interface ReactotronConfig {
...
-  state?: {
-    /** log the initial data that we put into the state on startup? */
-    initial?: boolean
-    /** log snapshot changes. */
-    snapshots?: boolean
-  }
+  /** log the initial data that we put into the state on startup? */
+  logInitialState?: boolean
+  /** log snapshot changes. */
+  logSnapshots?: boolean
```
jamonholmgren, line 13: I de-nested these, as it's easier to merge the config that way and there wasn't really a good reason to have them nested.

boilerplate/app/services/reactotron/reactotron.ts
```diff
@@ -1,203 +1,194 @@
-import { Tron } from "./tron"
```
jamonholmgren, line 1: It looks like this file has a ton of changes, but primarily it was just de-"environment"-alizing it and removing the class structure. Instead, it's just some local variables and exported functions. Much simpler.

jamonholmgren:
I'll figure out the CI failures tomorrow

boilerplate/app/app.tsx
```diff
@@ -12,17 +12,29 @@
...
-// stack navigation, use `createNativeStackNavigator` in place of `createStackNavigator`:
-// https://github.com/kmagiera/react-native-screens#using-native-stack-navigator
+// Set up Reactotron, which is a free desktop app for inspecting and debugging
+// React Native apps. Learn more here: https://github.com/infinitered/reactotron
+setupReactotron({
+  // clear the Reactotron window when the app loads/reloads
+  clearOnLoad: true,
+  // generally going to be localhost
+  host: "localhost",
+  // Reactotron can moniter AsyncStorage for you
```
frankcalise, line 31: moniter => monitor

boilerplate/app/app.tsx
yulolimum: I think this is good overall. Before, we had a bit simpler way to control when the components below would show by delaying setting the rootStore before certain items in the callback would finish (async ones in some cases). 

For example, you could have functions that kickoff async permissions checks or auth checks, then set the rootStore, then hide the splash screen.

As an alternative, I would see myself adding another useEffect to react to `rehydrated` to run certain things before setting a new `useState` variable that controls the visibility of below elements.

boilerplate/app/models/helpers/use-stores.ts
```diff
@@ -0,0 +1,80 @@
...
+ * const { someStore, someOtherStore } = useStores()
+ */
+export const useStores = () => useContext(RootStoreContext)
+
+/**
+ * Used only in the app.tsx file, this hook sets up the RootStore
+ * and then rehydrates it. It connects everything with Reactotron
+ * and then lets the app know that everything is ready to go.
+ */
+export const useInitialRootStore = (callback: () => void) => {
```
yulolimum, line 51: callback should probably be typed as `() => Promise<void>` just in case.

yulolimum (approved):

boilerplate/app/app.tsx
jamonholmgren: Good point — in a way, there’s almost a whole “flow” that has to happen before we display the app.

Maybe we should expose an async function that really formalizes that whole flow. I’ll think about this a bit more today.

boilerplate/app/models/helpers/use-stores.ts
jamonholmgren: Good call

PR #2032, opened 2022-08-19
Fix: properly rotates drawer icon in RTL mode

frankcalise:
## Describe your PR
Fixes the drawer icon pointing the wrong direction to close the drawer in RTL mode

|    `Before`    |    `After`    |
| ------------------------------- | ------------------------------- |
|image|image|


mazenchami (approved):
good catch! 🎉 🚀

PR #2036, opened 2022-08-22
Fixes RNR screen initial load and first-run UX on login screen

robinheinze:

## Describe your PR

- Adds a default email / password to the auth store which allows the user to just click login when they first boot up their Ignited app. 
- Adds observer to the RNR episodes screen to fix the initial data load bug
- Fixes the ESlint no unused var rule to allow _ prefixing to ignore

### Visuals
Screen Shot 2022-08-22 at 11 29 42 AM

### DIFF
--- a/boilerplate/app/models/AuthenticationStore.ts
+++ b/boilerplate/app/models/AuthenticationStore.ts
@@ -4,8 +4,8 @@ export const AuthenticationStoreModel = types
   .model("AuthenticationStore")
   .props({
     authToken: types.maybe(types.string),
-    authEmail: types.optional(types.string, ""),
-    authPassword: types.optional(types.string, ""),
+    authEmail: types.optional(types.string, "ignite@infinite.red"),
+    authPassword: types.optional(types.string, "ign1teIsAwes0m3"),
   })
   .views((store) => ({
     get isAuthenticated() {
--- a/boilerplate/app/models/AuthenticationStore.ts
+++ b/boilerplate/app/models/AuthenticationStore.ts
@@ -4,8 +4,8 @@ export const AuthenticationStoreModel = types
   .model("AuthenticationStore")
   .props({
     authToken: types.maybe(types.string),
-    authEmail: types.optional(types.string, "ignite@infinite.red"),
-    authPassword: types.optional(types.string, "ign1teIsAwes0m3"),
+    authEmail: types.optional(types.string, ""),
+    authPassword: types.optional(types.string, ""),
   })
   .views((store) => ({
     get isAuthenticated() {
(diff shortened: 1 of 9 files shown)


frankcalise (approved):

boilerplate/app/screens/DemoPodcastListScreen.tsx
```diff
@@ -27,7 +30,7 @@ export function DemoPodcastListScreen(props: DemoTabScreenProps<"DemoPodcastList
   return (
     <Screen preset="fixed" safeAreaEdges={["top"]}>
       <FlatList<Episode>
-        data={episodeStore.episodes}
+        data={episodeStore.episodes.slice()}
```
jamonholmgren, line 33: @robinheinze Do we need `slice` here after all?

jamonholmgren (approved):
Not sure what the CI failure is for. I restarted it.

boilerplate/app/screens/DemoPodcastListScreen.tsx
robinheinze: No i don't think we do. Updated.

boilerplate/app/models/AuthenticationStore.ts
```diff
@@ -4,8 +4,8 @@ export const AuthenticationStoreModel = types
   .model("AuthenticationStore")
   .props({
     authToken: types.maybe(types.string),
-    authEmail: types.optional(types.string, ""),
-    authPassword: types.optional(types.string, ""),
+    authEmail: types.optional(types.string, "ignite@infinite.red"),
+    authPassword: types.optional(types.string, "ign1teIsAwes0m3"),
```
yulolimum, line 8: I'm almost wondering if we should move the default values to the LoginScreen's useLayoutEffect (or similar). The reason being - this implementation is great to demo the "default value" feature of MST, but it's not a realistic use-case for the email/pass. If moved to the LoginScreen, we could add some comments like "this is where you could fetch the user/pass from keychain" or whereever. 

Anyways, not a huge deal, just throwin out ideas for a clearer intent.

yulolimum (approved):

boilerplate/app/models/AuthenticationStore.ts
robinheinze: @yulolimum Hmm, yeah i agree. Let me try something like that.

robinheinze:
> Not sure what the CI failure is for. I restarted it.

Looks like it failed on `npm run lint`, not totally sure why as it succeeds when I run it

yulolimum:
> Looks like it failed on `npm run lint`, not totally sure why as it succeeds when I run it

Try running yarn lint/format from both root and boilerplate directories. Sometimes, one of the variations catches the error for me.

robinheinze:
> > Looks like it failed on `npm run lint`, not totally sure why as it succeeds when I run it
> 
> Try running yarn lint/format from both root and boilerplate directories. Sometimes, one of the variations catches the error for me.

Turns out I was running from ignite and not from boilerplate. There are some actual linter errors after my unused-vars config change, which makes sense. 

But I wonder if there's a way to have the actual output show up on CI

Screen Shot 2022-08-22 at 3 37 28 PM

Issue #2035, opened 2022-08-22
Use command to strip out demo code


joshuayoes:
**What's going on?**
Whenever users generate a boilerplate, the first thing that most users do is delete most of the demo code out to start their project.

Jamon had the idea of having a command that allows us to programmatically delete all the demo code out of an ignite project, using comments to signify the delete markers, like so:

```
// @demo remove-file

const foo = true // @demo remove

// @demo remove-next-line
import { DemoScreen } from "./demo/screen"

// @demo remove-block
stuff
// @demo end-remove-block
```

Then, npx ignite-cli remove-demo will scan through ./app and remove the appropriate lines.
We can have an e2e test for this which will then scan for code that should have been removed to see if it’s still there.
Removing whole folders might be more tricky. I think for now we just remove those manually in the remove-demo command.

joshuayoes:
Main command is finished.

I'm going to leave this issue open until Maverick is mostly stabilized, and we can mark up all the demo files with the new comments and make sure that it builds correctly

joshuayoes:
Resolved by https://github.com/infinitered/ignite/pull/2140

PR #2039, opened 2022-08-22
Add `remove-demo` command

joshuayoes:

## Describe your PR
Resolves https://github.com/infinitered/ignite/issues/2035


boilerplate/app/navigators/AppNavigator.tsx
```diff
@@ -36,7 +36,9 @@ import { navigationRef, useBackButtonHandler } from "./navigation-utilities"
 export type AppStackParamList = {
   Welcome: undefined
   Login: undefined
+  // @demo remove-block-start
   Demo: NavigatorScreenParams<DemoTabParamList>
+  // @demo remove-block-end
```
joshuayoes, line 41: Added these blocks comments to test the main `src/commands/remove-demo.ts` command. Not the best use, will need to spend some more time rethinking this.

We should either:
- remove references to `@demo remove-*` in the boilerplate
- do a more in-depth test of what we want to remove

boilerplate/app/screens/DemoComponentsScreen/demos/DemoAutoImage.tsx
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable react/jsx-key */
+// @demo remove-file
```
joshuayoes, line 2: See https://github.com/infinitered/ignite/pull/2039/files#r954277628

boilerplate/app/screens/DemoComponentsScreen/demos/DemoButton.tsx
```diff
@@ -7,7 +7,7 @@ import { Demo } from "../DemoComponentsScreen"
 import { DemoDivider } from "../DemoDivider"
 import { DemoUseCase } from "../DemoUseCase"
 
-const $iconStyle: ImageStyle = { width: 30, height: 30 }
```
joshuayoes, line 10: See https://github.com/infinitered/ignite/pull/2039/files#r954277628

boilerplate/app/screens/index.ts
```diff
@@ -1,6 +1,7 @@
 export * from "./WelcomeScreen"
 export * from "./LoginScreen"
 export * from "./DemoCommunityScreen"
+// @demo remove-next-line
```
joshuayoes, line 4: See https://github.com/infinitered/ignite/pull/2039/files#r954277628

src/commands/remove-demo.ts
```diff
@@ -0,0 +1,98 @@
...
+    const TARGET_DIR = parameters.first ?? CWD
+    p()
+    p(`Removing demo code from '${TARGET_DIR}'`)
+
+    const getAllFilePaths = createGetAllFilePaths(filesystem)
+    const paths = getAllFilePaths(TARGET_DIR)
+
+    // Go through every file path and handle the operation for each demo comment
+    const demoCommentResults = await Promise.allSettled(
+      paths.map(async (path) => {
```
joshuayoes, line 23: Decided to go with 
```ts
Promise.allSettled(
    paths.map(async (path) => {
        // ...
    })
)`
```
So that all of these will be processed in parallel instead of synchronously like `for of` or `.forEach`

Trade off is that we don't have a guarantee that they will be processed the same way each time, but I think that's a worth while trade off for a potential speed improvement

src/commands/remove-demo.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    // Handle the results of the demo comment operations
+    demoCommentResults
+      // Sort the results by the path in alphabetical order
+      .sort((a, b) => {
+        if (a.status === "fulfilled" && b.status === "fulfilled") {
+          return a.value.path.localeCompare(b.value.path)
+        }
+        return 0
+      })
```
joshuayoes, line 64: Since we don't have a guarantee that files will processed in the same order each time we run the comment, in order to have a consistent log output for our snapshot test, we are sorting the results by file path

src/commands/remove-demo.ts
```diff
@@ -0,0 +1,98 @@
...
+      files.forEach((file) => {
+        const filePath = `${dir}/${file}`
+        if (filesystem.isDirectory(filePath)) {
+          removeEmptyDirs(filePath)
+        }
+      })
+    }
+    removeEmptyDirs(TARGET_DIR)
+
+    p(`Done removing demo code from '${TARGET_DIR}'`)
```
joshuayoes, line 96: We discussed the idea of having a `@demo remove-folder` directive, but I think until it becomes a big issue, we can just add `// @demo remove-file` and then check to see if all the files are deleted in a folder

src/tools/demo.ts
```diff
@@ -0,0 +1,57 @@
+export enum CommentType {
+  REMOVE_FILE = `@demo remove-file`,
+  REMOVE_CURRENT_LINE = `@demo remove-current-line`,
+  REMOVE_NEXT_LINE = `@demo remove-next-line`,
+  REMOVE_BLOCK_START = `@demo remove-block-start`,
+  REMOVE_BLOCK_END = `@demo remove-block-end`,
```
joshuayoes, line 6: So in order to support `//`, `/* */`, and `{/* */}` jsx comments, I decided to just include the `@demo remove-*` string as the keywords.

This could potentially run into errors where we find the keywords as strings and remove those lines instead of comments. However, I can't think of a use case where we'd write these keywords and not have them in a comment.

The alternative is bringing in much more heavy tooling with using an AST to look for comment tokens. That may be worth it, but I think this simpler approach of find and replace is worth it for now

src/tools/demo.ts
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Take the file content as a string and remove any
+ * line of code with an `// @demo remove-current-line` comment
+ */
+function removeCurrentLine(contents: string, comment = CommentType.REMOVE_CURRENT_LINE): string {
+  const lines = contents.split("\n")
+  const result = lines.filter((line) => !line.includes(comment))
+  return result.join("\n")
+}
```
joshuayoes, line 17: I had a really nice experience with copilot writing these functions. I'd write the comment and the function name + types. Then I'd write the test. Then I'd go open up copilot suggestions and try different ones until it passed. I felt like I was just code reviewing my own code 😆

src/commands/remove-demo.ts
jamonholmgren: I like this solution, Josh.

src/tools/demo.ts
jamonholmgren: I'm pretty against going to an AST, just because the understandability goes wayyy down and you'll be maintaining that yourself until you die. This simpler (less powerful) approach is superior.

docs/Remove-Demo-Code.md
```diff
@@ -0,0 +1,68 @@
+# Remove Demo Code
+
+Whenever users use command line boilerplate tool like `ignite` or `create-react-app`, the first thing that most users do is delete all the demo code to start their project.
```
jamonholmgren, line 3:
```suggestion
Whenever users use command line boilerplate tool like `ignite` or `create-react-app`, the first thing that many users do is delete all the demo code to start their project.
```

docs/Remove-Demo-Code.md
```diff
@@ -0,0 +1,68 @@
...
+
+Instead of going through the files yourself, or using find and replace, run `npx ignite-cli remove-demo`.
+
+After the command, the Ignite boilerplate will now have the smallest amount of demo code possible, while having all of Ignite's set up still at your fingertips.
+
+## Contributors
```
jamonholmgren, line 9:
```suggestion
## How It Works
```

docs/Remove-Demo-Code.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+### `// @demo remove-current-line`
+
+Remove the current line from the source code.
+
+#### Example:
+
+```tsx
+const $style: ViewStyle = { padding: 10 } // @demo remove
```
jamonholmgren, line 32:
```suggestion
const $style: ViewStyle = { padding: 10 } // @demo remove-current-line
```

jamonholmgren: Does it make sense to make this `@demo remove-line` instead? Shorter but still understandable?

docs/Remove-Demo-Code.md
```diff
@@ -0,0 +1,68 @@
...
+
+#### Example:
+
+```tsx
+import { DemoScreen } from "./demo/screen"
+```
+
+### `// @demo remove-block-start && // @demo remove-block-end`
+
+Remove the entire next line from the source block
```
jamonholmgren, line 47:
```suggestion
Remove the entire block between these lines from the source code
```

src/commands/remove-demo.ts
jamonholmgren: If this causes issues, I'm okay if we change to synchronous. Speed isn't a _yuge_ deal here, since it's rare people will run this command very often.

jamonholmgren (approved):
This is amazing work, Josh!

Only question really is about `remove-current-line` vs `remove-line`, and a few minor docs tweaks suggestions.

src/commands/remove-demo.ts
joshuayoes: Thank you! :)

src/tools/demo.ts
joshuayoes: I'm in the same boat, I think for this use case find and replace is perfect

docs/Remove-Demo-Code.md
joshuayoes: The main reason I decided to go with `@demo remove-current-line` is because if we use `@demo remove`, then our logic for checking other comments like `@demo remove-next-line` vs `@demo remove` gets more complicated that `file.includes(comment)`. Essentially, each `@demo remove` keyword is a unique id for an operation

joshuayoes: Ah, I misread the diff. I'll update this to remove to `@demo remove-current-line`. It's a little more verbose, but I feel like it's a clearer distinction with `@demo remove-next-line`.

PR #2041, opened 2022-08-23
Update Header component documentation to reflect maverick changes

silasjmatson:


docs/Components-Header.md
```diff
@@ -45,12 +99,55 @@ The `leftIcon` is an optional prop that is used to set the icon for the left nav
...
+### `leftIconColor`
+
+The `leftIconColor` is an optional prop that is used to set the tint color of the left navigation icon.
+
+```tsx
+<Header
+  leftIcon="back"
+  onLeftPress={() => navigation.goBack()}
+  leftIconColor="white"
+/>
```
jamonholmgren, line 111:
```suggestion
<Header
  leftIcon="back"
  leftIconColor="white"
  onLeftPress={() => navigation.goBack()}
/>
```

jamonholmgren: Tiny nit, just swapping the order of these.

docs/Components-Header.md
```diff
@@ -61,34 +158,85 @@ The `onLeftPress` is an optional prop that is used to set the function to be cal
...
+
+The `onRightPress` is an optional prop that is used to set the function to be called when the right navigation button is pressed.
+
+```tsx
+<Header rightIcon="back" onRightPress={() => navigation.goBack()} />
+```
+
+### `safeAreaEdges`
+
+The `safeAreaEdges` optional prop can be used to override the default safe area edges. By default, the header will use the `top` safe area edge. If you want to use the `bottom` safe area edge, you can pass in `["bottom"]` to the `safeAreaEdges` prop.
```
jamonholmgren, line 230: I'd love more info here on when you'd want to use the safeAreaEdges. (Generally -- I think it's when you have things like the notch on the top, or that weird bar on the bottom, but not if the view doesn't land on the top or bottom.)

jamonholmgren (approved):
Looks great!

docs/Components-Header.md
```diff
@@ -45,12 +99,55 @@ The `leftIcon` is an optional prop that is used to set the icon for the left nav
...
+/>
+```
+
+### `LeftActionComponent`
+
+The `LeftActionComponent` is an optional prop that is used to set a custom component for the left navigation button. Overrides the `leftIcon`, `leftText`,  and `leftTx` props.
+
+```tsx
+<Header
+  LeftActionComponent={() => <Text>Back</Text>}
```
yulolimum, line 148: Unliked Button's `LeftAccessory`, the Header's Left/Right Action Components are of type `ReactElement` or more broadly, a `ReactNode`.  So:

```
<Header LeftActionComponent={<Text>Back</Text>} />
```

This is similar to FlatList's `ListHeaderComponent`. The reason why it's an element here is, well, it's simpler. We also, don't need to pass any props back to the parent (in Button's case, we passed back the component's pressed state in case users had to change sometime in the accessories based on  that).

docs/Components-Header.md
yulolimum: Yea, for the most part, Header will always be at the top; hence, the default. I don't know many usecases when it might need diff values. Maybe if there's a situation where you want to have a banner on top of the header - then you'd remove the safearea by doing `safeAreaEdges={[]}`.

docs/Components-Header.md
```diff
@@ -61,34 +158,85 @@ The `onLeftPress` is an optional prop that is used to set the function to be cal
...
+```tsx
+<Header safeAreaEdges={["bottom"]} />
+```
+
+### `SafeAreaViewProps`
+
+The `SafeAreaViewProps` optional prop can be used to pass props to the `SafeAreaView` component. This is useful if you need to override the default `SafeAreaView` behavior.
+
+```tsx
+<Header SafeAreaViewProps={{ forceInset: { top: "never" } }} />
```
yulolimum, line 241: The SafeAreaView from `react-native-safe-area-context` does not have a `forceInset` prop.

docs/Components-Header.md
jamonholmgren: Good point.

docs/Components-Header.md
```diff
@@ -61,34 +158,85 @@ The `onLeftPress` is an optional prop that is used to set the function to be cal
...
+```
+
+### `RightActionComponent`
+
+The `RightActionComponent` is an optional `ReactElement` prop that is used to set a custom component for the right navigation button. Overrides the `rightIcon`, `rightText`,  and `rightTx` props.
+
+```tsx
+<Header
+  RightActionComponent={<Text>Back</Text>}
+  onRightPress={() => navigation.goBack()}
```
yulolimum, line 216: One thing i forgot to mention. When passing in a Right/Left ActionComponent, it overrides the props listed above in addition to the on Right/Left Press props (since that wraps the text/icons only).

PR #2043, opened 2022-08-23
Fixing formatting issues

jamonholmgren:
Should fix the test failures we've had lately...hopefully...


boilerplate/app/utils/set-dev.js
```diff
@@ -0,0 +1,10 @@
+/* eslint-disable-file */
+
+/**
+ * If running in an Expo Snack, __DEV__ is undefined.
+ * We set to `false` in that case.
+ *
+ * This file is not supposed to be eslinted or type checked.
+ */
+
+if (__DEV__ === undefined) __DEV__ = false
```
jamonholmgren, line 10: @curlyCoder15 I put this in its own file, as a `.js` file, so that it's not stripped when the over-eager formatter runs.

PR #2045, opened 2022-08-23
Checkbox component add labelStyle, modify the default label color

lionet1224:
By default, using storybook to view checkbox component text will be white and invisible, so I changed the default color to `color.palette.black` to get a better presentation.

And I think adding labelStyle will make better use of the checkbox component.

image


jamonholmgren (approved):
Thanks @lionet1224! We're redoing components in v8 Maverick, but I'll release this as a point release for v7.

Issue #1810, opened 2021-10-15
Add useScrollToTop in the screen component


derekgreenberg:
Add `useScrollToTop` from React Navigation in the screen component's `ScreenWithScrolling` function.

Be sure to add a link to [this documentation](https://reactnavigation.org/docs/use-scroll-to-top/) in a code comment so developers can look up the options.

jamonholmgren:
Included in v8 Maverick

PR #2046, opened 2022-08-23
Implemented useScrollToTop

frankcalise:

## Describe your PR

Implements request from #1810 - Add #1810 useScrollToTop in Screen component. Pressing the active tab will scroll to the top of the content within the `<ScrollView />`.

scroll-to-top


boilerplate/app/components/Screen.tsx
```diff
@@ -159,12 +160,17 @@ function ScreenWithScrolling(props: ScreenProps) {
...
 
+  // Add native behavior of pressing the active tab to scroll to the top of the content
+  // More info at: https://reactnavigation.org/docs/use-scroll-to-top/
+  useScrollToTop(ref)
+
   return (
     <ScrollView
-      keyboardShouldPersistTaps={keyboardShouldPersistTaps}
-      scrollEnabled={scrollEnabled}
+      {...{ keyboardShouldPersistTaps, scrollEnabled, ref }}
```
frankcalise, line 173: I'm fine with reverting this back, but for a bunch of props that just pass the same name/var the shorthand can look less busy when reading through the render function.

yulolimum: I actually kinda like this. Will try using more.

boilerplate/app/components/Screen.tsx
```diff
@@ -159,12 +160,17 @@ function ScreenWithScrolling(props: ScreenProps) {
     style,
   } = props as ScrollScreenProps
 
+  const ref = useRef<ScrollView>()
+
   const { scrollEnabled, onContentSizeChange, onLayout } = useAutoPreset(props as AutoScreenProps)
 
+  // Add native behavior of pressing the active tab to scroll to the top of the content
+  // More info at: https://reactnavigation.org/docs/use-scroll-to-top/
+  useScrollToTop(ref)
```
yulolimum, line 169: Do you know, is this a no-op if the screen is inside a stack or a nav other than tab?

frankcalise: @yulolimum `react-navigation` early returns from the `useEffect` if there is no tab nav found, so no listener would be set up in that case.

image

See this file for more info:
https://github.com/react-navigation/react-navigation/blob/main/packages/native/src/useScrollToTop.tsx#L56

jamonholmgren (approved):

PR #2048, opened 2022-08-23
RNR Metadata And Favorites

robinheinze:

## Describe your PR

- Adds metadata to RNR episode cards and persistent like button. 

### Visuals

Screen Shot 2022-08-23 at 11 39 43 AM

https://user-images.githubusercontent.com/6894653/186272093-211a4e92-bb90-4f25-989e-66ed5bdd713b.mp4

### DIFF
--- a/boilerplate/app/models/Episode.ts
+++ b/boilerplate/app/models/Episode.ts
@@ -18,7 +18,7 @@ export const EpisodeModel = types
   .props({
     guid: types.identifier,
     title: "",
-    pubDate: "",
+    pubDate: "", // Ex: 2022-08-12 21:05:36
     link: "",
     author: "",
     thumbnail: "",
--- a/boilerplate/app/models/Episode.ts
+++ b/boilerplate/app/models/Episode.ts
@@ -26,8 +26,6 @@ export const EpisodeModel = types
     content: "",
     enclosure: types.frozen<Enclosure>(),
     categories: types.array(types.string),
-    // additional properties
-    favorite: false,
   })
   .actions(withSetPropAction)
   .views((episode) => ({
--- a/boilerplate/app/models/Episode.ts
+++ b/boilerplate/app/models/Episode.ts
@@ -1,5 +1,14 @@
 import { Instance, SnapshotIn, SnapshotOut, types } from "mobx-state-tree"
 import { withSetPropAction } from "./helpers/with-set-prop-action"
+import { format } from "date-fns"
+
+interface Enclosure {
+  link: string
+  type: string
+  length: number
+  duration: number
+  rating: { scheme: string; value: string }
+}
 
 /**
  * This represents an episode of React Native Radio.
@@ -15,16 +24,27 @@ export const EpisodeModel = types
     thumbnail: "",
     description: "",
     content: "",
-    enclosure: types.frozen(),
+    enclosure: types.frozen<Enclosure>(),
     categories: types.array(types.string),
-
     // additional properties
     favorite: false,
   })
   .actions(withSetPropAction)
   .views((episode) => ({
     get datePublished() {
-      return new Date(episode.pubDate)
+      const date = new Date(episode.pubDate)
+      return format(date, "MMM dd, yyyy")
+    },
+    get duration() {
+      const seconds = Number(episode.enclosure.duration)
+      const h = Math.floor(seconds / 3600)
+      const m = Math.floor((seconds % 3600) / 60)
+      const s = Math.floor((seconds % 3600) % 60)
+
+      const hDisplay = h > 0 ? `${h}:` : ""
+      const mDisplay = m > 0 ? `${m}:` : ""
+      const sDisplay = s > 0 ? s : ""
+      return hDisplay + mDisplay + sDisplay
     },
   }))
 
--- a/boilerplate/app/models/EpisodeStore.ts
+++ b/boilerplate/app/models/EpisodeStore.ts
@@ -1,12 +1,13 @@
-import { Instance, SnapshotOut, types } from "mobx-state-tree"
+import { clone, Instance, SnapshotOut, types } from "mobx-state-tree"
 import { api } from "../services/api"
-import { EpisodeModel } from "./Episode"
+import { Episode, EpisodeModel } from "./Episode"
 import { withSetPropAction } from "./helpers/with-set-prop-action"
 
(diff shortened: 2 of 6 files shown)


boilerplate/app/models/Episode.ts
```diff
@@ -18,7 +18,7 @@ export const EpisodeModel = types
   .props({
     guid: types.identifier,
     title: "",
-    pubDate: "",
+    pubDate: "", // Ex: 2022-08-12 21:05:36
```
jamonholmgren, line 21: Eventually, it would be a good idea to handle dates how we would on a client project.

boilerplate/app/models/Episode.ts
```diff
@@ -9,22 +18,31 @@ export const EpisodeModel = types
...
     link: "",
     author: "",
     thumbnail: "",
     description: "",
     content: "",
-    enclosure: types.frozen(),
+    enclosure: types.frozen<Enclosure>(),
```
jamonholmgren, line 27: Nice, I like this.

boilerplate/app/models/EpisodeStore.ts
```diff
@@ -1,12 +1,13 @@
...
 import { api } from "../services/api"
-import { EpisodeModel } from "./Episode"
+import { Episode, EpisodeModel } from "./Episode"
 import { withSetPropAction } from "./helpers/with-set-prop-action"
 
 export const EpisodeStoreModel = types
   .model("EpisodeStore")
   .props({
     episodes: types.array(EpisodeModel),
+    favorites: types.map(EpisodeModel),
```
jamonholmgren, line 10: This isn't bad, but I'm thinking we probably want this to be an array of references instead.

```suggestion
    favorites: types.array(types.reference(EpisodeModel)),
```

boilerplate/app/models/EpisodeStore.ts
```diff
@@ -18,6 +19,24 @@ export const EpisodeStoreModel = types
         console.tron.error(`Error fetching episodes: ${JSON.stringify(response)}`, [])
       }
     },
+    addFavorite(episode: Episode) {
+      store.favorites.set(episode.guid, clone(episode))
```
jamonholmgren, line 23: Rather than cloning the episode, I think we'd just want to push the GUID into the references array.

```suggestion
      store.favorites.push(episode.guid)
```

boilerplate/app/models/EpisodeStore.ts
```diff
@@ -18,6 +19,24 @@ export const EpisodeStoreModel = types
...
+      store.favorites.delete(episode.guid)
+    },
+    hasFavorite(episode: Episode) {
+      return store.favorites.has(episode.guid)
+    },
+  }))
+  .actions((store) => ({
+    toggleFavorite(episode: Episode) {
+      if (store.favorites.has(episode.guid)) {
+        store.removeFavorite(episode)
```
jamonholmgren, line 35: Changing to an array of references means this part would change somewhat. Probably `.find(ep => ep.guid === episode.guid)`

boilerplate/app/screens/DemoPodcastListScreen.tsx
```diff
@@ -31,6 +31,7 @@ export const DemoPodcastListScreen = observer(function DemoPodcastListScreen(
     <Screen preset="fixed" safeAreaEdges={["top"]}>
       <FlatList<Episode>
         data={episodeStore.episodes}
+        extraData={episodeStore.favorites.size}
```
jamonholmgren, line 34: This would be .length with the array of references.

boilerplate/app/models/Episode.ts
```diff
@@ -9,22 +18,31 @@ export const EpisodeModel = types
...
-
-    // additional properties
-    favorite: false,
   })
   .actions(withSetPropAction)
   .views((episode) => ({
     get datePublished() {
-      return new Date(episode.pubDate)
+      const date = new Date(episode.pubDate)
+      return format(date, "MMM dd, yyyy")
```
jamonholmgren, line 34: I don't know if we want to add `date-fns` just for this. Do you think it is widely used enough to validate including it in the base boilerplate?

jamonholmgren (changes requested):
I'm proposing using an array of references rather than a map type for the favorites. Happy to have a discussion about this though.

boilerplate/app/models/Episode.ts
jamonholmgren: Actually, this probably works fine, given the datePublished() view.

robinheinze: Yeah, this was just a comment to remind the dev what format the date comes in from the server

boilerplate/app/models/EpisodeStore.ts
robinheinze: `includes` should work

boilerplate/app/models/Episode.ts
robinheinze: I've used it on nearly ever client project

boilerplate/app/models/EpisodeStore.ts
robinheinze: For some reason, I didn't think references would work since we replace the whole list on refresh. But it seems to work just fine. Updated.

boilerplate/app/models/Episode.ts
robinheinze: I'm open to alternatives though

frankcalise: Moment is also a huge import. `date-fns` is good or `luxon` is another alternative (I think by an ex-moment team member that realized the pitfalls of it). 

Not sure if @mazenchami or @leonskim want to weigh in from an International support perspective as well.

jamonholmgren: Let's include date-fns ... thanks @robinheinze !

jamonholmgren (approved):

boilerplate/app/models/Episode.ts
mazenchami: great question @frankcalise. I know they both have internationalization, but not sure if Arabic or Korean are part of it already. Would be worth testing that part when setting this up

mazenchami: documentation for `date-fns/i18n`  [here](https://date-fns.org/v2.29.2/docs/I18n)

PR #2058, opened 2022-08-25
Snackify maverick

curlyCoder15:
Purpose of this PR is to add ability to view ignite projects in snack.
To create a snack from ignite project 

1. Delete ios, android, e2e, test, jest.config.js, metro.config.js, webpack.config.js
2. All files from ./app/services/reactotron need to be removed and a new index file to be added, with the following code
` export const setupReactotron = (arg1:any) => null
    export const setReactotronRootStore=(arg1:any, arg2:any)=>null"`
this is required as we won't need to do any changes to reactotron functions anywhere else in the code
3. Update some packages as expo snack suggestions
4. Remove packages -> @expo/webpack-config, reactotron-react-native, detox detox-expo-helpers, ts-jest, jest, metro-config, @rnx-kit/metro-config, reactotron-mst, reactotron-react-js, reactotron-core-client

How to test this?
1. create a new project
2. ignite snackify
3. It's ready to import on snack - The best way to do this is to publish a public project on github. This is a more efficient way of doing this instead of importing every file to snack.

185521206-5c1d17e9-9615-4e5a-8fd7-aaefb3705ab0
185521215-1769827e-0bce-49f7-8402-5a8c16383182


yulolimum:
Kind of low GAFO on this but presumably the end-users have the Ignite-CLI installed, should we not make this an Ignite command instead? It would allow us to ship bugfixes/tweaks/additions to this script easier since they'd just have to update the CLI vs having to update their boilerplate's script file manually. I guess the benefit of keeping this in boilerplate is that people could customize it to keep their app 'snack-friendly'. 

Also, if we keep this in the boilerplate, thought's on converting this to a node script instead of bash? More accessibly to JS devs that way.

Again, low GAFO on both points.

curlyCoder15:
@yulolimum you are right the idea of keeping it inside boilerplate is to have the ability for the user to keep the app 'snack friendly'
I will look into node scripts

boilerplate/bin/snackify.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+#
+# Setup Script
+#
+echo "This will snackify your project and will delete some important folders like ios and android"
```
robinheinze, line 5: If I were a user, I'd be scared off by this line and never want to run this command. Maybe we could add some documentation somewhere explaining why someone might want to run this script?

curlyCoder15: I agree and updated this to a nice question and less scary

curlyCoder15:
For now we'll have the bash script but I'm working on using ignite-cli in the future

boilerplate/bin/snackify.sh
joshuayoes: I think if we could make this echo command longer, and list all the files that will be deleted, I would be satisfied as an end user to decided whether I want to continue or not

Something like:
```
In order to snackify this project, we must delete:
- /ios
- /android
- /e2e
- /test
- /app/services/reactotron
- jest.config.js
- metro.config.js
- webpack.config.js

Are you sure you want to continue?
```

boilerplate/bin/snackify.sh
```diff
@@ -0,0 +1,22 @@
...
+read -p "Are you sure you want to continue? " -n 1 -r
```
joshuayoes, line 6: Didn't know you could prompt to read lines from bash, TIL

joshuayoes:
Would you mind adding some notes about how you import into snack after running the bash script? 

I'm running the script locally, and it works great. Now I'm trying to drag and drop the files into a snack, and having some issues getting it to build

boilerplate/bin/snackify.sh
curlyCoder15: That's a great idea

src/commands/snackify.ts
```diff
@@ -0,0 +1,85 @@
...
+      await packager.remove("jest")
+      await packager.remove("metro-config")
+      await packager.remove("@rnx-kit/metro-config")
+      await packager.remove("reactotron-mst")
+      await packager.remove("reactotron-react-js")
+      await packager.remove("reactotron-core-client")
+    }
+    info("All done, your app is ready to be imported into expo snack")
+    info(
+      "The best way to import this project into an expo snack is via publishing this project into a public repository on github and then adding your repo's URL to an expo snack",
```
joshuayoes, line 82: It would be nice to add a link to https://snack.expo.dev/ here

joshuayoes:
This is looking really good! 

I followed this workflow to test:
1. `code ignite` - open branch in VSCode
2. `yarn build && npm link` (links `npx ignite-cli` to local version)
3. `npx ignite-cli new PizzaApp`
4. `code ../PizzaApp` - open new app in VSCode
5. `npx ignite-cli snackify`
6. [Create a new repo on Github](https://github.com/new) - I made [pizza-app-snackify](https://github.com/joshuayoes/pizza-app-snackify)
8. `git remote add origin https://github.com/joshuayoes/pizza-app-snackify.git && git branch -M main && git push -u origin main` - connect new app to repo (I changed the default branch to master)
9. `git push -u origin snack-090722` - or whatever your local snack branch is
10. Got to https://snack.expo.dev/
11. Click "Import Git Repository"
12. Added Repository URL and branch name

Then I got this error:

Screen Shot 2022-09-07 at 3 13 34 PM

For some reason, Expo seems to really not like tree shakeable imports from `date-fns`. 

Screen Shot 2022-09-07 at 3 26 39 PM

They offer suggestions to add those imports as dependencies, and that seems to fix it!

Screen Shot 2022-09-07 at 3 26 12 PM

There also seems to be fatal crashes around i18n on Snack

https://user-images.githubusercontent.com/37849890/188994854-fd602dbd-425d-4663-b575-5f0e8353dbe3.mov

Here is my github repo: https://github.com/joshuayoes/pizza-app-snackify/tree/snack-090722
And my snack: https://snack.expo.dev/@joshyoes-ir/ad51ef?platform=web

I think there is some tinkering that needs to happen to make Expo Snack happy, but this is really dope! Great job so far Simran!

src/commands/snackify.ts
```diff
@@ -0,0 +1,85 @@
...
+      await packager.remove("reactotron-react-native")
+      await packager.remove("detox")
+      await packager.remove("detox-expo-helpers")
+      await packager.remove("ts-jest")
+      await packager.remove("jest")
+      await packager.remove("metro-config")
+      await packager.remove("@rnx-kit/metro-config")
+      await packager.remove("reactotron-mst")
+      await packager.remove("reactotron-react-js")
+      await packager.remove("reactotron-core-client")
```
joshuayoes, line 78: _TOL: this is out of scope for this PR._ But we should really change packager in `src/tools/packager.ts` to accept an array of arguments, and add multiple packagers in one pass

src/commands/snackify.ts
```diff
@@ -0,0 +1,85 @@
...
+      if (!branchName.includes("snack")) {
+        info("Creating a new branch for your snack")
+        system.run(`git checkout -b snack-$(date +'%m%d%y')`)
+      }
+    } else {
+      info(
+        "There are uncommitted changes in your current branch, please commit changes before running this",
+      )
+      return
+    }
```
joshuayoes, line 27: Might want to add a `system.which('git')` check to make sure that we have a git repo at all, since not adding git is a possible options with ignite

src/commands/snackify.ts
jamonholmgren: Should be able to just put them all in one of the commands, space-separated.

```
await packager.remove("@expo/webpack-config reactotron-react-native detox detox-expo-helpers") // etc
```

src/commands/snackify.ts
jamonholmgren: Probably check for existence of `.git` directory, rather than existence of the `git` command.

src/commands/snackify.ts
curlyCoder15: I'll check this

jamonholmgren: @curlyCoder15 Probably have a list of what should be added and what should be removed, and just join them.

```tsx
const addPackages = [
  "expo-constatns@~13.2.4",
  "expo-localization@~13.1.0",
  // ... etc
]

const removePackages = [
  "@expo/webpack-config",
  "reactotron-react-native",
  // .. etc
]

await packager.add(addPackages.join(" "))
await packager.add(removePackages.join(" "))
```

curlyCoder15: I tried a similar approach using a loop - but that didn't work. using `join` makes more sense

Issue #2056, opened 2022-08-25
New project walkthrough fails when project names differs from project dir


frankcalise:
### Describe the bug

During the `new` command walkthrough, I tried to name my project different from the directory it would be located in and saw the following error:

`node ignite/bin/ignite new KristieLengel`

```
Error: ENOENT: no such file or directory, chdir '/Users/fcalise/code' -> 'KristieLengel'
    at wrappedChdir (node:internal/bootstrap/switches/does_own_process_state:112:14)
    at process.chdir (node:internal/worker:98:5)
    at /Users/fcalise/code/ignite/src/commands/new.ts:341:13
    at step (/Users/fcalise/code/ignite/src/commands/new.ts:44:23)
    at Object.next (/Users/fcalise/code/ignite/src/commands/new.ts:25:53)
    at fulfilled (/Users/fcalise/code/ignite/src/commands/new.ts:16:58) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'chdir',
  path: '/Users/fcalise/code',
  dest: 'KristieLengel'
```

### Ignite version

Maverick 7a20616be0d5c33e79cec4894d3100d2d7111e49

### Additional info

<img width="712" alt="image" src="https://user-images.githubusercontent.com/374022/186558556-a7bc2e36-3337-400d-aa80-e1c5b6a72be9.png">

joshuayoes:
I'll take a look at this today

joshuayoes:
I've put together a PR to resolve this. If you could take a look at it @frankcalise and make sure that it works on your end, I would appreciate it

PR #2062, opened 2022-08-25
Fix `new` command target directory flag for non-default arguments

joshuayoes:

## Describe your PR
Resolves https://github.com/infinitered/ignite/issues/2056 by checking if directory exists or not before moving into target directory

### DIFF
--- a/src/commands/new.ts
+++ b/src/commands/new.ts
@@ -13,6 +13,7 @@ import {
   clearSpinners,
 } from "../tools/pretty"
 import type { ValidationsExports } from "../tools/validations"
+import { bool } from "../tools/flag"
 
 // CLI tool versions we support
 const deps: { [k: string]: string } = {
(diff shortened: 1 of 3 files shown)


src/tools/flag.ts
```diff
@@ -0,0 +1,20 @@
...
+export function bool<Value extends Parameters<typeof Boolean>[0]>(value: Value): boolean {
+  if (value === "true") {
+    return true
+  }
+  if (value === "false") {
+    return false
+  }
+
+  return Boolean<Value>(value)
+}
```
jamonholmgren, line 20: Let's simplify this, since `Boolean("true") === true`. Also, the type signature seems unnecessarily complex, unless I'm missing something important:

```suggestion
export function bool(value: string | boolean): boolean {
  if (value === "false") return false
  return Boolean<Value>(value)
}
```

src/commands/new.ts
```diff
@@ -126,6 +127,7 @@ export default {
 
     const yname = !!options.y || !!options.yes
     const useDefault = (option: unknown) => yname && option === undefined
+    const boolFlag = (option: unknown) => (option !== undefined ? bool(option) : undefined)
```
jamonholmgren, line 130: Actually, let's add this logic into your `bool` function and just call the exported function `boolFlag`. Check for `undefined` first with

```
if (value === undefined) return value
```

jamonholmgren (commented):
A couple tweaks, then we're good!

src/commands/new.ts
joshuayoes: Extracted in this commit: https://github.com/infinitered/ignite/pull/2062/commits/010b913d12df0738c6c9a2faacf502bea814981d

PR #2065, opened 2022-08-26
miscellaneous Maverick tweaks

yulolimum:
## Please verify the following:

- [x] `yarn ci:test` **jest** tests pass with new tests, if relevant
- [x] `README.md` has been updated with your changes, if relevant

## Describe your PR

- There was a recent change made to TextField that made it inconsistent with the design. It somewhat broke right/left accessory default alignment as well. Adds tweaks to incorporate the change but make it inline with design.
- Adds setting to hide TabBar when keyboard opens in DemoNavigator.
- The RNR screen was causing crash. Tracked it to the passing `NaN` to the date-fns parsing function. Changes to parse using date-fns and wraps in try..catch.
- Adds podcast icon to RNR screen.

## Proofs

| RNR Screen Icon | TextField Tweaks |
|---|---|
| yulolimum-capture-2022-08-25--22-08-09@2x | yulolimum-capture-2022-08-25--22-08-44@2x |
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |

PR #2067, opened 2022-08-26
Maverick Styling and Date to account for i18n (and RTL)

mazenchami:

## Describe your PR

This PR address some changes to the work @yulolimum did in another [PR](https://github.com/infinitered/ignite/pull/2065). The two main aspects are how we're stying left/right/top/bottom for margins and padding and updating our dates formatter to take locale into account.

### Stying

Since we're supporting RTL, we should move away from using `{margin/padding}Left` and `{margin/padding}Right`. 

For LTR: start = left, end = right
For RTL: start = right, end = left

In [Android 4.2's support for RTL layouts](http://developer.android.com/about/versions/android-4.2.html#RTL) they give an example of how to use this flow following the above mentioned.

### Date

Added a new `/app/utils/format-date.ts` util file that makes our `date-fns` format function locale specific. 

NOTE: there is a known bug where this package doesn't localize numbers. Here is some activity about it on [StackOverflow](https://stackoverflow.com/questions/64869932/date-fns-format-is-not-localizing-numbers)

### DIFF
--- a/boilerplate/app/utils/format-date.ts
+++ b/boilerplate/app/utils/format-date.ts
@@ -1,0 +1,28 @@
+import { Locale, format, parseISO } from "date-fns"
+import I18n from "i18n-js"
+
+import ar from "date-fns/locale/ar-SA"
+import ko from "date-fns/locale/ko"
+import en from "date-fns/locale/en-US"
+
+interface Options {
+  locale?: Locale
+  weekStartsOn?: 0 | 1 | 2 | 3 | 4 | 5 | 6
+  firstWeekContainsDate?: number
+  useAdditionalWeekYearTokens?: boolean
+  useAdditionalDayOfYearTokens?: boolean
+}
+
+const getLocale = (): Locale => {
+  const locale = I18n.currentLocale().split("-")[0]
+  return locale === "ar" ? ar : locale === "ko" ? ko : en
+}
+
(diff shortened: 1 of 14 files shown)


boilerplate/app/utils/format-date.ts
```diff
@@ -0,0 +1,28 @@
+import { Locale, format, parseISO } from "date-fns"
+import I18n from "i18n-js"
+
+import ar from "date-fns/locale/ar-SA"
+import ko from "date-fns/locale/ko"
+import en from "date-fns/locale/en-US"
+
+interface Options {
```
robinheinze, line 8: Does date-fns not export this type?

robinheinze (approved):
The new spacing scale is merged, so you might have some conflicts, but LGTM!

boilerplate/app/utils/format-date.ts
mazenchami: The base of this PR is Yulian's so the conflicts might come in there.

I wish! they only export 4 types: 

Screen Shot 2022-08-26 at 2 13 21 PM

yulolimum: @mazenchami 

```
Parameters<typeof format>[2] 
```
will get you the type for the `options` arg

mazenchami: ummmm TIL!!!

PR #2072, opened 2022-08-26
add glob constraints to remove-demo command

joshuayoes:

## Describe your PR

- Make the script target directories/files that could potentially have demo code. Generated files or node_modules should be excluded (node_modules already were, we are just expanding that).
- Resolves snapshot tests.


src/commands/remove-demo.ts
```diff
@@ -1,9 +1,25 @@
...
+  "!**/android/build",
+  "!**/android/app/build",
+  "!**/.DS_Store",
+  "!**/.expo/**",
+  "!**/node_modules/**",
+  "!**/ios/build/**",
+  "!**/ios/Pods/**",
+  "!**/android/build/**",
+  "!**/android/app/build/**",
+]
```
yulolimum, line 21: @joshuayoes these are the glob matchers we're going to target. fs-jetpack uses this: https://github.com/isaacs/node-glob

src/commands/remove-demo.ts
```diff
@@ -15,14 +31,19 @@ module.exports = {
...
-    )
+    const filePaths = filesystem
+      .cwd(TARGET_DIR)
+      .find({
+        matching: MATCHING_GLOBS,
+        recursive: true,
+        files: true,
+        directories: false,
+      })
+      .map((path) => pathlib.join(TARGET_DIR, path))
```
yulolimum, line 42: I removed the need for the `path.ts` helper since we can achieve what we need through this. this should find all files that exclude the paths defined above using the globs array

src/commands/remove-demo.ts
```diff
@@ -78,22 +99,21 @@ module.exports = {
...
+      .find({
+        matching: MATCHING_GLOBS,
+        recursive: true,
+        files: false,
+        directories: true,
       })
-    }
-    removeEmptyDirs(TARGET_DIR)
+      .map((path) => pathlib.join(TARGET_DIR, path))
+      .filter((path) => !filesystem.list(path)?.length)
```
yulolimum, line 111: we can do the same here. we don't want to delete all empty dirs, only those that match

joshuayoes:
This looks good to me! Very elegant solution. 

I've pulled it down and run the test suite locally, with artifacts like node_modules in the boilerplate. I've also added some demo comments locally to the boilerplate, and the remove demo command picks them up

Issue #2068, opened 2022-08-26
Updating spacing scale in demo screens per new update 


kateinkim:
### What's going on?

Per PR https://github.com/infinitered/ignite/pull/2061, the spacing scale names have been updated. Planning on updating demo screens accordingly.

PR #2073, opened 2022-08-26
chore(boilerplate): update to new spacing scale with named properties

kateinkim:

## Describe your PR

Per issue #2068  All the margins and padding that were set by numbers before are now updated to the new spacing scale with named properties.


kateinkim:
I'm wondering if we should update the docs with paddings and margins to use `spacing` instead of numbers as well. Any thoughts? @robinheinze 

CleanShot 2022-08-29 at 09 14 31@2x
.

mazenchami (approved):
LGTM! 🚀

boilerplate/app/components/checkbox/checkbox.tsx
```diff
@@ -52,3 +26,29 @@ export function Checkbox(props: CheckboxProps) {
...
+  marginTop: 2, // finicky and will depend on font/line-height/baseline/weather
+  justifyContent: "center",
+  alignItems: "center",
+  borderWidth: 1,
+  borderColor: colors.border,
+  borderRadius: 1,
+}
+
+const $fill: ViewStyle = {
+  width: $dimensions.width - 4,
```
robinheinze, line 49: I'm guessing we want these 4's to be `spacing.tiny`

robinheinze: Fixed width / height wouldn't normally use the spacing scale, but when you're doing it as an offset from screen width or other fixed dimensions, that can be considered spacing since it's usually done to create padding around something

robinheinze (approved):
Just one comment! Looks good 👍

Issue #2075, opened 2022-08-29
When spinning up a new app `DemoPodcastListScreen` doesn't render list right away


kateinkim:
### Describe the bug

When spinning up a new app `DemoPodcastListScreen` is empty unless:
1. You tab to other screens and come back
2. Reload since waiting a few seconds for the list to render didn't work

|Issue|Swith to different screen|Wait and reload|
|-|-|-|
|<img width="334" alt="CleanShot 2022-08-29 at 09 48 06@2x" src="https://user-images.githubusercontent.com/53795920/187241931-98eda634-a9a4-4026-9170-b3d6349840ea.png">|![CleanShot 2022-08-29 at 10 15 27](https://user-images.githubusercontent.com/53795920/187247016-8f36ad6c-4834-47b6-b1cc-a0f049f5e5b0.gif)|![CleanShot 2022-08-29 at 10 23 49](https://user-images.githubusercontent.com/53795920/187248773-2f418c3b-4d98-4bcb-8482-43b56c42e403.gif)|

### Ignite version

maverick

### Additional info

<img width="723" alt="CleanShot 2022-08-29 at 09 56 19@2x" src="https://user-images.githubusercontent.com/53795920/187244095-48c62d6f-43b5-4a3a-8634-da2767f603e3.png">

kateinkim:
Also toggling `Switch` would work. Seems like just need a little trigger?

![CleanShot 2022-08-29 at 10 37 58](https://user-images.githubusercontent.com/53795920/187251285-fa060907-1bc7-462d-aa07-3bf44888e279.gif)

PR #2077, opened 2022-08-29
fix(boilerplate): Demo podcast list screen initial render

kateinkim:

## Describe your PR

Found an issue when spinning up a new app, `DemoPodcastListScreen` is empty: #2075 
I think the favorites feature was related. The length of favorites doesn't change on initial load so extra data doesn't change and the flatList never re-renders. But I still don't understand why the lists would show up on the reload.

### Photo/Video

|Fixed|Favorite List persists|
|-|-|
|CleanShot 2022-08-29 at 11 08 27|CleanShot 2022-08-29 at 11 09 38|


kateinkim:
Found another issue after I changed the code where the liked button would not update when clicked.

|Issue I created|Fixed|
|-|-|
|CleanShot 2022-08-29 at 12 44 19|CleanShot 2022-08-29 at 12 45 29|

Issue #2074, opened 2022-08-27
New command walkthrough errors when opting out of deps install


frankcalise:
### Describe the bug

When answering `n` to `Want us to install dependencies for you? (adds 50-100 seconds) (Y/n)`, the command errors out:

```
'yarn run v1.22.19\n' +
  '$ prettier --write "app/**/*.{js,jsx,json,md,ts,tsx}" --silent\n' +
  '/bin/sh: prettier: command not found\n' +
  'error Command failed with exit code 127.\n' +
  'info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.\n'
```

Also, when passing `y` to this, deps install but warnings get displayed from the yarn output and the command to copy in the future gets committed due to large output I'm thinking.

### Ignite version

maverick e955f53af8b6bda2573a363bacaa4f5f644d62cc

### Additional info

<img width="715" alt="image" src="https://user-images.githubusercontent.com/374022/187028490-c1270774-b224-4b98-9f25-606c7c2bcb49.png">

joshuayoes:
Huh, the prettier must be getting run even if it doesn't get installed locally. I'll take a look at this today

frankcalise:
```
node ignite/bin/ignite doctor
System
  platform           darwin
  arch               arm64
  cpu                10 cores     Apple M1 Pro
  directory          code         /Users/fcalise/code

JavaScript (and globally-installed packages)
  node                16.14.2      /Users/fcalise/.nvm/versions/node/v16.14.2/bin/node
  npm                 8.5.0        /Users/fcalise/.nvm/versions/node/v16.14.2/bin/npm
    corepack          0.10.0
    eas-cli           0.60.0
    expo-cli          6.0.2
    gatsby-cli        4.21.0
    npm               8.5.0
    yarn              1.22.19
  yarn                1.22.19      /Users/fcalise/.yarn/bin/yarn
    create-expo-app   1.1.1
  pnpm                -            not installed
  expo                -            not installed

Ignite
  ignite-cli         7.15.0       null
  ignite src         src          /Users/fcalise/code/ignite/src

Android
  java               11.0.14.1    /Users/fcalise/.asdf/shims/java
  android home       -            /Users/fcalise/Library/Android/sdk

iOS
  xcode              13.4.1
  cocoapods          1.11.3       /opt/homebrew/bin/pod
```

joshuayoes:
> Just tried the fix/new-install-deps branch and it completed fine when opting not to install them now
[...]

PR #2080, opened 2022-08-29
Speed up `new` command by caching dependencies locally

joshuayoes:

### Problem
`ignite-cli new` with the `--install-deps` option is pretty slow. This makes it hard to test because we need to wait 1-2 minutes between each run to test changes. Frank has filed a few regressions ([1](https://github.com/infinitered/ignite/issues/2056), [2](https://github.com/infinitered/ignite/issues/2074)) after we have [added flags](https://github.com/infinitered/ignite/pull/1999). It would be nice to improve the install speed so iterating on changes in manual QA or in tests would be faster.

#### Cache dependencies
We can take the turbo repo approach where we copy all the folders and files after an install to a local file system cache, and then copy them to the boilerplate target on subsequent installs


src/commands/new.ts
```diff
@@ -387,13 +388,71 @@ export default {
...
+    // check if there is a cache of node_modules using a hash of the package.json
+    function hash(str: string) {
+      return crypto.createHash("md5").update(str).digest("hex")
+    }
+
+    // check if there is a cache of node_modules using a hash of the package.json
+    const packageJsonHash = hash(packageJsonRaw)
+
+    // store hash at `~/.ignite/cache/${packageJsonHash}`
+    const cachePath = path(filesystem.homedir(), ".ignite", "cache", packageJsonHash)
```
jamonholmgren, line 400: If it's directly in the home directory, it'll be backed up by macOS, so we'll want to put it somewhere like `~/Library/Caches/.ignite` instead. I'm not sure if there's a Node API for getting the recommended operating system cache directory?

(See also: https://github.com/yarnpkg/yarn/issues/1637)

src/commands/new.ts
```diff
@@ -387,13 +388,71 @@ export default {
...
+
     if (installDeps === true) {
       const unboxingMessage = `Unboxing ${packagerName} dependencies`
       startSpinner(unboxingMessage)
       await packager.install({ ...packagerOptions, onProgress: log })
       stopSpinner(unboxingMessage, "🧶")
     }
 
+    // if there is no cache, create one
+    if (!cacheExists) {
```
jamonholmgren, line 444: We'll want a flag to disallow this.

jamonholmgren: Other ideas:

```
ignite-cli deps-cache # help
ignite-cli deps-cache path # returns location of cache
ignite-cli deps-cache clear
```

jamonholmgren (commented):
I like the direction you're going! Some comments.

src/commands/new.ts
jamonholmgren: I would probably not worry about trying to cache this on Windows, btw.

joshuayoes: Yeah windows has some weird stuff with anti-virus stuff that makes moving lots of files slow, so I can add a check to only allow this on mac/linux

joshuayoes: That is a good catch, I'll do some research on the most light weight cache location

src/commands/new.ts
joshuayoes: I think `--deps-cache=false` would be a good option to add here. 

Logging the location of the cache path when `--debug` is enabled would be another good way to help users figure out what is going on

src/commands/new.ts
joshuayoes: Looks like [linux uses `/home/username/.cache`](https://medium.com/linuxforeveryone/clear-program-cache-files-on-ubuntu-20-10-eac6a05c29a2) as a convention

joshuayoes: Seems like `~/Library/Caches/` is the convention on MacOS, I'll update that

joshuayoes: Added a `cachePaths` constant for this: https://github.com/infinitered/ignite/pull/2080/files#diff-0092032ec4a204634552d43c12ee8d637a7d97ccb1533d5973a89646d10c0822R14-R18

src/tools/cache.ts
```diff
@@ -0,0 +1,88 @@
...
+ * Function to create cache tool
+ *
+ * @example
+ * export default {
+ *   run: async (toolbox: GluegunToolbox) => {
+ *      const cache = createCacheTool(toolbox.filesystem)
+ *   }
+ * }
+ */
+export function createCacheTool(filesystem: GluegunFilesystem) {
```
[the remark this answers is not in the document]
jamonholmgren: I could be missing something, so please let me know if I am.

jamonholmgren (commented):
Just one small comment!

src/tools/cache.ts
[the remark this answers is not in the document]
joshuayoes: I think you are right about this, but it's a default habit in my brain rn

The builder is nice for doing unit tests because you can pass in a stub of tools like filesystem without having to use jest mocking, but I don't think that use case will be necessary here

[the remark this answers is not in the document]
joshuayoes: Especially since most of the methods on here are pure functions

joshuayoes:
New `ignite cache` command is done 👀 

ignite cache help command output

joshuayoes:
On my local machine, `ignite-cli new` is now 5x times faster with a warm cache 🐎 

**Warm Cache: ~10s**

Logo of ignite-cli new with warm cache

**No Cache: ~55s**

Screen Shot 2022-09-01 at 3 35 13 PM

joshuayoes:
I've recorded a video talking about the PR, why I decided to implement it, some of my reasoning: https://www.loom.com/share/abe0dd5852df42a7b1fba1c5bf848fec

joshuayoes:
Now you can go from `ignite new` to `yarn expo:start` in a minute: https://www.loom.com/share/edaca7d62eae4e5280f57e885c44a01c

cdanwards:
Took it for a spin and I'm getting an error on the second app (First using the cache) that I try to spin up:
Screen Shot 2022-09-02 at 4 21 25 PM

joshuayoes:
> Took it for a spin and I'm getting an error on the second app (First using the cache) that I try to spin up: Screen Shot 2022-09-02 at 4 21 25 PM

Thank you for catching this 🙏 I will re-create it locally and push up a fish

joshuayoes:
Screen Shot 2022-09-06 at 10 19 43 AM

@cdanwards would you mind verifying you did that with the latest commit on origin feat/new-perf? I'm on `e612ae01448bc62b907bc06fbd1751e4d59e3d97` and I am not having that issue again.

I realized a few commits ago, overwrite did not work as expected for non-default target paths, so that may be what's happening here.

cdanwards:
@joshuayoes checked it out again and it's working with no error now. Super fast too! Good job

jamonholmgren (approved):

joshuayoes (commented):
We are ready for speed 🐎

mazenchami:
@joshuayoes: [I have the need...the need for speed]

PR #2082, opened 2022-08-30
2055 accessibility

robinheinze:

- [ ] `README.md` has been updated with your changes, if relevant (Will do this in another ticket)

## Describe your PR

- Adds an animation to the like button, and a long-press action (necessary to keep the like action accessible since it's a nested touchable)
- Adds accessibility actions, labels and hints to the episode card
- Adds basic accessibility props to Button, Icon, and TextField. 

## Visuals

like-animate

https://user-images.githubusercontent.com/6894653/187343748-ee9ff65f-a1d7-4f32-9bf2-ed7a0a7754f4.mov

https://user-images.githubusercontent.com/6894653/187343585-3a027446-711d-428a-8926-eaaff41b9218.mov


boilerplate/app/components/switch/switch.tsx
```diff
@@ -47,8 +47,16 @@ const THUMB: ViewStyle = {
 
 const makeAnimatedValue = (switchOn) => new Animated.Value(switchOn ? 1 : 0)
 
-export function Switch(props: SwitchProps) {
```
robinheinze, line 50: @yulolimum I know you're re-doing Switch, so you'll have to make sure this works in the new version. Basically i just wanted to able to pass any additional accessibility props to Switch, so i extended the props type to included all ViewProps, and then did named deconstruction so i could pass a `...rest` into the main wrapper view.

robinheinze:
Not sure what's going on with CI.

cdanwards (approved):
I tested it out with VoiceOver on iOS and it worked great! The animation felt real smooth too.

kateinkim (approved):

boilerplate/app/components/Icon.tsx
```diff
@@ -60,12 +60,17 @@ export function Icon(props: IconProps) {
...
+  const isPressable = !!WrapperProps.onPress
   const Wrapper: ComponentType<TouchableOpacityProps> = WrapperProps?.onPress
     ? TouchableOpacity
     : View
 
   return (
-    <Wrapper {...WrapperProps} style={$containerStyleOverride}>
+    <Wrapper
+      {...WrapperProps}
+      accessibilityRole={isPressable ? "imagebutton" : "image"}
```
lindboe, line 71: I think there are a good amount of times where non-pressable icons or images are better off "invisible" to screen readers -- it depends whether they're decorative or conveying important meaning (e.g., ✔️ vs ❌ ). So I'm not sure this is something we want by default? If it is, at least `{...WrapperProps}` should probably come after these props so they can be overridden (that's probably generally true, not just for a11y)

boilerplate/app/screens/DemoPodcastListScreen.tsx
```diff
@@ -77,28 +91,126 @@ const EpisodeCard = observer(function EpisodeCard({
...
+          <Animated.View style={[$iconContainer, animatedUnlikeButtonStyles]}>
+            <Icon
+              icon="heart"
+              size={ICON_SIZE}
+              color={colors.palette.primary400} // pink
+              onPress={handlePressFavorite}
+              accessibilityLabel={
+                isFavorite
+                  ? translate("demoPodcastListScreen.accessibility.unfavoriteIcon")
+                  : undefined
```
lindboe, line 196: undefined if not `isFavorite`? why can't we just always use the one translation here, since they're different components when favorite/not favorite?

boilerplate/app/models/Episode.ts
```diff
@@ -31,9 +31,10 @@ export const EpisodeModel = types
   .views((episode) => ({
     get datePublished() {
       try {
-        return formatDate(episode.pubDate)
+        const formatted = formatDate(episode.pubDate)
+        return { textLabel: formatted, accessibilityLabel: `Published on ${formatted}` }
```
lindboe, line 35: nit: can remove "on", (e.g., "Published 2/12/2022")

lindboe: Should this string template be in i18n strings?

boilerplate/app/models/Episode.ts
```diff
@@ -45,7 +46,10 @@ export const EpisodeModel = types
       const hDisplay = h > 0 ? `${h}:` : ""
       const mDisplay = m > 0 ? `${m}:` : ""
       const sDisplay = s > 0 ? s : ""
-      return hDisplay + mDisplay + sDisplay
+      return {
+        textLabel: hDisplay + mDisplay + sDisplay,
+        accessibilityLabel: `Duration: ${h} hours ${m} minutes ${s} seconds`,
```
lindboe, line 51: Also shoudl be in i18n strings?

boilerplate/app/screens/DemoPodcastListScreen.tsx
```diff
@@ -77,28 +91,126 @@ const EpisodeCard = observer(function EpisodeCard({
...
+
   return (
+    //  MAVERICKTODO: Switch this out for Card component when that is ready
     <TouchableOpacity
       style={[$rowLayout, $item]}
-      onPress={() => openLinkInBrowser(episode.enclosure.link)}
+      onPress={handlePressCard}
+      onLongPress={handlePressFavorite}
+      // MAVERICKTODO: This button role should be set on the Card Component
+      accessibilityRole="button"
```
lindboe, line 166: Are we going with long press here because the favorite button is nested in a pressable?

This works, but I think a simpler design for screen readers that wouldn't affect visuals would involve _not_ nesting buttons. The majority of the card area could still be a button to listen, but the favorite icon button has its own separate pressable area. Thoughts?

lindboe (commented):
Did a quick pass and left some notes 👍

boilerplate/app/screens/DemoPodcastListScreen.tsx
robinheinze: The issue is they both exist on screen due to the way the animation works, one just gets set to 0 opacity. So I don't want it to read out the one that's not active.

boilerplate/app/models/Episode.ts
robinheinze: Ah good catch

boilerplate/app/screens/DemoPodcastListScreen.tsx
robinheinze: Yes, that's exactly the reason for it. I'm not sure how that would work. I know @yulolimum is working on a `<Card />` component, which I was planning to incorporate in this once it was finished, but I'm not sure how flexible the touchable area would be on that. Any thoughts @yulolimum?

yulolimum: >but I'm not sure how flexible the touchable area would be on that

Generally I have been trying to expose customizations to all sub-components via a general prop similar to this: https://github.com/infinitered/ignite/blob/maverick/boilerplate/app/components/ListItem.tsx#L51-L54 . Or, as a spread for the wrapper: https://github.com/infinitered/ignite/blob/maverick/boilerplate/app/components/ListItem.tsx#L14.

It'll be the same for Card.

mazenchami (commented):
just one comment, but will push updated translations a little later

boilerplate/app/screens/DemoPodcastListScreen.tsx
robinheinze: I'll do a second PR integrating Card and I'll see if I can implement @lindboe's suggestion.

PR #2083, opened 2022-08-30
fix(i18n): Fixes direction of caret on links for RTL

cdanwards:

## Describe your PR

For languages that are RTL, carets or arrow point in the direction that the new window opens. This PR fixes this.


mazenchami (approved):
Good catch even though you don't speak Arabic 😆 🚀

PR #2107, opened 2022-09-01
Toggle Component

yulolimum:

## Describe your PR

- Adds a `Toggle` component. This component covers all "boolean-type" inputs: checkbox, radio, and switch.
- Updates `DemoPodcastListScreen` with new Toggle component.
- Adds a Toggle entry to the Showcase.
- Removes `Checkbox` component (replaced by `Toggle`).
- Removes `BulletItem` component. This can easily be achieved with `Text`.
- Removes `Switch` component (replaced by `Toggle`).

## Proofs

| Toggle Component Showcase | Toggle Component Demo |
|---|---|
| yulolimum-capture-2022-09-01--07-44-34 | <video src="https://user-images.githubusercontent.com/1775841/187944023-3fc9145e-2001-4ad5-b363-ec68f4e72e67.mp4"/> |
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |


boilerplate/app/components/Toggle.tsx
```diff
@@ -0,0 +1,493 @@
...
+
+type Variants = "checkbox" | "switch" | "radio"
+
+export interface ToggleProps extends Omit<TouchableOpacityProps, "style"> {
+  /**
+   * The variant of the toggle.
+   * Options: "checkbox", "switch", "radio"
+   * Default: "checkbox"
+   */
+  variant?: Variants
```
yulolimum, line 28: The Toggle component has 3 variants which are all "boolean-types"

boilerplate/app/components/Toggle.tsx
```diff
@@ -0,0 +1,493 @@
...
+   */
+  editable?: TextInputProps["editable"]
+  /**
+   * The value of the field. If true the component will be turned on.
+   */
+  value?: boolean
+  /**
+   * Invoked with the new value when the value changes.
+   */
+  onValueChange?: SwitchProps["onValueChange"]
```
yulolimum, line 44: Changes from `onToggle` (that we had in legacy Switch) to use a callback prop more inline with the [official Switch from RN](https://reactnative.dev/docs/switch#onvaluechange).

boilerplate/app/components/Toggle.tsx
```diff
@@ -0,0 +1,493 @@
...
+   */
+  variant?: Variants
+  /**
+   * A style modifier for different input states.
+   */
+  status?: "error" | "disabled"
+  /**
+   * If false, input is not editable. The default value is true.
+   */
+  editable?: TextInputProps["editable"]
```
yulolimum, line 36: Adds an `editable` prop to be inline with `TextField`'s editable prop. This is an alternative way to specific `status="disabled"` and does the same thing.

boilerplate/app/components/Toggle.tsx
```diff
@@ -0,0 +1,493 @@
...
+    },
+    !disabled && { backgroundColor: colors.palette.neutral200 },
+    $inputOuterStyleOverride,
+  ]
+
+  function handlePress(e: GestureResponderEvent) {
+    if (disabled) return
+    onValueChange?.(!value)
+    onPress?.(e)
+  }
```
yulolimum, line 154: Since this is a "controlled" component (same as RN's `Switch`) pressing on it won't change it's value. A handler needs to change the `value` prop in order for the component to change state. 2 convenience methods are provided - one with the touch event and the other with the potential new value. Either can be used to determine the `next` state of the Toggle.

boilerplate/app/components/Toggle.tsx
```diff
@@ -0,0 +1,493 @@
...
+
+  return (
+    <Wrapper
+      activeOpacity={1}
+      accessibilityRole={variant}
```
yulolimum, line 159: We can pass the variant string directly in since those exist on this prop.

boilerplate/app/components/Toggle.tsx
```diff
@@ -0,0 +1,493 @@
...
+    status,
+    disabled,
+    innerStyle: $radioContainerStyleOverride,
+    detailStyle: $radioDotStyleOverride,
+  } = props
+
+  const $animatedRadioContainer = useAnimatedStyle(
+    () => ({ opacity: withTiming(on ? 1 : 0) }),
+    [on],
+  )
```
yulolimum, line 238: Radio is as simple as the checkbox. Simple fade does the job.

boilerplate/app/components/Toggle.tsx
```diff
@@ -0,0 +1,493 @@
...
+      }
+      if (disabled) return colors.transparent
+      if (status === "error") return colors.error
+      return colors.palette.secondary500
+    })()
+
+    const backgroundColor = withTiming(on ? onBackgroundColor : colors.transparent)
+
+    return { backgroundColor }
+  }, [status, disabled, on])
```
yulolimum, line 285: The animation for the switches track. Since we can't simply fade in/out the track (due to a nested knob), we animate the background color instead. This could actually be done with a fade, but we'd have to un-nest the knob. I will experiment with this. Interpolating colors isn't ideal in certain situations since there isn't any control over the tone/hue when transitioning - with some color customizations, you might see a weird color transition.

boilerplate/app/components/Toggle.tsx
```diff
@@ -0,0 +1,493 @@
...
+      if (disabled) return colors.palette.neutral600
+      return colors.palette.neutral100
+    })()
+
+    const backgroundColor = withTiming(on ? onBackgroundColor : offBackgroundColor)
+    const start = withTiming(on ? "100%" : "0%")
+    const marginStart = withTiming(on ? -renderedKnobWidth : 0)
+
+    return { backgroundColor, start, marginStart }
+  }, [status, disabled, on])
```
yulolimum, line 308: Animation for the knob. The position (`start`) could have been done simpler. But since we support customization (including sizes), we can't animate between static ranges (e.g. `0, 25`). Instead, we animate from 0 to 100 and offset the knob by it's rendered-size using `marginStart`. Theoretically, we can pull it's size from the defined styles, but since a user can customize with a string value, this becomes a bit more complex. Instead we just measure it and use its point value here.

boilerplate/app/components/Toggle.tsx
```diff
@@ -0,0 +1,493 @@
...
+
+      {switchAccessibilityMode === "icon" && shouldLabelBeVisible && (
+        <Image
+          style={[$switchAccessibilityIcon, { tintColor: color }]}
+          source={role === "off" ? iconRegistry.hidden : iconRegistry.view}
+        />
+      )}
+    </View>
+  )
+}
```
yulolimum, line 368: This component provides more visual markers for the state of the switch. In "text" mode, we output `| or ⭘`. In "icon" mode, we have a visible/hidden icons.

boilerplate/app/components/Toggle.tsx
```diff
@@ -0,0 +1,493 @@
...
+    <Text
+      preset="formLabel"
+      text={label}
+      tx={labelTx}
+      txOptions={labelTxOptions}
+      {...LabelTextProps}
+      style={$labelStyle}
+    />
+  )
+}
```
yulolimum, line 400: FIeldLabel is the component that renders text to the right or left of the Toggle input. It matches TextField's label style.

boilerplate/app/screens/DemoPodcastListScreen.tsx
```diff
@@ -61,12 +61,15 @@ export const DemoPodcastListScreen = observer(function DemoPodcastListScreen(
...
-                accessibilityLabel={translate("demoPodcastListScreen.accessibility.switch")}
+              <Toggle
                 value={episodeStore.favoritesOnly}
-                onToggle={() => episodeStore.setProp("favoritesOnly", !episodeStore.favoritesOnly)}
+                onValueChange={() =>
+                  episodeStore.setProp("favoritesOnly", !episodeStore.favoritesOnly)
+                }
+                variant="switch"
+                labelTx="demoPodcastListScreen.onlyFavorites"
+                accessibilityLabel={translate("demoPodcastListScreen.accessibility.switch")}
```
yulolimum, line 71: Updates to use Toggle on podcast screen.

boilerplate/app/components/Toggle.tsx
```diff
@@ -0,0 +1,493 @@
...
+    if (disabled) return
+    onValueChange?.(!value)
+    onPress?.(e)
+  }
+
+  return (
+    <Wrapper
+      activeOpacity={1}
+      accessibilityRole={variant}
+      accessibilityState={{ checked: value }}
```
robinheinze, line 160: I believe this should also contain the `disabled` state as well as the checked state.

robinheinze (approved):
Just one minor accessibility change. Otherwise this looks fantastic! Really thorough.

boilerplate/app/components/Toggle.tsx
mazenchami: @robinheinze would this be:

`accessibilityState={{ checked: value, disabled }}`

mazenchami (approved):
LGTM! 🎉 🚀

PR #2111, opened 2022-09-01
Remove deep nesting from generators

robinheinze:

## Describe your PR

By default, we are moving away from deeply nested structure, so this moves generated components, models, and screens out of their individual named folders and into the relevant root folder instead. 

If the user prefers it the old way, the templates can be updated using front-matter and the destinationDir property:
```
destinationDir: app/models/<%= props.kebabCaseName %>/<%= props.kebabCaseName %>\"
```

## Visuals

Screen Shot 2022-09-01 at 11 16 53 AM


kateinkim:
LGTM 👍🏼

kateinkim (approved):

cdanwards:
In the new file structure, the files are capitalized but the generators are creating them with lower-case naming: `Pizza` vs `pizza`. Are we wanting the generators to match the provided nomenclature?

robinheinze:
@cdanwards that's a great point. I think we should match the new pattern. I'll update this PR to do that as well.

robinheinze:
Actually, I think I will do that in a separate PR since it's a fairly substantial change and I don't want to hold this part up in case i don't finish before EOD.

PR #2123, opened 2022-09-06
Improve CI test job speed

joshuayoes:

### Problem
Currently, many test jobs on PRs are hitting the 5 minute limit in CircleCI, and failing. We need to speed up the test

### Solution
Break up longest running test, `test/vanilla/ignite-new.test.ts`, into a series of smaller tests that can be run in parallel


joshuayoes:
Currently, CI seems to be taking between 5:00-6:30 minutes for most Maverick PRs

Screen Shot 2022-09-06 at 12 38 31 PM

[CI job of most recent commit](https://app.circleci.com/pipelines/github/infinitered/ignite/1759/workflows/2c732f6f-e9fb-456a-bf04-5f1f06a80f58/jobs/2311) shaves about minute off the CI suite

Screen Shot 2022-09-06 at 12 37 32 PM

joshuayoes:
Last test is the longest out of all the chunks. Thinking of ways to break these up, since the generator file creation may mess with the linting/format/compile checks

Screen Shot 2022-09-06 at 1 00 40 PM

.circleci/config.yml
```diff
@@ -9,16 +9,11 @@ defaults: &defaults
     - image: cimg/node:14.19
   working_directory: /mnt/ramdisk/repo
 
-# mac: &mac
-#   macos:
-#     xcode: "13.2.1"
-#   working_directory: ~/repo
-
```
joshuayoes, line 16: Removed comments. They are in the git history, so we can get them back. Otherwise they create a lot of noise

.circleci/config.yml
```diff
@@ -9,16 +9,11 @@ defaults: &defaults
...
 version: 2.1
 jobs:
   tests:
     <<: *defaults
-    parallelism: 4
```
joshuayoes, line 21: So parallelism is useful for splitting the workload of several tests: https://circleci.com/docs/parallelism-faster-jobs

We aren't really using this feature as intended. We are just running 4 versions of the same job every time.

We don't have a lot of tests, we just have on that is slower. So using `resource_class: large` is a better move here because we want the most resources to run the one big test as fast as possible

.circleci/config.yml
```diff
@@ -45,83 +40,22 @@ jobs:
           command: |
             git config --global user.email "ci@infinite.red"
             git config --global user.name "Infinite Red"
-      - run:
-          name: Install PNPM for tests
-          command: |
-            sudo chmod 777 /usr/local/bin
-            curl -f https://get.pnpm.io/v6.16.js | node - add --global pnpm@7
-            pnpm config set store-dir ../.pnpm-store
```
joshuayoes, line 53: Didn't have a huge impact on job time, but we don't use pnpm for tests anymore

.circleci/config.yml
```diff
@@ -45,83 +40,22 @@ jobs:
...
       - restore_cache:
           name: Restore ignite dependency cache
           key: ignite-deps-cache-{{ .Environment.IGNITE_DEPS_PACKAGER }}-{{ checksum "boilerplate/package.json" }}-{{ arch }}-{{ .Environment.IGNITE_DEPS_KEY_SUFFIX }}
       - run:
-          name: Run tests
-          command: yarn ci:test
+          name: Run static tests
+          command: yarn format:check & yarn lint & yarn typecheck
```
joshuayoes, line 48: Using the single `&` means all 3 of these jobs will run in parallel. They are pretty fast, so I think it's not worth the cost of splitting this into another machine

.circleci/config.yml
```diff
@@ -45,83 +40,22 @@ jobs:
...
+      - run:
+          name: Run jest tests
+          command: yarn test
```
joshuayoes, line 51: Ran this seperately than the other tests because it is slower, and running multiple yarn scripts with `&` will combine all the std outputs to the same line, and it looks weird with jest test runner

.circleci/config.yml
```diff
@@ -45,83 +40,22 @@ jobs:
...
-  #         command: cd tempApp && ignite new DetoxTests --detox --skip-git -b ../
-  #     - run:
-  #         name: Build Detox in ignited app
-  #         command: cd tempApp/DetoxTests && yarn ci:build:e2e
-  #     - run:
-  #         name: Run detox tests in ignited app
-  #         command: cd tempApp/DetoxTests && yarn ci:test:e2e
-  #     - run:
-  #         name: Remove temp directory
-  #         command: rm -rf tempApp
```
joshuayoes, line 123: See https://github.com/infinitered/ignite/pull/2123/files#r964214546

.circleci/config.yml
```diff
@@ -133,6 +67,9 @@ jobs:
             - v1-dependencies-{{ checksum "package.json" }}-{{ arch }}
             # fallback to using the latest cache if no exact match is found
             - v1-dependencies-
+      - run:
+          name: Build package
+          command: yarn build
```
joshuayoes, line 72: Remove `yarn ci:prep` from tests since it's not necessary, but added it back here

.prettierignore
```diff
@@ -0,0 +1 @@
+/build
```
joshuayoes, line 1: Can't put this in package.json :( https://github.com/prettier/prettier/issues/3460

Decided it was easier to add a file than create a more complicated glob in `format` npm script

package.json
```diff
@@ -61,8 +56,8 @@
     "@types/jest": "^27.0.1",
     "@types/node": "16.6.1",
     "@types/sharp": "^0.30.2",
-    "@typescript-eslint/eslint-plugin": "^4.19.0",
-    "@typescript-eslint/parser": "^4.19.0",
+    "@typescript-eslint/eslint-plugin": "^5.36.2",
+    "@typescript-eslint/parser": "^5.36.2",
```
joshuayoes, line 60: Updated these because eslint was complaining about this package being too old for our typescript version

package.json
```diff
@@ -122,7 +117,10 @@
       "no-use-before-define": "off",
       "quotes": 0,
       "space-before-function-paren": 0
-    }
+    },
+    "ignorePatterns": [
+      "**/*.snap"
+    ]
```
joshuayoes, line 123: Updated prettier, and it started included snap files. We don't care since they are auto generated

test/_test-helpers.ts
```diff
@@ -31,266 +31,3 @@ export async function runError(cmd: string): Promise<string | any> {
...
-  // #endregion
-
-  // #region Assert package.json Scripts Can Be Run
-  // run the tests; if they fail, run will raise and this test will fail
-  await run(`npm run test`, runOpts)
-  await run(`npm run lint`, runOpts)
-  await run(`npm run compile`, runOpts)
-  expect(await run("git diff HEAD", runOpts)).toEqual("")
-  // #endregion
-}
```
joshuayoes, line 296: Since we don't share these, this logic can live in the actual tests

test/vanilla/ignite-new.test.ts
```diff
@@ -1,70 +1,349 @@
...
+
+      // we're done!
+    })
   })
 })
 
-const ignoreFolders = ["/xcuserdata", ".git", "node_modules", "Pods", "/build"]
-
 async function checkForLeftoverHelloWorld(filePath: string) {
+  const ignoreFolders = ["/xcuserdata", ".git", "node_modules", "Pods", "/build"]
```
joshuayoes, line 346: We don't share this anywhere else, I made it local to the function so it will fold in VSCode

test/vanilla/ignite-new.test.ts
```diff
@@ -1,70 +1,349 @@
...
+      result = await runIgnite(`new ${APP_NAME} --debug --packager=npm --yes`, {
+        pre: `cd ${tempDir}`,
+        post: `cd ${originalDir}`,
+      })
+      appPath = filesystem.path(tempDir, APP_NAME)
+    })
+
+    afterAll(() => {
+      filesystem.remove(tempDir) // clean up our mess
+    })
```
joshuayoes, line 42: Now this test suite will set everything else up once, then we can create tests to describe individual regions we are testing for. That way jest will report when one piece fails, but not fail the entire test block

Also it will run these tests in parallel, which offers a modest speed increase

package.json
```diff
@@ -24,21 +24,16 @@
   },
   "scripts": {
     "compile": "tsc -p .",
+    "typecheck": "tsc -p . --noEmit",
     "build": "yarn compile && cp ./boilerplate/.gitignore ./boilerplate/.gitignore.template",
-    "format": "yarn format:js && yarn format:json && yarn format:md && yarn format:ts;",
-    "format:js": "prettier --write '**/*.js'",
-    "format:json": "prettier --write '**/*.json'",
-    "format:md": "prettier --write '**/*.md'",
-    "format:ts": "prettier --write '**/*.ts{,x}'",
```
joshuayoes, line 32: These are all redundant, we can write these as one script with a glob pattern to speed this up instead of running it 4 times

package.json
```diff
@@ -24,21 +24,16 @@
...
-    "lint": "eslint src test --ext .ts --fix && yarn format",
```
joshuayoes, line 33: I would expect this based on the name just to run eslint, not also run a side effect of formatting files with prettier

package.json
```diff
@@ -24,21 +24,16 @@
...
+    "format:check": "yarn format --check",
+    "lint": "eslint 'src/**' 'test/**'",
     "test": "jest",
     "watch": "jest --watch",
     "watch:debug": "yarn watch --runInBand --verbose",
     "coverage": "jest --coverage",
-    "ci:prep": "yarn build && yarn ci:lint",
-    "ci:lint": "eslint src test --ext .ts && yarn format",
-    "ci:test:vanilla": "jest ./test/vanilla",
-    "ci:test": "yarn ci:prep && yarn test && yarn clean",
```
joshuayoes, line 41: These scripts aren't used anymore. Most of the other scripts can be run for other purposes than CI

joshuayoes:
Looking good! https://app.circleci.com/pipelines/github/infinitered/ignite/1764/workflows/59bcfd0f-6cdb-41d4-85a0-ba2e4d11282e/jobs/2318

Screen Shot 2022-09-06 at 3 18 18 PM

See my comments for the changes that made a difference

.circleci/config.yml
joshuayoes: I tried wrapping these in one `yarn static-test` script, but for some reason `yarn` will hang a lot even on success

joshuayoes: Update this to exit with a 1 exit code if any fail

jamonholmgren (approved):

PR #2127, opened 2022-09-08
Improves generator update output

jamonholmgren:
Previous output:

CleanShot 2022-09-08 at 09 29 28@2x

New output:

CleanShot 2022-09-08 at 09 29 33@2x


joshuayoes (approved):
This looks good! I think having validation warning before the rest of the output is the best pattern

PR #2132, opened 2022-09-09
Rename e2e to detox, update scripts

jamonholmgren:
1. Renames `e2e` to `detox`, which is more explicit.
2. Updates the package.json scripts to be more consistent:
 - `build-ios` is now `bundle:ios`
 - `build-android` is now `bundle:android`
3. Adds two new scripts:
 - `release:ios` - not implemented yet, but gives the link to the docs
 - `release:android` - creates a release APK for Android
4. Removes `react-native-clean-project` devDependency and invokes it using `npx` instead


boilerplate/package.json
```diff
@@ -7,27 +7,29 @@
...
     "postinstall": "node ./bin/postInstall",
-    "build-ios": "react-native bundle --entry-file index.js --platform ios --dev false --bundle-output ios/main.jsbundle --assets-dest ios",
-    "build-android": "react-native bundle --platform android --dev false --entry-file index.js --bundle-output android/app/src/main/assets/index.android.bundle --assets-dest android/app/src/main/res",
-    "clean": "react-native-clean-project",
+    "bundle:ios": "react-native bundle --entry-file index.js --platform ios --dev false --bundle-output ios/main.jsbundle --assets-dest ios",
+    "bundle:android": "react-native bundle --platform android --dev false --entry-file index.js --bundle-output android/app/src/main/assets/index.android.bundle --assets-dest android/app/src/main/res",
+    "release:ios": "echo 'Not implemented yet: release:ios. Use Xcode. More info: https://reactnative.dev/docs/next/publishing-to-app-store'",
+    "release:android": "cd android && rm -rf app/src/main/res/drawable-* && ./gradlew assembleRelease && cd - && echo 'APK generated in ./android/app/build/outputs/apk/release/app-release.apk'",
```
jamonholmgren, line 24: I added this `rm -rf` as I was running into issues where it would tell me there were duplicate files in that folder. They're copied over anyway, so this shouldn't be destructive...I don't think.

boilerplate/package.json
```diff
@@ -131,29 +132,31 @@
...
       "ios.sim.release": {
         "binaryPath": "ios/build/Build/Products/Release-iphonesimulator/HelloWorld.app",
-        "build": "xcodebuild -workspace ios/HelloWorld.xcworkspace -scheme HelloWorld -configuration Release -sdk iphonesimulator -derivedDataPath ios/build -destination 'name=iPhone 11'",
+        "build": "xcodebuild -workspace ios/HelloWorld.xcworkspace -scheme HelloWorld -configuration Release -sdk iphonesimulator -derivedDataPath ios/build -destination 'name=iPhone 13'",
         "type": "ios.simulator",
         "device": {
-          "name": "iPhone 11",
-          "os": "iOS 13.2"
+          "name": "iPhone 13",
+          "os": "iOS 15.5"
```
jamonholmgren, line 153: I would prefer that we not have to hardcode these, but we'll have to look at this later.

jamonholmgren:
Works for regular detox builds! (Expo still failing, will investigate why next week.)

CleanShot 2022-09-09 at 12 17 33@2x

PR #2139, opened 2022-09-09
feat(generator): custom file names

kateinkim:
## Describe your PR

Per the discussion https://github.com/infinitered/ignite/discussions/2133, the users can now configure the case of the generated filenames. The default is still `pascalCaseName` but it is configurable to both `kebabCaseName` or `camelCaseName`

## Photos

|Using Generator|`camelCaseName` in template|
|-|-|
|CleanShot 2022-09-09 at 14 19 44@2x|CleanShot 2022-09-09 at 13 46 45@2x|
|List of screens|index.ts|
|CleanShot 2022-09-09 at 14 15 19@2x|CleanShot 2022-09-09 at 13 48 58@2x|

### DIFF
--- a/boilerplate/ignite/templates/navigator/NAMENavigator.tsx.ejs
+++ b/boilerplate/ignite/templates/navigator/NAMENavigator.tsx.ejs
@@ -2,7 +2,7 @@
 destinationDir: app/navigators
 patch:
   path: "app/navigators/index.ts"
-  append: "export * from \"./<%= props.kebabCaseName %>-navigator\"\n"
+  append: "export * from \"./<%= props.pascalCaseName %>Navigator\"\n"
   skip: <%= props.skipIndexFile %>
 ---
 import React from "react"
--- a/boilerplate/ignite/templates/screen/NAMEScreen.tsx.ejs
+++ b/boilerplate/ignite/templates/screen/NAMEScreen.tsx.ejs
@@ -1,7 +1,7 @@
 ---
 patch:
   path: "app/screens/index.ts"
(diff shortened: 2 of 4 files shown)


kateinkim:
The inconvenience I'm seeing here is when people want to use `kebabCaseName`, they also have to change `Screen` to `-screen` in the path and the file names if they still want to keep "screen" at the end. The model and the component names are not going to have this issue but the screen and the navigator names and paths have to be configured accordingly.

jamonholmgren:
@kateinkim Can you explain further what you mean?

BTW, I'm probably going to hold off on merging this until either a future beta or post-Maverick.

kateinkim:
>  BTW, I'm probably going to hold off on merging this until either a future beta or post-Maverick. 

Yes, this is totally fine with me!

Here are the examples. as you see in the first picture, this default will generate `PizzaScreen` but if you see in the second picture with `kebabCaseName`, the screen name is supposed to be `pizza-screen` and to have that "-screen", the users have to manually change that in the template too. Not just the case options. Does this make sense @jamonholmgren ?

|Pascal Case|Kebab Case|
|-|-|
|CleanShot 2022-09-09 at 15 10 56@2x|CleanShot 2022-09-09 at 15 11 20@2x|

joshuayoes (approved):
This looks good! 

I see what you are talking about here regarding kebabCase: https://github.com/infinitered/ignite/pull/2139#issuecomment-1242453736

But I think if people are writing their own templates, they can handle that. I think the benefit of this PR standardizing on PascalCase is worth it.

PR #2142, opened 2022-09-10
Switching to async for copy/remove

jamonholmgren:
I noticed the CLI spinners were locking up when doing certain operations, so I made them async and parallelized them, and gained a little speed when restoring from a cache, plus not locking up the UI.

Before (no cache): 90s
After (no cache): 85s
Before (cached): 21s
After (cached): 19s

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
   "scripts": {
     "compile": "tsc -p .",
     "typecheck": "tsc -p . --noEmit",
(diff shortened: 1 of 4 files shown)


jamonholmgren:
Going to merge without review when CI passes, but would still like @joshuayoes's eyes on this when he gets a chance next week!

jamonholmgren:
Disabled cache by default until we get it rock-solid. Having issues with it.

PR #2144, opened 2022-09-10
Add `--dry-run` to `remove-demo` command

### LINKED ISSUE #2117: Add `--dry-run` to remove-demo command
This doesn't need to be done before v8, but it could be good to add a `--dry-run` flag to the `ignite-cli remove-demo` command, where we just print out what would be modified and deleted.

jamonholmgren:
Fixes #2117.

CleanShot 2022-09-09 at 18 56 20@2x

### DIFF
--- a/src/commands/remove-demo.ts
+++ b/src/commands/remove-demo.ts
@@ -2,6 +2,7 @@ import { GluegunToolbox } from "gluegun"
 import * as pathlib from "path"
 import type { CommentType } from "../tools/demo"
 import { demo } from "../tools/demo"
+import { boolFlag } from "../tools/flag"
 import { p, warning } from "../tools/pretty"
 
 const MATCHING_GLOBS = [
@@ -26,7 +27,7 @@ module.exports = {
 
     const CWD = process.cwd()
     const TARGET_DIR = parameters.first ?? CWD
-    const dryRun = parameters.options.dryRun ?? false
+    const dryRun = boolFlag(parameters.options.dryRun) ?? false
 
     p()
     p(`Removing demo code from '${TARGET_DIR}'${dryRun ? " (dry run)" : ""}`)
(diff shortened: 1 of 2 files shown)


jamonholmgren:
btw @joshuayoes, I think we shouldn't display the full path, but rather the relative path of the files.

So instead of this:

```
   Found '@demo remove-file' in /Users/jh/Code/Temp/PizzaApp/app/models/AuthenticationStore.ts
   Found '@demo remove-file' in /Users/jh/Code/Temp/PizzaApp/app/models/Episode.test.ts
   Found '@demo remove-file' in /Users/jh/Code/Temp/PizzaApp/app/models/Episode.ts
```

it would be this:

```
   Found '@demo remove-file' in app/models/AuthenticationStore.ts
   Found '@demo remove-file' in app/models/Episode.test.ts
   Found '@demo remove-file' in app/models/Episode.ts
```

Less visual clutter that way.

src/commands/remove-demo.ts
```diff
@@ -19,14 +19,17 @@ const MATCHING_GLOBS = [
...
   alias: ["rd", "remove-demos"],
-  description: "Remove demo code from generated boilerplate",
+  description:
+    "Remove demo code from generated boilerplate. Add --dry-run to see what would be removed.",
   run: async (toolbox: GluegunToolbox) => {
     const { parameters, patching, filesystem } = toolbox
 
     const CWD = process.cwd()
     const TARGET_DIR = parameters.first ?? CWD
+    const dryRun = parameters.options.dryRun ?? false
```
joshuayoes, line 29: Going to want to wrap this in a boolFlag, or else it will be `'true'`, `'false'`, or `undefined`

joshuayoes, line 29: Going to want to wrap this in a boolFlag, or else it will be `'true'`, `'false'`, or `undefined`

joshuayoes (commented):
Looks good except for my one comment about wrapping `--dry-run` in a boolFlag!

PR #2151, opened 2022-09-13
Fix: Calling console.tron in release builds causes crash

soulkito:
If the import to reactotron is locked behind `__DEV__ `the interface-compatible mock at [services/reactotron/reactotron.ts#L30](https://github.com/infinitered/ignite/blob/master/boilerplate/app/services/reactotron/reactotron.ts#L30) never gets attached causing the app to fail in production builds unless you prefix every `console.tron` with `__DEV__ && `.

### DIFF
--- a/boilerplate/app/models/environment.ts
+++ b/boilerplate/app/models/environment.ts
@@ -1,8 +1,8 @@
 import { Api } from "../services/api"
-
+import { Reactotron } from "../services/reactotron"
+  
 let ReactotronDev
 if (__DEV__) {
-  const { Reactotron } = require("../services/reactotron")
   ReactotronDev = Reactotron
 }
 


joshuayoes:
Thank you for this PR @soulkito. We are currently working toward a new major release, and we have changed the behavior of the Reactotron a decent amount. 

Can you try `npx ignite-cli@next new` and see if your workflow is fixed in the latest beta release?

soulkito:
I didn't have time to actually tested it but I ran `npx ignite-cli@next new` and it looks like you won't have the same problem there. 

The import (or require) to reactotron.ts is no longer behind a `if(__DEV__)` clause which means that the fake no-op code will be included for release builds. Everything looks good in the beta. 👍

joshuayoes:
[Yeah Reactotron is no longer imported behind a `if(__DEV__)` flag](https://github.com/infinitered/ignite/blob/maverick/boilerplate/app/services/reactotron/reactotron.ts#L16), so your use case should be good!

Issue #2141, opened 2022-09-10
[Maverick] Relative target-path throws an error


jamonholmgren:
### Describe the bug

Specifying a relative target-path on the CLI throws an error.

```
npx ignite-cli@new new PizzaApp --bundle=com.pizzaapp --git --install-deps --packager=yarn --target-path=./PizzaApp --overwrite
```

I get this error (note the double PizzaApp in the path):

```
/Users/jh/Code/ignite-cli/node_modules/fs-jetpack/lib/copy.js:52
  const err = new Error(`Path to copy doesn't exist ${path}`);
              ^
Error: Path to copy doesn't exist /Users/jh/Code/Temp/PizzaApp/PizzaApp/yarn.lock
    at generateNoSourceError (/Users/jh/Code/ignite-cli/node_modules/fs-jetpack/lib/copy.js:52:15)
    at /Users/jh/Code/ignite-cli/node_modules/fs-jetpack/lib/copy.js:170:15 {
  code: 'ENOENT'
}
```

### Ignite version

Maverick

### Additional info

no

jamonholmgren:
Note this also happens if you specify from home (`~`), like `~/Code/PizzaApp`.

Probably just need to run `filesystem.path` on it to normalize the path.

joshuayoes:
Resolved by https://github.com/infinitered/ignite/pull/2154

PR #2154, opened 2022-09-13
Handle relative paths and `~` character in `--target-path` of `new` command

joshuayoes:

## Describe your PR
Closes https://github.com/infinitered/ignite/issues/2141

### DIFF
--- a/src/commands/new.ts
+++ b/src/commands/new.ts
@@ -196,7 +196,7 @@ export default {
 
     // #region Project Path
     const defaultTargetPath = path(projectName)
-    let targetPath = useDefault(options.targetPath) ? defaultTargetPath : options.targetPath
+    let targetPath = useDefault(options.targetPath) ? defaultTargetPath : path(options.targetPath)
     if (targetPath === undefined) {
       const targetPathResponse = await prompt.ask(() => ({
         type: "input",
--- a/src/commands/new.ts
+++ b/src/commands/new.ts
@@ -505,7 +505,7 @@ export default {
     // #endregion
 
     // #region Cache dependencies
-    if (shouldFreshInstallDeps && cacheExists === false) {
+    if (shouldFreshInstallDeps && cacheExists === false && useCache) {
       const msg = `Saving ${packagerName} dependencies for next time`
       startSpinner(msg)
       log(targetPath)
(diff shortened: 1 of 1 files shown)


src/commands/new.ts
```diff
@@ -195,8 +195,12 @@ export default {
     // #endregion
 
     // #region Project Path
+    const parsePath = (p: string) =>
+      p.startsWith("~") ? path(p.replace("~", filesystem.homedir())) : path(p)
```
joshuayoes, line 199: For some reason, `filesystem.path` does not seem to like `~/Code/PizzaApp`. So instead, we are replacing the `~` with the actual home dir path

src/commands/new.ts
```diff
@@ -195,8 +195,12 @@ export default {
...
     const defaultTargetPath = path(projectName)
-    let targetPath = useDefault(options.targetPath) ? defaultTargetPath : options.targetPath
+    let targetPath = useDefault(options.targetPath)
+      ? defaultTargetPath
+      : parsePath(options.targetPath)
```
joshuayoes, line 203: Wrapping flag args in `filesystem.path` expands `./Code/PizzaApp` to `~/Code/PizzaApp`

src/commands/new.ts
```diff
@@ -397,7 +401,7 @@ export default {
     // Release Ignite installs have the boilerplate's .gitignore in .gitignore.template
     // (see https://github.com/npm/npm/issues/3763); development Ignite still
     // has it in .gitignore. Copy it from one or the other.
-    const targetIgnorePath = log(path(process.cwd(), ".gitignore"))
+    const targetIgnorePath = log(path(boilerplatePath, ".gitignore"))
```
joshuayoes, line 404: This is a regression that wasn't caught earlier

src/commands/new.ts
```diff
@@ -501,7 +505,7 @@ export default {
     // #endregion
 
     // #region Cache dependencies
-    if (shouldFreshInstallDeps && cacheExists === false) {
+    if (shouldFreshInstallDeps && cacheExists === false && useCache) {
```
joshuayoes, line 508: Even though `useCache` is `false` by default now, this was still getting copied. Added `useCache` check to only copy if `--use-cache` is `true`

jamonholmgren (approved):
looks great

PR #2159, opened 2022-09-14
Add test for `ignite generate model` and update Generator Template docs

joshuayoes:

## Describe your PR

Add tests for model generation and update docs

### DIFF
--- a/test/vanilla/ignite-generate.test.ts
+++ b/test/vanilla/ignite-generate.test.ts
@@ -1,0 +1,164 @@
+import { filesystem } from "gluegun"
+import * as tempy from "tempy"
+import { runIgnite } from "../_test-helpers"
+
+const BOILERPLATE_PATH = filesystem.path(__dirname, "../../boilerplate")
+
+const setup = (): { TEMP_DIR: string } => {
+  const TEMP_DIR = tempy.directory({ prefix: "ignite-" })
+
+  beforeEach(() => {
+    filesystem.copy(BOILERPLATE_PATH, TEMP_DIR, { overwrite: true })
+  })
+
+  afterEach(() => {
+    filesystem.remove(TEMP_DIR) // clean up our mess
+  })
+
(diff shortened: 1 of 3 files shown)


test/vanilla/ignite-generate.test.ts
```diff
@@ -0,0 +1,164 @@
...
+        /**
+         * The RootStore instance.
+         */
+        export interface RootStore extends Instance<typeof RootStoreModel> {}
+        /**
+         * The data of a RootStore.
+         */
+        export interface RootStoreSnapshot extends SnapshotOut<typeof RootStoreModel> {}
+        "
+      `)
```
jamonholmgren, line 161: I think for this if we are going to get so specific we should just do a regular snapshot instead.

I'd probably just check for `export const RootStoreModel` and call it good, unless there's a specific reason we need to test all the other details.

jamonholmgren (approved):
Looks good, concerned a bit about hardcoding the generated template as it'll make maintenance harder.

test/vanilla/ignite-generate.test.ts
joshuayoes: So the nice thing about this, is that Jest automatically creates this snapshot: https://jestjs.io/docs/snapshot-testing#inline-snapshots

It's the same as `toMatchSnapshot`, but the snapshot is colocated

joshuayoes: And the snapshot can be updated by pressing `u` in the test runner

joshuayoes: Main thing I like about snapshots or things like this is that we get more detailed error messages when a test throws

joshuayoes: Here is a video of what I'm talking about

https://user-images.githubusercontent.com/37849890/190252650-8603ffda-e0ef-4d8c-96aa-0fca7d599b27.mov

Issue #2145, opened 2022-09-10
Catch non-camelCase project name, reformat to camelCase as default project name


edjw:
I started with a kebab case project name. I got the error that it needed to be camel case and it showed what my title would be in camel case. I think it'd be nicer if the cli would not quit completely at that point. And instead it would go again to the project title question with the camel case version of the user's first attempt at a project name as the default answer

Here's a screenshot of what I did without concentrating very hard

<img width="571" alt="grafik" src="https://user-images.githubusercontent.com/6547979/189474998-cec798ae-452c-406c-a3bf-b553810a5c22.png">

joshuayoes:
I've created a PR to update the prompt to show the error message, but give the opportunity to fix it before exiting the process

https://user-images.githubusercontent.com/37849890/190263094-aa8a6a8e-3bfa-41e1-9662-0f902bcd3940.mov

joshuayoes:
Resolved now by https://github.com/infinitered/ignite/pull/2161

PR #2161, opened 2022-09-14
Add `validateProjectName` guards to prompt

joshuayoes:

- [X] `README.md` has been updated with your changes, if relevant

## Describe your PR
Resolves: https://github.com/infinitered/ignite/issues/2145

Solves the issue by extracting existing `validateBundleIdentifier` validations into guard functions. Now, each guard can be used both by the enquirer prompt and the cli flag arg

### DIFF
--- a/src/tools/validations.ts
+++ b/src/tools/validations.ts
@@ -1,9 +1,50 @@
-import { GluegunToolbox } from "gluegun"
+import { GluegunToolbox, strings } from "gluegun"
 import { prefix } from "./pretty"
+const { pascalCase } = strings
+
+// #region Error Guards
+type IsError = (str: string) => boolean
+type ErrorMessage = (str?: string) => string
+type ErrorGuard = [IsError, ErrorMessage]
+
+const isIgnite: ErrorGuard = [
+  (str) => str.toLowerCase() === "ignite",
+  (str) => `Hey...that's my name! Please name your project something other than '${str}'.`,
+]
+const isOnlyNumbers: ErrorGuard = [
+  (str) => /^\d+$/.test(str),
+  () => `Please use at least one non-numeric character for your project name`,
+]
+const isKebabCase: ErrorGuard = [
+  (str) => str.includes("-"),
+  (str) =>
+    `Please use 'PascalCase', not 'kebab-case' for your project name. Ex: ${pascalCase(str)}`,
+]
+const isNotAlphaNumeric: ErrorGuard = [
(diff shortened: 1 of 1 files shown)


src/tools/validations.ts
```diff
@@ -1,9 +1,50 @@
...
+const isKebabCase: ErrorGuard = [
+  (str) => str.includes("-"),
+  (str) =>
+    `Please use 'PascalCase', not 'kebab-case' for your project name. Ex: ${pascalCase(str)}`,
+]
+const isNotAlphaNumeric: ErrorGuard = [
+  (str) => !/^[a-z_][a-z0-9_]+$/i.test(str),
+  () =>
+    `The project name can only contain alphanumeric characters and underscore, but must not begin with a number.`,
+]
```
joshuayoes, line 27: All of these guard function are the same checks as the existing conditionals in `validateProjectName`

However, the advantage is that we can use these guards now both for the enquirer prompt and cli parameter

src/tools/validations.ts
```diff
@@ -1,9 +1,50 @@
...
+ * @returns error message from the first guard that matches, or `true` if no guards match
+ */
+const validate = (value: string): true | string => {
+  for (const [isError, errorMessage] of guards) {
+    if (isError(value)) {
+      return errorMessage(value)
+    }
+  }
+  return true
+}
```
joshuayoes, line 42: [`GluegunPrompt` can accept a `validate` function](https://github.com/infinitered/gluegun/blob/master/src/toolbox/prompt-enquirer-types.ts#L14), that either returns `true` if it is a valid value, or a string if it is an error

Here is an demo of what that looks like:

https://user-images.githubusercontent.com/37849890/190262670-a849633a-59c7-4171-aaf3-29c4d9529aed.mov

src/tools/validations.ts
```diff
@@ -24,34 +66,9 @@ export async function validateProjectName(toolbox: GluegunToolbox): Promise<stri
...
-  }
-
-  // check for alphanumeric name, beginning with a letter
-  if (!/^[a-z_][a-z0-9_]+$/i.test(projectName)) {
-    print.error(
-      `The project name can only contain alphanumeric characters and underscore, but must not begin with a number.`,
-    )
+  const error = validate(projectName)
+  if (typeof error === "string") {
+    print.error(error)
```
joshuayoes, line 71: This preserves the original logic checking the `projectName` when passed via `parameters.first`

jamonholmgren (approved):

Issue #2174, opened 2022-09-17
[Maverick] Missing .gitignore?


juddey:
### Describe the bug

When running ignite-new PizzaApp and choosing to create git repoisitory, theres no .gitignore present in the created directories.

this means everything is committed to source control, including `node_modules`....

Is this expected behaviour?

![Screen Shot 2022-09-17 at 5 33 02 PM](https://user-images.githubusercontent.com/7445602/190842444-eecbb46b-f654-432e-aab4-537d084866c6.png)

### Ignite version

8.0.0-beta-5

### Additional info

As per #2170

frankcalise:
Hi @juddey thanks for reporting this issue! Can you please confirm that this is just an issue when using `npm` as the package manager? Also if you don't mind, paste the command that gets output in the terminal for the options you used:

<img width="461" alt="image" src="https://user-images.githubusercontent.com/374022/190863468-797413c4-87f4-4f58-a8dc-c4854c3ad319.png">

juddey:
Hi Frank,

Here's what I'm running: 

```
npx ignite-cli@next new DemoFrank
```

I select npm, and I noticed I also install deps: 

<img width="384" alt="Screen Shot 2022-09-19 at 10 39 46 AM" src="https://user-images.githubusercontent.com/7445602/190931255-a3e38369-7492-4a00-84b0-b06d36b9d6fe.png">

To eliminate the deps difference I created a repo and did not install deps:

<img width="519" alt="Screen Shot 2022-09-19 at 10 43 27 AM" src="https://user-images.githubusercontent.com/7445602/190931313-8ea3e372-3a2e-45ce-afd8-1b90748d3b85.png">

As for the question about the yarn package manager, I've found that I can't have yarn on my system and have ignite successfully create a new app. I think it gets confused about the node source, as yarn also installs a node reference of some kind...

However, if you're able to install with yarn and still get a .gitignore then that's enough of an indicator that its restricted to npm 😄

frankcalise:
@juddey I can't seem to reproduce this, can you run `npx ignite-cli@next doctor` from the project directory for some extra info and paste it here?

juddey:
Sure!

```
Jocelyns-Mini:DemoFrank Justin$ npx ignite-cli@next doctor
System
  platform           darwin                                                
  arch               x64                                                   
  cpu                12 cores     Intel(R) Core(TM) i7-8700B CPU @ 3.20GHz 
[...]

PR #2184, opened 2022-09-20
Simplify `.gitignore` copy logic

joshuayoes:

## Describe your PR
Resolves https://github.com/infinitered/ignite/issues/2174

### DIFF
--- a/src/commands/new.ts
+++ b/src/commands/new.ts
@@ -389,30 +389,19 @@ export default {
       overwrite,
     })
     stopSpinner(" 3D-printing a new React Native app", "🖨")
+    // copy the .gitignore if it wasn't copied over
+    // Release Ignite installs have the boilerplate's .gitignore in .gitignore.template
+    // (see https://github.com/npm/npm/issues/3763); development Ignite still
+    // has it in .gitignore. Copy it from one or the other.
+    const boilerplateIgnorePath = log(path(boilerplatePath, ".gitignore.template"))
+    const targetIgnorePath = log(path(targetPath, ".gitignore"))
+    copy(boilerplateIgnorePath, targetIgnorePath, { overwrite: true })
 
     // note the original directory
     const cwd = log(process.cwd())
 
     // jump into the project to do additional tasks
     process.chdir(targetPath)
-
-    // copy the .gitignore if it wasn't copied over
-    // Release Ignite installs have the boilerplate's .gitignore in .gitignore.template
-    // (see https://github.com/npm/npm/issues/3763); development Ignite still
-    // has it in .gitignore. Copy it from one or the other.
-    const targetIgnorePath = log(path(boilerplatePath, ".gitignore"))
-    if (!exists(targetIgnorePath)) {
-      // gitignore in dev mode?
-      let sourceIgnorePath = log(path(boilerplatePath, ".gitignore"))
-
-      // gitignore in release mode?
-      if (!exists(sourceIgnorePath)) {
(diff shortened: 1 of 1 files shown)


src/commands/new.ts
```diff
@@ -389,30 +391,26 @@ export default {
       overwrite,
     })
     stopSpinner(" 3D-printing a new React Native app", "🖨")
+    // copy the .gitignore if it wasn't copied over
+    // Release Ignite installs have the boilerplate's .gitignore in .gitignore.template
+    // (see https://github.com/npm/npm/issues/3763); development Ignite still
+    // has it in .gitignore. Copy it from one or the other.
+    const boilerplateIgnorePath = exists(boilerplate(".gitignore.template"))
+      ? boilerplate(".gitignore.template")
+      : boilerplate(".gitignore")
```
joshuayoes, line 400: Simplified this logic a little. The `.gitignore.template` should always be there in a production build

However, in development, when running `npx ignite-cli` after running `npx link`, there is chance that `.gitignore.template` will not be there after the `yarn clean` command.

src/commands/new.ts
```diff
@@ -389,30 +391,26 @@ export default {
...
+    const targetIgnorePath = log(path(targetPath, ".gitignore"))
+    copy(log(boilerplateIgnorePath), targetIgnorePath, { overwrite: true })
```
joshuayoes, line 402: We always want to copy this over not matter what

src/commands/new.ts
```diff
@@ -389,30 +391,26 @@ export default {
...
+
+    if (exists(targetIgnorePath) === false) {
+      warning(`  Unable to copy ${boilerplateIgnorePath} to ${targetIgnorePath}`)
+      process.exit(1)
+    }
```
joshuayoes, line 407: Add this so we can get better reporting when it fails

src/commands/new.ts
```diff
@@ -389,30 +391,25 @@ export default {
...
+    // has it in .gitignore. Copy it from one or the other.
+    const boilerplateIgnorePath = exists(boilerplate(".gitignore.template"))
+      ? boilerplate(".gitignore.template")
+      : boilerplate(".gitignore")
+    const targetIgnorePath = log(path(targetPath, ".gitignore"))
+    copy(log(boilerplateIgnorePath), targetIgnorePath, { overwrite: true })
+
+    if (exists(targetIgnorePath) === false) {
+      warning(`  Unable to copy ${boilerplateIgnorePath} to ${targetIgnorePath}`)
+    }
```
joshuayoes, line 406: Added a warn here so we can get better reporting when it fails

src/commands/new.ts
```diff
@@ -389,30 +391,25 @@ export default {
...
 
     // jump into the project to do additional tasks
     process.chdir(targetPath)
-
-    // copy the .gitignore if it wasn't copied over
-    // Release Ignite installs have the boilerplate's .gitignore in .gitignore.template
-    // (see https://github.com/npm/npm/issues/3763); development Ignite still
-    // has it in .gitignore. Copy it from one or the other.
-    const targetIgnorePath = log(path(boilerplatePath, ".gitignore"))
-    if (!exists(targetIgnorePath)) {
```
joshuayoes, line 404: I suspect this is the offending logic here. This file isn't in the global NPM package artifacts. So it will never get copied over in production.

However, in development this actually exists

jamonholmgren (approved):

PR #2187, opened 2022-09-21
More Toggle updates.

### LINKED ISSUE #2180: Update Toggle with new design updates.
The Toggle designs have been updated to include extended examples of different states. @cdanwards did most of these updates but some very specific state combo styles are still missing.

- [x] Add `helper` props similar to TextField.
yulolimum-capture-2022-09-20--00-36-34
- [x] Add tint to the `label` when disabled.
yulolimum-capture-2022-09-20--00-36-58
- [x] Fix border animation issue with Switch's border when transitioning from `on` to `off` state.
yulolimum-capture-2022-09-20--00-39-19
- [x] Fix tint of `switchAccessibility` according to new colors.
yulolimum-capture-2022-09-20--00-40-00
- [x] Fix Radio's disabled  state
yulolimum-capture-2022-09-20--00-41-06

yulolimum:

## Describe your PR

- Resolves #2180.
- Adds "helper" functionality/props to Toggle.
- Adds error tint to label/helper in Toggle when.
- Removes border from switch variant.
- Fixes switchAccessibility tint.
- Fixes radio variant regression.

### DIFF
--- a/boilerplate/app/components/Toggle.tsx
+++ b/boilerplate/app/components/Toggle.tsx
@@ -3,7 +3,6 @@ import {
   GestureResponderEvent,
   Image,
   ImageStyle,
-  Platform,
   StyleProp,
   SwitchProps,
   TextInputProps,
@@ -48,6 +47,10 @@ export interface ToggleProps extends Omit<TouchableOpacityProps, "style"> {
    */
   containerStyle?: StyleProp<ViewStyle>
   /**
+   * Style overrides for the input wrapper
+   */
+  inputWrapperStyle?: StyleProp<ViewStyle>
+  /**
    * Optional input wrapper style override.
    * This gives the inputs their size, shape, "off" background-color, and outer border.
    */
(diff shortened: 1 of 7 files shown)


boilerplate/app/components/Toggle.tsx
```diff
@@ -184,13 +228,13 @@ function Checkbox(props: ToggleInputProps) {
...
   const offBackgroundColor = [
-    disabled && colors.palette.neutral600,
+    disabled && colors.palette.neutral400,
     status === "error" && colors.errorBackground,
     colors.palette.neutral200,
   ].filter(Boolean)[0]
 
   const outerBorderColor = [
-    disabled && colors.border,
+    disabled && colors.palette.neutral400,
```
cdanwards, line 237: @yulolimum just curious about this. We had a conversation about using palette vs semantic name on my design update PR. Would this not be a use case for `colors.border` since it's being used in the `outerBorderColor`?

cdanwards (approved):

boilerplate/app/components/Toggle.tsx
yulolimum: @cdanwards I saw that convo and I'm torn tbh. I made this change for the sake of consistency since we reference the palette for most other colors inside components. Honestly, I think either approach works _for now_. The reason I say _for now_ is because one of my next tasks will be to add an additional layer to styling to support multiple themes - a color mapping file. So, all of these color references will change in the coming weeks anyways.

PR #2232, opened 2022-09-28
Implement ignite-cli remove-demo-markup command

### LINKED ISSUE #2228: Add `ignite remove-demo-markup` command
Add a new command, `npx ignite-cli remove-demo-markup`, which strips all the `@demo remove-line` style comments.

This is probably not going to be super simple as we'd need to recognize different types of comments (`//`, `#`, `/* */`, `{ /* */ }`, etc). Might be tough to do, actually? But worth noting as an idea.

frankcalise:

## Describe your PR
- Closes #2228
- Removes comments related to our demo mark notation noted below
- Runs prettier format after comment removal to clean up spacing

```
@demo remove-file
@demo remove-current-line
@demo remove-next-line
@demo remove-block-start
@demo remove-block-end
```

in the following comment types

```
//
/* */
{/* */}
```

*NOTE*: this will currently _NOT_ handle the case for a multi line @demo markup comment such as the following:
```
/*   @demo

    more content here

*/
```

image


src/commands/remove-demo-markup.ts
```diff
@@ -0,0 +1,117 @@
+import { GluegunToolbox } from "gluegun"
```
frankcalise, line 1: Quite a bit of this is duplicated from `./src/commands/remove-demo.ts`

I'd be happy to refactor `remove-demo` to include another switch for `--markup` or something similar and just conditionally check if we should remove files or remove markup.

Thoughts?

src/tools/demo.ts
```diff
@@ -6,6 +6,8 @@ export enum CommentType {
   REMOVE_BLOCK_END = `@demo remove-block-end`,
 }
 
+export const demoMarkupRegex = /\s*\/\/\s*@demo.*|{?\/.*@demo.*\/}?/gm
```
frankcalise, line 9: As in the PR notes, this will not capture multiline block comments including @demo

Examples that are currently found:
- /* @demo */
- // @demo
- {/* @demo */}

src/tools/demo.ts
```diff
@@ -85,10 +87,21 @@ function remove(contents: string): string {
   return result
 }
 
+/**
+ * Perform replace on all types of @demo markup
+ * @param contents The file contents as a string
+ * @return The file contents with all @demo related CommentType removed
+ */
+function sanitize(contents: string): string {
```
frankcalise, line 95: Maybe rename to `sanitizeMarkup` ?

src/commands/remove-demo-markup.ts
jamonholmgren: Maybe `--only-markup` ?

jamonholmgren (approved):

jamonholmgren:
This looks amazing.

src/commands/remove-demo-markup.ts
joshuayoes: I think we could also migrate a lot of this functionality into the `demo` tool as new methods. 
 
```ts
const update = (filePaths: string[]): Promise<{ path: string; comments: string[] }>[] => {
  // demoCommentResults block
  // ...
  return demoCommentResults.sort((a, b) => {
    if (a.status === "fulfilled" && b.status === "fulfilled") {
      return a.value.path.localeCompare(b.value.path)
    }
    return 0
  })
}

// ...

const find = (targetDir: string, matching?: string[]) => {
  const MATCHING_GLOBS = [
    "!**/.DS_Store",
    "!**/.expo{,/**}",
    "!**/.git{,/**}",
    "!**/.vscode{,/**}",
    "!**/node_modules{,/**}",
    "!**/ios/build{,/**}",
    "!**/ios/Pods{,/**}",
    "!**/ios/*.xcworkspace{,/**}",
    "!**/android/build{,/**}",
    "!**/android/app/build{,/**}",
  ]

  const filePaths = filesystem
    .cwd(targetDir)
    .find({
      matching: matching ?? MATCHING_GLOBS,
      recursive: true,
      files: true,
      directories: false,
    })
    .map((path) => pathlib.join(targetDir, path))
  return filePaths
}

export const demo = {
  // ...
  update,
  find,
} as const
```

src/tools/demo.ts
joshuayoes: Could we leave this limitation as a comment in the codebase?

src/tools/demo.ts
joshuayoes: I'm okay with `sanitize` as long as the comment is there.

joshuayoes (commented):
This looks good! My refactor comment is a nice to have: https://github.com/infinitered/ignite/pull/2232/files#r982746738

But I'm also good with merging this as is. Thanks for tackling this @frankcalise!

src/tools/demo.ts
frankcalise: Yeah good idea I'll add that.

src/commands/remove-demo-markup.ts
frankcalise: @joshuayoes could have sworn I wrote a comment earlier 🙃 . Would we need to pass in filesystem and patching here to make this work as well? Is that something we want to do?

joshuayoes: Yeah, we can import these directly. I've talked with @jamonholmgren about this, and we don't really utilize dependency injection in our tests, so we can import them directly from toolbox

joshuayoes:
```
import { filesystem, patching } from "gluegun"
```

PR #2244, opened 2022-10-03
chore(docs): update ignite-cli@next to ignite-cli@latest

joshuayoes:

## Describe your PR
We have started showing example usage of `ignite-cli` to use `npx ignite-cli@next`. However, we are no longer publishing the latest version using those tags. It appears NPM already has an `@latest` tag applied, so we should update our docs to use the build-in tag

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,10 +52,10 @@ Run the CLI:
 
 ```bash
 # Get walked through the prompts for the different options to start your new app
-npx ignite-cli@next new PizzaApp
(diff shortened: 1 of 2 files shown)


joshuayoes:
Another possible solution could be to update semantic release on master to use the `--tag=next` flag

joshuayoes:
I think `@next` should point to the latest version for semantic releases, or the latest release of new major version betas like Maverick

PR #2246, opened 2022-10-03
Add Logout button to Debug screen

### LINKED ISSUE #2215: Add a way to logout after the welcome screen
Currently there's on a logout button on the welcome screen, but once you move into the tab navigator, there's no obvious way to get back to that screen, only swiping. 

We can add logout to the debug screen or to the dashboard, or something similar. Open to thoughts.

We should also move the Debug tab to the the furthest right.

frankcalise:

## Describe your PR
- Closes #2215
- Moves Debug tab to the fourth position on the tab bar
- Adds logout button to the Debug screen
- Refactors logout function to store action for Welcome and Debug screen usage

image

### DIFF
--- a/boilerplate/app/models/AuthenticationStore.ts
+++ b/boilerplate/app/models/AuthenticationStore.ts
@@ -40,6 +40,8 @@ export const AuthenticationStoreModel = types
     },
     logout() {
       store.authToken = undefined
+      store.authEmail = ""
+      store.authPassword = ""
     },
   }))
 
--- a/boilerplate/app/models/AuthenticationStore.ts
+++ b/boilerplate/app/models/AuthenticationStore.ts
@@ -38,6 +38,9 @@ export const AuthenticationStoreModel = types
     setAuthPassword(value: string) {
       store.authPassword = value.replace(/ /g, "")
     },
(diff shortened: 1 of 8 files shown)


boilerplate/app/models/AuthenticationStore.ts
```diff
@@ -38,6 +38,9 @@ export const AuthenticationStoreModel = types
     setAuthPassword(value: string) {
       store.authPassword = value.replace(/ /g, "")
     },
+    logout() {
+      store.authToken = undefined
+    },
```
mazenchami, line 43: [q] do we need to also clear `authEmail` and `authPassword`?

frankcalise: Yeah we could do that - we didn't before but the `useEffect` in `LoginScreen` to prepopulate should run again since that screen would get mounted again with the navigator being conditional on `isAuthenticated` (which boils down to !!authToken)

frankcalise: Updated!

mazenchami (approved):
LGTM! 🚀

PR #2258, opened 2022-10-06
fix(rdm): allow prettier formatting w/o deps installed

### LINKED ISSUE #2241: `npx ignite-cli remove-demo-markup` unable to run prettier if no deps installed
The way `npx ignite-cli remove-demo-markup` works is removing some comment blocks related to marking demo code and then runs prettier over the codebase to correct any formatting that is needed.

However, if the user decided not to install dependencies yet and decides to run this command, the prettier formatting will fail.

We'll need to improve the regex to get the spacing right during string replacement, consider running `npx prettier` or come up with another solution (possibly just warning the user it can't be run as the dependency wasn't found).

frankcalise:

## Describe your PR
- Closes #2241 
- Originally the `remove-demo-markup` command had to have dependencies installed to properly run prettier
- Substituted the local command for `npx prettier` so this can be run prior to installing deps

### DIFF
--- a/src/commands/remove-demo-markup.ts
+++ b/src/commands/remove-demo-markup.ts
@@ -1,7 +1,6 @@
 import { GluegunToolbox } from "gluegun"
 import { demo } from "../tools/demo"
 import { boolFlag } from "../tools/flag"
-import { packager } from "../tools/packager"
 import { p, warning } from "../tools/pretty"
 
 module.exports = {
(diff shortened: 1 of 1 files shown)


joshuayoes (approved):
Looks good! 

I know that the packager has a "global" option to run `npx` vs `npm`, but I don't think it's worth messing with on this PR

PR #2281, opened 2022-10-31
vendor: remove expo-linear-gradient

### LINKED ISSUE #2280: Remove unused package, expo-linear-gradient
`expo-linear-gradient` was left over from the previous (before Maverick) components, where we had a `GradientBackground`. This no longer exists and does not need to be in the boilerplate.

Making a task out of this because although it has Experimental Fabric support in Expo v47, it would help get use closer to New Architecture compatible if we drop it.

frankcalise:

## Describe your PR
- Closes #2280 

`expo-linear-gradient` was left over from an old pre Maverick component that we don't have anymore.

### DIFF
--- a/boilerplate/ios/Podfile.lock
+++ b/boilerplate/ios/Podfile.lock
@@ -10,30 +10,28 @@ PODS:
     - ExpoModulesCore
   - EXFileSystem (14.1.0):
     - ExpoModulesCore
-  - EXFont (10.2.0):
+  - EXFont (10.2.1):
     - ExpoModulesCore
-  - Expo (46.0.10):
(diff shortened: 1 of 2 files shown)


frankcalise:
@yulolimum Am I right in saying we don't provide that GradientBackground component anymore? Was it planned to be added or can we go through with this change?

yulolimum:
> @yulolimum Am I right in saying we don't provide that GradientBackground component anymore? Was it planned to be added or can we go through with this change?

Yea man, no need for it right now. I think we rarely use this package on projects as well.

yulolimum (approved):

PR #2283, opened 2022-11-01
feat(issue-cmd): added new GH issue command

### LINKED ISSUE #932: ignite doctor --bug should work
In [the docs](https://github.com/infinitered/ignite/blob/1c677e1e53b6de27a4c9386e5aa2c6a980fe8071/docs/quick-start/ignite-commands.md#doctor), we mention `ignite doctor --bug` which ostensibly should open a browser with an Ignite issue ready to go and the results of `ignite doctor` pre-filled.

I'm not sure where this idea came from, but it's a good one. We should implement it.

frankcalise:
## Describe your PR
- Closes #932 
- Adds new `npx ignite-cli issue "my bug title here"` command
- Opens browser window to `https://github.com/infinitered/ignite/issue/new` with title prefilled and prefilled issue body with `npx ignite doctor` output for issue reporter to easily paste
- Fixes title for user if they forget to double quote the title string

Tested
- [x] MacOS
- [x] Windows

image

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,8 @@ Note:
 
 If you run into problems, first search the issues and discussions in this repository. If you don't find anything, you can come talk to our friendly and active developers in the Infinite Red Community Slack ([community.infinite.red](http://community.infinite.red)).
 
+If you still need help after reaching out to the proper channels, feel free to open a new GitHub issue via `npx ignite-cli issue "Unable to Ignite new app"` (as an example). This will help start writing your issue with the correct diagnostic information included.
+
 ## No time to learn React Native? Hire Infinite Red for your next project
 
 We get it – sometimes there just isn’t enough time on a project to learn the ins and outs of a new framework. Infinite Red’s here to help! Our experienced team of React Native engineers have worked with companies like GasBuddy, Zoom, and Mercari to bring even some of the most complex projects to life.
--- a/src/commands/issue.ts
+++ b/src/commands/issue.ts
@@ -84,7 +84,7 @@ module.exports = {
     // TODO looked into gh cli but not sure that is worth it either
     try {
       p(cyan("Starting GitHub issue..."))
-      system.run(`${START_CMD} ${url}`)
+      await system.run(`${START_CMD} ${url}`)
 
       p(cyan("Paste the following in the comment body and fill out the description:"))
       p()
--- a/src/commands/issue.ts
+++ b/src/commands/issue.ts
@@ -84,7 +84,7 @@ module.exports = {
 
     try {
       p(cyan("Starting GitHub issue, thanks for alerting us!"))
-      await system.run(`${START_CMD} \"${url}?${params.toString()}\"`)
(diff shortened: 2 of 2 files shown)


frankcalise:
Thoughts @joshuayoes @jamonholmgren ?

joshuayoes:
I love this! I know you mentioned the GitHub CLI but I like this approach. I think allowing people to quickly open up an issue and then edit it in the browser is the right approach here.

frankcalise:
@joshuayoes actually I think I can get it to paste the body correctly, was passing the url params incorrectly. however, do you know of any way to capture the output from `system.run` without colors? I think the colored output from `npx ignite-doctor` is messing things up here

image

joshuayoes:
You should be able to use the strip-ansi package, we use it for the test output: https://github.com/infinitered/ignite/blob/master/test/_test-helpers.ts#L2

frankcalise:
@joshuayoes wow awesome thanks! worked perfectly

image

frankcalise:
I think it could be cool to search the existing issues for the passed in title to output a few of the links to the results in the terminal for them to look up prior to submitting, but this implementation is probably a good first step to start with.

joshuayoes (approved):

src/commands/issue.ts
```diff
@@ -0,0 +1,94 @@
...
+  startSpinner,
+  stopSpinner,
+  prettyPrompt,
+  prefix,
+} from "../tools/pretty"
+const stripANSI = require("strip-ansi")
+
+const isMac = process.platform === "darwin"
+
+const START_CMD = isMac ? "open" : "start"
```
jamonholmgren, line 19: Won't work on Linux, but we'll let them fix with their own command.

jamonholmgren: And of course it's ridiculous over there: https://stackoverflow.com/questions/5116473/linux-command-to-open-url-in-default-browser

jamonholmgren: We could use `open` -- https://github.com/sindresorhus/open

Or this command, which won't work with all distros, but get us mostly there: https://stackoverflow.com/a/49013356/204044

But for this PR, let's not worry about it.

README.md
```diff
@@ -92,6 +92,8 @@ Note:
 
 If you run into problems, first search the issues and discussions in this repository. If you don't find anything, you can come talk to our friendly and active developers in the Infinite Red Community Slack ([community.infinite.red](http://community.infinite.red)).
 
+If you still need help after reaching out to the proper channels, feel free to open a new GitHub issue via `npx ignite-cli issue "Unable to Ignite new app"` (as an example). This will help start writing your issue with the correct diagnostic information included.
```
jamonholmgren, line 95: We really need an `ignite-cli` doc, documenting all commands. Not in this PR though, filed an issue #2291.

jamonholmgren (approved):

src/commands/issue.ts
frankcalise: Ah was hoping there was some cross platform *nix usage with `open` but I guess not

PR #2284, opened 2022-11-02
fix(new-cmd): dedupe pod install running 2x

### LINKED ISSUE #2227: Look into `new` command pod install
When igniting a new app, if the user chooses to install dependencies we run the following:

```
   // #region Install CocoaPods
    // install pods
    if (shouldFreshInstallDeps) {
      startSpinner("Baking CocoaPods")
      await spawnProgress(`npx pod-install@${deps.podInstall}`, {
        onProgress: log,
      })
      stopSpinner("Baking CocoaPods", "☕️")
    }
    // #endregion
```

Is it possible this action is then duplicated via our `./boilerplate/bin/postInstall` script during `yarn install`? This script also goes through the necessary MacOS and ios folder checks to then run `pod install`.

frankcalise:

## Describe your PR
- Closes #2227 
- Enhances postInstall script from `pod install` to `npx pod-install`
- Removes `npx pod-install` call from `new` command since handled via postInstall


boilerplate/bin/postInstall
```diff
@@ -15,8 +15,9 @@ if (["darwin", "linux"].includes(process.platform)) {
 }
 
 // On iOS, make sure CocoaPods are installed
+// set env var IGNITE_SKIP_POD_INSTALL to any value to skip
 if (process.platform === "darwin") {
-  run('if [ -d "ios" ]; then cd ios && pod install && cd -; fi')
+  run('if [ -d "ios" ] && [ -z "$IGNITE_SKIP_POD_INSTALL" ]; then cd ios && pod install && cd -; fi')
```
jamonholmgren, line 20: Should this be `npx pod-install` here?

src/commands/new.ts
```diff
@@ -515,17 +509,6 @@ export default {
     stopSpinner(renameSpinnerMsg, "🎨")
     // #endregion
 
-    // #region Install CocoaPods
-    // install pods
-    if (shouldFreshInstallDeps) {
-      startSpinner("Baking CocoaPods")
-      await spawnProgress(`npx pod-install@${deps.podInstall}`, {
```
frankcalise, line 522: Simply remove `npx pod-install` during new command since it would have just been called on post install from the package manager command

boilerplate/bin/postInstall
```diff
@@ -15,8 +15,9 @@ if (["darwin", "linux"].includes(process.platform)) {
 }
 
 // On iOS, make sure CocoaPods are installed
+// set env var IGNITE_SKIP_POD_INSTALL to any value to skip
 if (process.platform === "darwin") {
-  run('if [ -d "ios" ]; then cd ios && pod install && cd -; fi')
+  run('if [ -d "ios" ]; then cd ios && npx pod-install && cd -; fi')
```
frankcalise, line 19: `npx pod-install` now here over just plain `pod install`. This helps check for updating the pod repo etc if things are out of sync

joshuayoes (approved):
I have a small nitpick, but this looks good! Thanks for doing this Frank

boilerplate/bin/postInstall
frankcalise: Oh nice I forgot that it did that, good call!

boilerplate/bin/postInstall
```diff
@@ -16,7 +16,7 @@ if (["darwin", "linux"].includes(process.platform)) {
 
 // On iOS, make sure CocoaPods are installed
 if (process.platform === "darwin") {
-  run('if [ -d "ios" ]; then cd ios && pod install && cd -; fi')
+  run('npx pod-install')
```
jamonholmgren, line 19: I kinda think the `if [ -d "ios" ]; then` is still useful here, in case someone goes full managed and removes their `ios` and `android` folders.

jamonholmgren (approved):
Ship it!

boilerplate/bin/postInstall
frankcalise: This is OK, `pod-install` package handles that case:

image

frankcalise: It checks for darwin too, so we could simplify further

frankcalise: image

Windows check for sanity

PR #2285, opened 2022-11-02
feat(generators): support for component subdirectory

### LINKED ISSUE #1746: Generate a component in specific path/subfolder
how can i generate a component or other with `generate` in a specific folder?
My project is split into several sub folders and is not flat like the boilerplate.
I tried  `ignite generate component path1/path2/Name` but do not work

Thank you

frankcalise:

  - [x] No subdirectory passes
  - [x] Subdirectory generated, relative paths unneeded by adding TS absolute path alias

## Describe your PR
- Closes #1746 
- Supports components to be generated in a subdirectory
- Supplies frontmatter props with `subdirectory` (to extend other templates)
- Added test to check the output of both root components and subdirectory nested components
- Updated documentation for new prop

image

image


frankcalise:
@kateinkim @joshuayoes @jamonholmgren 

Any suggestions here to solve the relative path import back to the `Text` component and `theme` imports?

I could of course clear out the Text component and remove the associated style.

How do we feel about adding a prop for relativeThemesPath and relativeComponentsPath?

joshuayoes:
Instead of adding multiple props, what if we just included one `relativeAppPath` prop? Then we wouldn't need to add props for each app subdirectory

It would make the paths a little more verbose than necessary in the output, but I think it would be more clear writing them in the generator file. 

The generator would look something like this:

```
import { colors, typography } from "<%= props.relativeRootPath %>/theme/colors"
import { Text } from "<%= props.relativeRootPath %>/components/Text"
```

The test case `generate component sub/to/my/Topping` would output this:

`/app/components/sub/to/my/Topping.tsx`
```
import { colors, typography } from "../../../../../theme/colors"
import { Text } from "../../../../components/Text"
```

Then the user can manually update it if it bother them

joshuayoes:
I think the alternative solution here would be to add typescript alias paths to ignite

frankcalise:
> I think the alternative solution here would be to add typescript alias paths to ignite

Ah right, which I did in a cookbook recipe. Good thought also.

kateinkim:
@frankcalise Sorry I'm not feeling so great right now so I might be missing something but I'm not really clear on the distinction we're making here between `subdirectory` and `destinationDir` (example [here](https://github.com/infinitered/ignite/blob/master/boilerplate/ignite/templates/navigator/NAMENavigator.tsx.ejs#L2)). It seemed like https://github.com/infinitered/ignite/issues/1746 would be resolved with the existing `destinationDir` except maybe allowing it to be overwritten via CLI.

frankcalise:
> @frankcalise Sorry I'm not feeling so great right now so I might be missing something but I'm not really clear on the distinction we're making here between `subdirectory` and `destinationDir` (example [here](https://github.com/infinitered/ignite/blob/master/boilerplate/ignite/templates/navigator/NAMENavigator.tsx.ejs#L2)). It seemed like #1746 would be resolved with the existing `destinationDir` except maybe allowing it to be overwritten via CLI.

Don't be sorry @kateinkim! You are correct in that `destinationDir` does tell the generator where to stash the new file. From what I gathered in #1746 (and I could be interpreting incorrectly), is that since we were considering the CLI parameter in some fashion that it would be more dynamic then just changing `destinationDir` each time you were going to generate a component.

@jamonholmgren want to weigh in here?

joshuayoes:
I wrote another draft PR of an idea to make implementing this easier: https://github.com/infinitered/ignite/pull/2317

If we go with using a tool like `zx` with generators, this could potentially thrown in the ignite cookbook

joshuayoes (approved):
This looks good! I'm a fan of the `~/` alias.

boilerplate/babel.config.js
```diff
@@ -1,4 +1,13 @@
 const plugins = [
+  [
+    "babel-plugin-root-import",
+    {
+      root: __dirname,
+      rootPathPrefix: "~/",
```
lindboe, line 6: I believe from the Slack poll earlier we voted for the `@` symbol as the preferred symbol: https://infinitered.slack.com/archives/C0X0P5LP7/p1639524256122600?thread_ts=1639519937.114400&cid=C0X0P5LP7

I don't want to bikeshed too much over this, but I personally have a strong dislike for `~` as to me that indicates my user's home directory, to use it for anything else feels like overloading it

joshuayoes: Could even get crazy and do the `@/` path alias that NextJs likes: https://nextjs.org/docs/advanced-features/module-path-aliases

Screenshot 2023-03-14 at 3 02 24 PM

joshuayoes: It looks like from the poll the absolute path alias (`components/card`) and the `@/` alias (`@/components/card`) were split.

Personally, of the two, I would prefer the absolute path alias (`components/card`) because it has had a better failure mode for intellisense when the tsconfig.json isn't indexed or found for some reason.

However, `@/` is nice semantically because it's very clear that it's an alias and distinct from other types of aliases.

lindboe: I believe the two alternatives were `app/components/card` vs `@/components/card`.

joshuayoes: Right, that is correct

frankcalise: > I believe from the Slack poll earlier we voted for the `@` symbol as the preferred symbol: https://infinitered.slack.com/archives/C0X0P5LP7/p1639524256122600?thread_ts=1639519937.114400&cid=C0X0P5LP7
> 
> I don't want to bikeshed too much over this, but I personally have a strong dislike for `~` as to me that indicates my user's home directory, to use it for anything else feels confusing

Oh I was going to say how did I miss this poll, but it was before I was around hah! I get that `~` does look like the user home directory. I'll update it to `@/` I think? just plain `components/card` doesn't given indication in any way that it is an alias

frankcalise: I set up the assets directory too seeing as we reference that one quite a bit as well for icons, images

joshuayoes: Nice addition!

boilerplate/tsconfig.json
```diff
@@ -14,7 +14,12 @@
     "target": "esnext",
     "lib": ["esnext", "dom"],
     "skipLibCheck": true,
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "paths": {
+      "app/*": ["app/*"],
+      "assets/*": ["assets/*"]
+    },
+    "baseUrl": "./"
```
lindboe, line 22: I thought maybe a baseUrl of "./" would be enough to achieve this? According to https://www.typescriptlang.org/tsconfig#baseUrl, looks like it should be.

boilerplate/babel.config.js
```diff
@@ -1,4 +1,19 @@
 const plugins = [
+  [
+    "babel-plugin-root-import",
```
lindboe, line 3: Just so I understand, what makes it so we have to configure babel like this in addition to tsconfig? From discussion in slack it had sounded like tsconfig alone might be enough but I never looked closely

frankcalise: Babel has to be configured to tell metro about how this magical path works.

The tsconfig paths alone will make things look like the path works as you just edit syntax, but when running metro won't know about that new path mapping.

boilerplate/tsconfig.json
frankcalise: I will double check!

frankcalise: Yeah you're right, pushing updates

jamonholmgren (approved):
Looks good to me!

PR #2297, opened 2022-11-11
Fix typo in DemoTextField from "props" to "prop"

KrastanD:

- [✅] `yarn test` **jest** tests pass with new tests, if relevant
- [✅] `README.md` has been updated with your changes, if relevant

## Describe your PR
In the DemoShowroomScreen, in the DemoTextField, in the Left and Right Accessory it says "This props". Clearly, it was intended to say "This prop" as it is referring to an individual prop.


jamonholmgren (approved):
Good catch @KrastanD !

Issue #2300, opened 2022-11-14
Demo with real use of Header


ryskin:
That's great Header comp is included in the library but there is no example of how to use it "right". For example, I usually include it in the Screen component but there are some problems with the insets. Maybe there is a better way to use it? I want to know how infinitered integrates Header in projects )

Thanks for all you do for the community!

ryskin:
@yulolimum you are the best, I will follow your recommendation, you just made my day

jamonholmgren:
If you want to use @yulolimum's suggestion 2, here's a small convenience hook:

```tsx
import React from "react"
import { useNavigation } from "@react-navigation/native"
import { useLayoutEffect } from "react"
import { Header, HeaderProps } from "../components"

/**
 * A hook to set the header for a screen.
 */
export function useHeader(headerProps: HeaderProps, deps: any[] = []) {
  const navigation = useNavigation()

  useLayoutEffect(() => {
    navigation.setOptions({
      headerShown: true,
      header: () => <Header {...headerProps} />,
    })
  }, deps)
}
```

You use it like this in your screen:

```tsx
  useHeader({
    rightTx: "common.logOut",
    onRightPress: logout,
  })
```

We could, perhaps, include it in Ignite if @yulolimum likes this approach.

ryskin:
@jamonholmgren thanks for the addition, I will use it

I think need to include both variants in the demo, so the user can choice.

ryskin:
Just tried and it's so simple and works great, hooks are my choice. I can make a different file for option if I need to reuse it many times.

yulolimum:
@ryskin 

Documentation added to https://github.com/infinitered/ignite/blob/master/docs/Components-Header.md#usage

`useHeader()` hook added: https://github.com/infinitered/ignite/blob/master/docs/Utils-useHeader.md

[Header updated](https://github.com/infinitered/ignite/blob/master/boilerplate/app/components/Header.tsx#L168) to use insets instead of SafeArea using new hook: https://github.com/infinitered/ignite/blob/master/docs/Utils-useSafeAreaInsetsStyle.md

Thx!

PR #2305, opened 2022-11-14
Adds a useHeader hook for setting a header in a screen

jamonholmgren:
In #2300, there was a discussion about the best way to use the `Header` component. @yulolimum's suggestion was to use the `navigation.setOptions` feature to set your header since that gives you maximum control over the header. This adds a convenience hook to make that a little easier and more understandable.

(If we decide this doesn't belong in Ignite, that's okay -- we can add it to the [Ignite Cookbook](https://infinitered.github.io/ignite-cookbook/).)

### DIFF
--- a/boilerplate/app/screens/WelcomeScreen.tsx
+++ b/boilerplate/app/screens/WelcomeScreen.tsx
@@ -1,19 +1,16 @@
 import { observer } from "mobx-react-lite"
-import React, {
-  FC,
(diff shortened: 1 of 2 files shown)


frankcalise:
I like this, just brought it over into Chain React App and got good use out of it!

yulolimum:
Closes in favor of #2307

PR #2306, opened 2022-11-16
fix(cli): move rename before install deps

frankcalise:

## Describe your PR
- Moves renaming of the react-native application to before dependencies are installed
- A bug was masked with `npx pod-install` being run after the rename that helped fix things, so now that we moved it to postInstall, rename must occur first
- This is only relevant to `yarn ios`

### DIFF
--- a/src/commands/new.ts
+++ b/src/commands/new.ts
@@ -481,18 +481,6 @@ export default {
       stopSpinner(msg, "📦")
     }
 
-    const shouldFreshInstallDeps = installDeps && shouldUseCache === false
-    if (shouldFreshInstallDeps) {
-      const unboxingMessage = `Installing ${packagerName} dependencies (wow these are heavy)`
-      startSpinner(unboxingMessage)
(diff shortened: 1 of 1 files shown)


mazenchami (approved):
LGTM when we looked over it earlier on zoom! 🎉 🚀

frankcalise:
@joshuayoes It's not with `npx pod-install`, basically if you did `yarn install`, `pod install` (or `npx pod-install`) and then ran the `rename` function, if you just did `yarn ios` after that, the app won't run.

This was masked before as we had:
- yarn install
- pod install
- rename
- npx pod install

Now that it is
- yarn install
- npx pod install
- rename

`yarn ios` didn't like something. It isn't until `npx pod-install` is run again that `yarn ios` succeeds.

Not sure if that means we need to look at something in our rename or our basic react-native ios template.

joshuayoes:
Yeah that probably means that rename is touching something that pod-install modifies. Probably worth investigating at some point but out of scope for this PR

PR #2315, opened 2022-11-18
Simplifies Toggle's switch-variant to not require measuring logic. Adds icon customization to Toggle's checkbox-variant.

yulolimum:

## Describe your PR

@robinheinze reported that the app she is working on has a weird screen offset when the Toggle is used. The reason for that is this issue: https://github.com/software-mansion/react-native-reanimated/issues/3368

The Toggle switch-variant uses reanimated's `FadeIn`. This was added to give the switch time to calculate the knob width to determine its position along the track. This was needed in case the component overrides specified a percentage width for the knob. 

I think it's reasonable to only allow point values for knob width customization. This cleans up the switch code and remove the need for measuring, which in effect, removes the need for the FadeIn.

This PR also separates the Types for the different Toggle variants for better prop hover descriptions and prop matches. 

This PR also adds the ability to customize the checkbox-variant icon.

### DIFF
--- a/boilerplate/app/components/Toggle.tsx
+++ b/boilerplate/app/components/Toggle.tsx
@@ -376,13 +376,15 @@ function Switch(props: ToggleInputProps) {
     detailStyle: $detailStyleOverride,
   } = props
 
-  const knobWidth = [$detailStyleOverride?.width, $switchDetail?.width, 24].filter(
-    (v) => typeof v === "number",
-  )[0]
+  const knobSizeFallback = 2
 
-  const knobHeight = [$detailStyleOverride?.height, $switchDetail?.height, 24].filter(
+  const knobWidth = [$detailStyleOverride?.width, $switchDetail?.width, knobSizeFallback].find(
     (v) => typeof v === "number",
-  )[0]
+  )
+
+  const knobHeight = [$detailStyleOverride?.height, $switchDetail?.height, knobSizeFallback].find(
+    (v) => typeof v === "number",
+  )
 
   const offBackgroundColor = [
     disabled && colors.palette.neutral400,
--- a/boilerplate/app/components/Toggle.tsx
+++ b/boilerplate/app/components/Toggle.tsx
@@ -376,8 +376,13 @@ function Switch(props: ToggleInputProps) {
     detailStyle: $detailStyleOverride,
   } = props
 
-  const knobWidth = $detailStyleOverride?.width ?? $switchDetail?.width ?? 24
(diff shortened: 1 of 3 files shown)


frankcalise (approved):
Looks good to me, added a couple of minor questions

yulolimum:
> added a couple of minor questions

Hey boss, I dont see any comments.

boilerplate/app/components/Toggle.tsx
```diff
@@ -346,15 +376,13 @@ function Switch(props: ToggleInputProps) {
     detailStyle: $detailStyleOverride,
   } = props
 
-  const knob = useRef<Animated.View>()
-  const [renderedKnobWidth, setRenderedKnobWidth] = useState(null)
+  const knobWidth = [$detailStyleOverride?.width, $switchDetail?.width, 24].filter(
```
frankcalise, line 379: Do we want to assign `24` to something like `defaultSize` for readability?

boilerplate/app/components/Toggle.tsx
```diff
@@ -12,20 +12,20 @@ import {
...
 
-export interface ToggleProps extends Omit<TouchableOpacityProps, "style"> {
+interface BaseToggleProps extends Omit<TouchableOpacityProps, "style"> {
   /**
    * The variant of the toggle.
    * Options: "checkbox", "switch", "radio"
    * Default: "checkbox"
    */
-  variant?: Variants
+  variant?: unknown
```
frankcalise, line 28: Can this go away since provided in each of the 3 types making up the discriminating union? Genuinely asking for my own understanding as I was trying to write some dynamic props earlier in the week

frankcalise:
> > added a couple of minor questions
> 
> Hey boss, I dont see any comments.

Weird, I must have not hit Start a review before - I added them back in

boilerplate/app/components/Toggle.tsx
yulolimum: I have this here for the description mainly. If I add descriptions to the 3 Types, then the documentation for the selected discriminating prop will be used.

boilerplate/app/components/Toggle.tsx
yulolimum: Good idea 👍🏻

boilerplate/app/components/Toggle.tsx
frankcalise: Oh ok cool thanks

PR #2317, opened 2022-11-19
feat(bin): add zx based generators

joshuayoes:

## Describe your PR

[There is a scripting tool called zx](https://github.com/google/zx) that is a hybrid between bash and node scripting. It has a lot of dev experience and productivity of gluegun. 

To play around with it, I decided to implement some of the existing generators in `ignite`. This was relatively simple to do. And [with this extension](https://marketplace.visualstudio.com/items?itemName=zjcompt.es6-string-javascript), you can even get syntax highlighting with the js strings.

This may be a nice alternative API for generators so that ignite users can update these scripts as a project grows.

Initially, I thought about creating an [alternative typescript generator API that I create a proof of concept of](https://codesandbox.io/s/ignite-generator-rfc-pgmbux?file=/src/generate/component.ts). But the nice thing about using `zx` is that it would pretty much remove any maintenance cost with generators from ignite. Additionally, these scripts are portable, so they can dropped into existing projects.

I will leave this PR here for discussion and an artifact of that exploration

### DIFF
--- a/boilerplate/bin/generate/component.mjs
+++ b/boilerplate/bin/generate/component.mjs
@@ -1,0 +1,53 @@
+#!/usr/bin/env zx
+import "zx/globals"
+import { prettier } from "../tools/string.mjs"
+import { dir } from "../tools/path.mjs"
+import { update } from "../tools/patch.mjs"
+
+const name = await question("Component name? ")
+
+const file = prettier(
+  /*ts*/ `
+import * as React from "react"
+import { StyleProp, TextStyle, View, ViewStyle } from "react-native"
+import { observer } from "mobx-react-lite"
+import { colors, typography } from "../theme"
+import { Text } from "./Text"
+
+export interface NameProps {
+  /**
+   * An optional style override useful for padding & margin.
+   */
+  style?: StyleProp<ViewStyle>
+}
+
+/**
+ * Describe your Name here
+ */
+export const Name = observer(function Name(props: NameProps) {
+  const { style } = props
+  const $styles = [$container, style]
+
+  return (
+    <View style={$styles}>
+      <Text style={$text}>Hello</Text>
+    </View>
+  )
+})
+
+const $container: ViewStyle = {
+  justifyContent: "center",
+}
+
+const $text: TextStyle = {
+  fontFamily: typography.primary.normal,
+  fontSize: 14,
+  color: colors.palette.primary500,
+}
+`,
+  { Name: name },
+)
+
+await fs.writeFile(dir.components(`${name}.tsx`), file)
+
+await update(dir.components("index.ts"), (file) => file + `export * from "./${name}"` + "\n")
--- a/boilerplate/bin/generate/component.mjs
+++ b/boilerplate/bin/generate/component.mjs
@@ -5,6 +5,8 @@ import { dir } from "../tools/path.mjs"
 import { update } from "../tools/patch.mjs"
 
 const name = await question("Component name? ")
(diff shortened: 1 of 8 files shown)


boilerplate/bin/generate/component.mjs
```diff
@@ -0,0 +1,55 @@
...
+  justifyContent: "center",
+}
+
+const $text: TextStyle = {
+  fontFamily: typography.primary.normal,
+  fontSize: 14,
+  color: colors.palette.primary500,
+}
+`,
+  { Name: name },
```
joshuayoes, line 50: Instead of using template literals for file strings, I decided to go with doing a regex replace. 

This was for a few reasons:
- using template literal syntax breaks [syntax highlighting from this extension](https://marketplace.visualstudio.com/items?itemName=zjcompt.es6-string-javascript)
- I find it's hard to read

A trade off is that it's not immediately obvious which strings are variables in a template string. But I still find it to be pretty readable, and one can always use Ctrl+F to see what is getting find and replaced

boilerplate/bin/tools/patch.mjs
```diff
@@ -0,0 +1,26 @@
...
+ * @param replacement {Replacement} callback to construct to replacement string
+ * @returns {string} string with the replacement made
+ */
+export const insert = (file, search, replacement) => file.replace(search, replacement(search))
+
+/**
+ * @callback UpdateCallback
+ * @param file {string} contents of the file
+ * @returns {string} updated file contents
+ */
```
joshuayoes, line 15: `zx` can be used with typescript, but it seems to take a lot wrangling to set it up. However, the existing globals have nice type defs already. And we can use jsdoc to create types otherwise. Another approach here may be to add `.d.ts` files and import them using `jsdoc`

joshuayoes:
This PR does not have the splash-screen or app-icon generator implemented, but I wanted to get some feedback about the reaction of this approach before implementing more sophisticated scripts like that

joshuayoes:
Other interesting `zx` features of note:
- [You can write these scripts as markdown files.](https://github.com/google/zx#markdown-scripts)
- [You can execute remote scripts from a URL. So potentially cookbook recipes could be executed](https://github.com/google/zx#executing-remote-scripts)
- Being able to import and export js modules means users can write their own "tools" as they need to in ignite projects.

Ashraf-Ali-aa:
@joshuayoes would something like https://github.com/plopjs/plop make more sense

joshuayoes:
> @joshuayoes would something like https://github.com/plopjs/plop make more sense 

This is very interesting, I'll need to take a deeper look at this. But at first glance, this seems like a more mature solution. 

The idea of this PR is that generators are something that should be co-located in each project and should be able to evolve with it. I have found many projects with generators from frameworks eventually abandoning them because a team adopted a style that was not compatible with it.

jamonholmgren:
@joshuayoes I think, unless there's a compelling reason to move, I'd like to keep our current generators. I see the benefits of what you're proposing here but the current ones work okay and that way we reduce churn.

If we used this on a real project for a while and people within IR liked it a lot, then I could see moving this direction. But I'm not seeing a lot of demand for this sort of change.

I do appreciate what you did here, though! You put a lot of thought into it, clearly.

frankcalise:
So I will also add to this that our current generator breaks down at the screen level when using expo-router, since you'll have to specify a path (or it would make sense to)

Right now we'd have to keep updating the cli and publishing to add parameters

PR #2337, opened 2022-12-12
fix: list item component to take children and add const assertion to colors

kateinkim:
## Describe your PR

- Adds const assertions to the color palette for better types.

CleanShot 2022-12-12 at 11 32 39@2x

- When one of our clients was trying to add a nested view to the `ListItem` component, they had issues since the ListItem component was wrapping the children with the Text component with default styling which was messing up the alignment of the list. This PR gives the component more flexibility when it needs to take children without tx or text. Recreated the issue below and shows after the fix: 

|Case A|Case B|Fix|
|-|-|-|
|CleanShot 2022-12-12 at 10 57 22@2x|CleanShot 2022-12-12 at 10 57 48@2x|CleanShot 2022-12-12 at 10 54 31@2x|

### DIFF
--- a/boilerplate/app/components/ListItem.tsx
+++ b/boilerplate/app/components/ListItem.tsx
@@ -144,9 +144,13 @@ export function ListItem(props: ListItemProps) {
           Component={LeftComponent}
         />
 
-        <Text {...TextProps} tx={tx} text={text} txOptions={txOptions} style={$textStyles}>
-          {children}
-        </Text>
+        {!!text || !!tx || typeof children === "string" ? (
+          <Text {...TextProps} tx={tx} text={text} txOptions={txOptions} style={$textStyles}>
+            {children}
+          </Text>
+        ) : (
+          children
+        )}
 
         <ListItemAction
           side="right"
--- a/boilerplate/app/theme/colors.ts
+++ b/boilerplate/app/theme/colors.ts
@@ -35,7 +35,7 @@ const palette = {
 
   overlay20: "rgba(25, 16, 21, 0.2)",
   overlay50: "rgba(25, 16, 21, 0.5)",
-}
+} as const
 
 export const colors = {
   /**


boilerplate/app/components/ListItem.tsx
```diff
@@ -144,9 +144,13 @@ export function ListItem(props: ListItemProps) {
...
 
-        <Text {...TextProps} tx={tx} text={text} txOptions={txOptions} style={$textStyles}>
-          {children}
-        </Text>
+        {!!text || !!tx || typeof children === "string" ? (
+          <Text {...TextProps} tx={tx} text={text} txOptions={txOptions} style={$textStyles}>
+            {children}
+          </Text>
+        ) : (
+          children
```
frankcalise, line 152: Looks good, do we need any View styling around this or should we just let the user take care of that with what they're passing in?

$textStyles has some styling associated with it that we'd be missing here.

If we do add that wrapper though, then we'd need some sort of container override, so maybe better not to.

Thoughts?

frankcalise (commented):
Nice work, left a comment for consideration!

boilerplate/app/components/ListItem.tsx
kateinkim: I think it makes more sense to just let the user style the children directly. The original issue the client had with this component was that `$textStyles` unintuitively styled the children being passed in so they were trying all sorts of different styles and getting unexpected results. I have a low gafo on this but I think adding additional styling would just result in the same issue.

frankcalise: @kateinkim Sounds good to me!

yulolimum: There's a potential deprecation here with not allowing to pass other Text children into the <Text /> (Text's are meant to be nestable here). I'm wondering if there's a better way to check what type `children` is without adding another prop... Thinking...

kateinkim: I can just leave this PR for adding const assertion to the color palette and open an issue for the ListItem?

frankcalise:
@kateinkim do you want to remove the ListItem change for now and leave the color change? I'm going to publish an update with a few other changes so we can lump this one in with those

kateinkim:
> @kateinkim do you want to remove the ListItem change for now and leave the color change? I'm going to publish an update with a few other changes so we can lump this one in with those

Yep! sounds good 👍🏼

frankcalise (approved):

Issue #2223, opened 2022-09-27
Ensure that password isn't stored in AsyncStorage by default


jamonholmgren:
In the AuthenticationStore, the password prop is a normal text type: 

https://github.com/infinitered/ignite/blob/01c2a11957b37303441fd4abbc4350d9e5d6159a/boilerplate/app/models/AuthenticationStore.ts#L8

Which means it's persisted into AsyncStorage as plain text:

https://github.com/infinitered/ignite/blob/7832f96b0dbb973d15d68f213dfe4b67d66381d7/boilerplate/app/models/helpers/setupRootStore.ts#L43

We should make sure it's not. I don't know for sure that it's a major problem, but it doesn't feel right.

jamonholmgren:
While we're in there, any `types.optional(types.string, "")` could be shortened to just an empty string, which is the same thing.

```diff
- authPassword: types.optional(types.string, ""), 
+ authPassword: "", 
```

joshuayoes:
I know that on the web, there isn't really a way to store passwords or secure tokens client side. The only way to do so is through a cookie, but that is because client side javascript cannot access a cookie.

The alternative is storing a refresh token, which can be used to make a request to the server to get an access token and make auth requests. 

Is AsyncStorage accessible by other apps?

joshuayoes:
Discussion here is relevant: https://reactnative.dev/docs/security#storing-sensitive-info

kateinkim:
Just opened a PR to address this. Do we want to add a way to store sensitive info like @joshuayoes mentioned? We used https://docs.expo.dev/versions/latest/sdk/securestore/ in our client project. I can open a separate ticket to file this as an issue if so. @jamonholmgren

lodev09:
I checked the PR for this but it doesn't really solve the issue. Correct me if I'm wrong but I think `preProcessSnapshot` is called after a snapshot is loaded into an instance. This means that the auth credentials were already in the storage to begin with. The PR will only do harm than good for the demo since it will present the login screen every time you close & open the app.

I believe the solution is to clear these "before saving" into storage. Or maybe refactor the `AuthenticationStore` model and move it out of the `RootStore` then use different storing strategy for it?

kateinkim:
@lodev09 I am very sorry for the late reply. IR took a week off over the holidays and just realized that I should have replied back to you sooner!

[...]

PR #2338, opened 2022-12-13
fix: do not store password  in AsyncStorage by default

kateinkim:
## Describe your PR

- Addresses issue #2223

### DIFF
--- a/boilerplate/app/models/AuthenticationStore.ts
+++ b/boilerplate/app/models/AuthenticationStore.ts
@@ -45,8 +45,13 @@ export const AuthenticationStoreModel = types
     },
   }))
   .preProcessSnapshot((snapshot) => {
-    // remove sensitive data from snapshot
+    // remove sensitive data from snapshot to avoid secrets
+    // being stored in AsyncStorage in plain text if backing up store
     const { authToken, authPassword, ...rest } = snapshot // eslint-disable-line @typescript-eslint/no-unused-vars
+    
+    // see the following for strategies to consider storing secrets on device
+    // https://reactnative.dev/docs/security#storing-sensitive-info
+    
     return rest
   })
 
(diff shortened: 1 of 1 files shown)


boilerplate/app/models/AuthenticationStore.ts
```diff
@@ -44,6 +44,11 @@ export const AuthenticationStoreModel = types
       store.authPassword = ""
     },
   }))
+  .preProcessSnapshot((snapshot) => {
+    // remove sensitive data from snapshot
+    const { authToken, authPassword, ...rest } = snapshot // eslint-disable-line @typescript-eslint/no-unused-vars
+    return rest
+  })
```
joshuayoes, line 51: I'm not as familiar with mobx, but won't this completely remove `authToken` and `authPassword` from being able to be used on the model?

kateinkim: I'm not that familiar with mobx yet either but from [the docs](https://mobx-state-tree.js.org/overview/hooks):

> Before creating an instance or applying a snapshot to an existing instance, this hook is called to give the option to transform the snapshot before it is applied. The hook should be a pure function that returns a new snapshot. This can be useful to do some data conversion, enrichment, property renames, etc. This hook is not called for individual property updates. 

So I believe it just removes it from the snapshot that's being applied/generated and it will still work on the instance. I'm actually wondering if it shouldn't be `postProcessSnapshot` now. Maybe @jamonholmgren can chime in?

frankcalise: I think @yulolimum mentioned this method, see https://infinitered.slack.com/archives/C0X0P5LP7/p1666302551358189?thread_ts=1666302363.164549&cid=C0X0P5LP7

joshuayoes: That works for me then!

joshuayoes: I think it would be worth adding a more specific comment to cue the developer to think about what they should do on their project

```suggestion
  .preProcessSnapshot((snapshot) => {
    // remove sensitive data from snapshot to avoid secrets
    // being stored in AsyncStorage in plain text if backing up store
    const { authToken, authPassword, ...rest } = snapshot // eslint-disable-line @typescript-eslint/no-unused-vars
    
    // see the following for strategies to consider storing secrets on device
    // https://reactnative.dev/docs/security#storing-sensitive-info
    
    return rest
  })
```

joshuayoes (approved):

PR #2360, opened 2023-01-19
refactor(boilerplate): remove authPassword from store

lodev09:

## Describe your PR

Following up on the issue in https://github.com/infinitered/ignite/issues/2223
- Removes `authPassword` from `authenticationStore` for security reasons. Simple but effective.
- Fixes regression issue on Login Screen that always prompts after refresh. See https://github.com/infinitered/ignite/pull/2338

_Note that this also removes password validation, which I think is normal -- we usually validate passwords during sign up and not during sign in. Please advise if we really need to validate password at login, I can modify this PR._


kateinkim (approved):
Thank you so much for working on this! 

@yulolimum I'm gonna go ahead and approve this PR.

Issue #2354, opened 2023-01-12
The command `ignite-cli generate model` overwrites the existing model files


hamzakat:
Hi everyone.

Today I executed the handy command `ignite-cli generate model MODEL_NAME` and used an existing model's name, and that's resulted in a direct overwriting, so the codes in the file gone! Luckily, there weren't a lot of code lines so it wasn't a frustrating loss. However, it would be very helpful to implement a check before executing the boilerplate magical generation, and I would be happy to help implementing that if anyone provides a hint!

Thanks in advance 🔥

jamonholmgren:
Oof, yes, that isn't good!

Right now, the output looks like this:

<img width="757" alt="CleanShot 2023-01-13 at 10 47 21@2x" src="https://user-images.githubusercontent.com/1479215/212396075-bff23b1c-2a45-444e-ae0b-d9cd6a587842.png">

I would implement a check (around [here](https://github.com/infinitered/ignite/blob/cd50d094190ead6a6c00739bbb4e505a6d03be64/src/tools/generators.ts#L278-L284)) for `filesystem.exists(...)`, and if it does, output something like this:

```
npx ignite-cli g model Dup


   Generated new files:
   <none>
   
   Skipped these files because they already exist:
   /Users/jh/Code/Temp/GenDouble/app/models/Dup.test.ts
   /Users/jh/Code/Temp/GenDouble/app/models/Dup.ts

  To overwrite these files, run the command again with the `--overwrite` flag
```

This will take a new [GeneratorOptions](https://github.com/infinitered/ignite/blob/cd50d094190ead6a6c00739bbb4e505a6d03be64/src/tools/generators.ts#L216-L219) flag, `overwrite: boolean`.

You will also probably need to change the default return value (`Promise<string>`) to something like `Promise<{ written: string[], overwritten: string[], exists: string[] }>` so you can output the right info at the end.

Thoughts?

frankcalise:
Closed by #2363

PR #2363, opened 2023-01-28
enhancement(cli) : add overwrite option when using generate command

aymane-missouri-nw:

## Describe your PR
 - For this issue https://github.com/infinitered/ignite/issues/2354
 - Add possibility to overwrite or not files when using command generate
 - Don't remove files that exists when using command generate  unless they specifiy the property --overwrite in it.

### DIFF
--- a/src/commands/generate.ts
+++ b/src/commands/generate.ts
@@ -52,8 +52,11 @@ async function generate(toolbox: GluegunToolbox) {
   })
 
   heading(`Generated new files:`)
+
   if (exists.length > 0) {
-    p(`<none>`)
+    if (written.length > 0) {
+      written.forEach((f) => p(f))
+    } else p(`<none>`)
     p()
     heading(`Skipped these files because they already exist:`)
     exists.forEach((f) => p(f))
--- a/src/commands/generate.ts
+++ b/src/commands/generate.ts
@@ -1,6 +1,8 @@
 import { GluegunToolbox } from "gluegun"
+import { boolFlag } from "../tools/flag"
 import { generateFromTemplate, runGenerator } from "../tools/generators"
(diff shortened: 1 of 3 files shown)


frankcalise:
@aymane-missouri-nw thanks for the contribution - could you please update the existing test to account for the new output and add a test case for the overwrite mechanism?

aymane-missouri-nw:
@frankcalise  I created tests for the overwrite mechanism and the tests passes locally but in pipelines they fail other tests , any thoughts ?

frankcalise:
Ok we'll look into it, thanks

aymane-missouri-nw:
@frankcalise I just update the branch with current master and test passes now

frankcalise:
@aymane-missouri-nw this looks great!

One improvement I'm wondering if we can try to address, let me know your thoughts.

When a file does exist and the overwritten flag is not passed (so for example, the Screen file doesn't get written), can we avoid any further patches being applied? Does it make sense to do this?

The patch to the `index.ts` files really doesn't have to occur again, not sure we can make that smarter. Or if we should assume that the export is there? Since the file exists, likely that export exists too, so we shouldn't end up with

**app/screens/index.ts**

```tsx
export * from "./OtherScreen"
export * from "./ExistingScreen"
export * from "./ExistingScreen"
```

aymane-missouri-nw:
@frankcalise  you're right i forgot this edge case , i will adress it and i think we should assume that the export is there because the file is already exist

aymane-missouri-nw:
@frankcalise i pushed a commit for handling patch , let me know your thoughts

aymane-missouri-nw:
@frankcalise i have handle this case , can u recheck and give me your thoughts ?

TheMissouri:
@frankcalise any news ?

PR #2372, opened 2023-02-07
fix: sync with latest expo

jr00n:
When trying to build and run a dev-client with latest Ignite, the dev-client crashed.
`EXDevLauncher/ExpoDevLauncherReactDelegateHandler.swift:91: Fatal error: Unexpectedly found nil while unwrapping an Optional value`

Looking at Expo doc's seems like AppDelegate now needs te be different. Tested it, and works no with latest dev-client 

https://github.com/expo/expo/blob/sdk-47/templates/expo-template-bare-minimum/ios/HelloWorld/AppDelegate.mm#L49

### DIFF
--- a/boilerplate/ios/HelloWorld/AppDelegate.mm
+++ b/boilerplate/ios/HelloWorld/AppDelegate.mm
@@ -45,7 +45,7 @@ static NSString *const kRNConcurrentRoot = @"concurrentRoot";
 #endif
 
   NSDictionary *initProps = [self prepareInitialProps];
-  UIView *rootView = RCTAppSetupDefaultRootView(bridge, @"HelloWorld", initProps);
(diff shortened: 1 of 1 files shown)


frankcalise:
Hey @jr00n thanks for the contribution! Can you tell us more in the PR notes about the steps you were trying to execute that lead to this change? Would want to test a before and after.

Thanks!

frankcalise:
Also if you're building a dev client, it's likely you should be deleting the android/ios dirs. 

See this for reference: https://ignitecookbook.com/docs/recipes/PristineExpoProject

frankcalise:
@jr00n Please see this comment: https://github.com/infinitered/ignite/pull/2372#issuecomment-1422532215

PR #2434, opened 2023-05-10
Android Only - Navigation Bar Background Color

morganick:
This uses the colors of the app to set the background
color of the software navigation bar buttons for Android.

## Describe your PR
On Android, the navigation bar background color can be set to match the background color of the app. This PR uses the background color from `colors.background` and uses it as the background color of the navigation bar software buttons.


mazenchami (approved):
🚀 thanks for bringing this over to Ignite too!

frankcalise (approved):

PR #2449, opened 2023-05-25
fix(boilerplate): added node version check

### LINKED ISSUE #2448: Add Node version check
As reported by @kevinvangelder:

Ignite needs to do a Node version check at the beginning as if you have an incorrect node version selected it generates the project but fails part-way through the process, leaving demo code in place and failing to install NPM modules.

frankcalise:

## Describe your PR
- Closes #2448 
- Does a check for Node 18 or higher (see https://endoflife.date/nodejs for active branch details)
- Kept engine check to `package.json` so @jamonholmgren keeps me around

### DIFF
--- a/boilerplate/package.json
+++ b/boilerplate/package.json
@@ -118,6 +118,9 @@
     "@types/react": "^18",
     "@types/react-dom": "^18"
   },
+  "engines": {
+    "node": ">=18"
+  },
   "prettier": {
     "printWidth": 100,
     "semi": false,


frankcalise:
@joshuayoes We do have `devEngines` in the root package.json, however I can't find any documentation on that, I'm guessing they changed the key to `engines`. Should be safe to just rename this and update versions you think?

```
  "devEngines": {
    "node": ">=7.x",
    "npm": ">=4.x"
  },
```

joshuayoes:
`engines` is the only official npm key: https://docs.npmjs.com/cli/v9/configuring-npm/package-json#engines

It looks like `devEngines` may be in reference to an old fbjs library? https://github.com/npm/npm/issues/9765 I'd say it's safe to remove at this point

jamonholmgren:
Looks like we need to bump the circleci node version.

joshuayoes:
Seems the tests failing is related to this: https://github.com/facebook/react-native/issues/36440

PR #2454, opened 2023-05-31
chore: enable eslint react-hook/recommended rule set

guiccbr:

## Describe your PR

Enable eslint react-hook/recommended rule set, as recommended in the react dev website: https://react.dev/learn/editor-setup#linting

Related discussion: https://github.com/infinitered/ignite/discussions/2453

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
     "eslint-plugin-import": "^2.22.1",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-promise": "^5.1.0",
+    "eslint-plugin-react-hooks": "^4.6.0",
     "husky": "^7.0.1",
     "jest": "^27.0.6",
     "prettier": "2.8.8",
@@ -82,7 +83,8 @@
     "extends": [
       "prettier",
       "plugin:@typescript-eslint/recommended",
-      "standard"
+      "standard",
+      "plugin:react-hooks/recommended"
     ],
     "plugins": [
       "@typescript-eslint"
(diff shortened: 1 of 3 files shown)


src/commands/new.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react-hooks/rules-of-hooks */
```
guiccbr, line 1: This file has a method that starts with `use*` and react-hooks/rules-of-hooks is complaining about that.

I'm not sure if this is the best approach here.

package.json
```diff
@@ -68,6 +68,7 @@
     "eslint-plugin-import": "^2.22.1",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-promise": "^5.1.0",
+    "eslint-plugin-react-hooks": "^4.6.0",
```
joshuayoes, line 71: Can you move this package to [boilerplate/package.json](https://github.com/infinitered/ignite/blob/master/boilerplate/package.json) instead? 

This top level package.json is for the cli, where as the `boilerplate` folder is for the app content.

package.json
```diff
@@ -82,7 +83,8 @@
     "extends": [
       "prettier",
       "plugin:@typescript-eslint/recommended",
-      "standard"
+      "standard",
+      "plugin:react-hooks/recommended"
```
joshuayoes, line 87: Can you move this package to [boilerplate/package.json](https://github.com/infinitered/ignite/blob/master/boilerplate/package.json) instead?

This top level package.json is for the cli, where as the boilerplate folder is for the app content.

joshuayoes (changes requested):
Made some comments to move some of the linting settings, but otherwise looks good so far

guiccbr:
Thanks. I'll close and reopen because after enabling the the linter failed for the generated app. I'll need some more time to fix that.

PR #2457, opened 2023-06-05
feat: full type inference to `useSafeAreaInsetsStyle`

sergiocarneiro:
The type of the value returned by `useSafeAreaInsetsStyle` is always:
```ts
{
  paddingTop?: string | number;
  paddingBottom?: string | number;
  paddingStart?: string | number;
  paddingEnd?: string | number;
  marginTop?: string | number;
  marginBottom?: string | number;
  marginStart?: string | number;
  marginEnd?: string | number;
}
```

Not only is inaccurate, but also wrong. `string` is never returned by `useSafeAreaInsets`, nor `undefined`.

This broad type makes it harder to perform operations with the safe-area values, and it's uncertain whether developers can force-cast to `number` without looking into the source code of `react-native-safe-area-context`.

By adding a little complexity to the hook, full type inference can be added.

### Result
```ts
const $insets = useSafeAreaInsetsStyle(["bottom"]);
// { paddingBottom: number }

const $insets = useSafeAreaInsetsStyle(["top"], "margin");
// { marginTop: number }

const $insets = useSafeAreaInsetsStyle(["left", "right"], "padding");
// { paddingLeft: number; paddingRight: number }
```

### Caveats

The type displayed _while hovering_ the property is not simplified, e.g.:
```ts
SafeAreaInsetsStyle<"padding", ("right" | "left")[]>
```

<details>
<summary>Workaround</summary>

We could introduce an extra type utility `Pretty`:
```ts
type Pretty<T> = {
  [K in keyof T]: T[K]
}

export type SafeAreaInsetsStyle<
  Property extends "padding" | "margin" = "padding",
  Edges extends Array<ExtendedEdge> = Array<ExtendedEdge>
> = Pretty<{
  [K in Edges[number] as `${Property}${Capitalize<K>}`]: number;
}>;
```

#### Result
```ts
Pretty<{
  paddingBottom: number;
}>
```
</details>


markrickert:
Hey @sergiocarneiro I just tried this out and i was getting a typescript warning with your changes on the latest version of ignite. 

I had to change `safeAreaEdges: Edges = [] as Edges,` to `safeAreaEdges: Edges = [] as unknown as Edges,` to get typescript 5.3.2 to pass.

Could you rebase on master and have another look at this change?

PR #2465, opened 2023-06-16
fix: simplify Reactotron config into one file

joshuayoes:

## Describe your PR

Our current Reactotron config is current spread across multiple files. It's nice to have well named functions, it doesn't match the [convention in the Reactotron docs](https://github.com/infinitered/reactotron/blob/master/docs/quick-start-react-native.md#configure-reactotron-with-your-project) where there is one `ReactotronConfig` file that sets everything up.

Summary of changes:
- moves all of the configuration into one file and flattens the conditional logic. 
- wraps any calls to `console.tron` in a `__DEV__` check, which will allow metro to automatically strip out any `console.tron` calls in production.
- adds monkey patch to console.log, console.error, etc to route them both through logbox in the terminal and Reactotron.


boilerplate/app/app.tsx
```diff
@@ -9,6 +10,9 @@
  * The app navigation resides in ./app/navigators, so head over there
  * if you're interested in adding screens and navigators.
  */
+if (__DEV__) {
+  require("./services/reactotron")
```
frankcalise, line 14: ❤️

boilerplate/app/models/helpers/useStores.ts
```diff
@@ -57,11 +56,13 @@ export const useInitialRootStore = (callback: () => void | Promise<void>) => {
     let _unsubscribe
     ;(async () => {
       // set up the RootStore (returns the state restored from AsyncStorage)
-      const { restoredState, unsubscribe } = await setupRootStore(rootStore)
+      const { unsubscribe } = await setupRootStore(rootStore)
```
frankcalise, line 59: Is restoredState no longer needed?

boilerplate/app/services/reactotron/index.ts
```diff
@@ -1 +1,174 @@
...
+    /**
+     * Reactotron client for logging, displaying, measuring performance, and more.
+     * @see https://github.com/infinitered/reactotron
+     *
+     * @example
+     * if (__DEV__) {
+     *  console.tron.log("here")
+     *
+     *  console.tron.display({
+     *    name: 'JOKE',
```
frankcalise, line 128: This is 🥇

boilerplate/app/screens/DemoDebugScreen.tsx
```diff
@@ -22,17 +22,19 @@ export const DemoDebugScreen: FC<DemoTabScreenProps<"DemoDebug">> = function Dem
...
-        value: {
-          appId: Application.applicationId,
-          appName: Application.applicationName,
-          appVersion: Application.nativeApplicationVersion,
-          appBuildVersion: Application.nativeBuildVersion,
-          hermesEnabled: usingHermes,
-        },
-        important: true,
-      })
+      if (__DEV__) {
```
frankcalise, line 25: Hm is this a regression in ease of use, though? Each time you'd want to console.tron.x we'd have to wrap it in this check?

frankcalise (commented):
Nice work, left some notes! Like the reorganization

boilerplate/app/models/helpers/useStores.ts
joshuayoes: Yeah, it was being used in the setupStore function to log the initial state to Reactotron: https://github.com/infinitered/ignite/pull/2465/files#diff-bcccc60bab0fb409a76223075a440e6b5f004da8e68734b92a8e4dbb0161bd0aL69

We could add this back, but I felt it was nice to reduce the noise a bit

boilerplate/app/screens/DemoDebugScreen.tsx
joshuayoes: Yeah, but according to the docs, Reactotron should really be disabled in production. So we need some mechanism to do that: https://github.com/infinitered/reactotron/blob/master/docs/tips.md#react-native

Ideally, it would be nice to have some developer tooling around this. There a babel plugin to strip Reactotron out of production builds: https://github.com/infinitered/babel-plugin-ignite-ignore-reactotron

We could also write an eslint rule

mazenchami: is there a way to wrap it under the hood in `__DEV__`? the babel plugin is helpful but i had never heard of it

joshuayoes: The alternative is that we can stub out `console.tron` calls like we do currently. We currently add all the keys manually, but I think a proxy like this would be more resilient:

```
// Make sure the console.tron object exists
console.tron = console.tron || {};

// Create a handler for the proxy
let handler = {
    get: function(target, prop, receiver) {
        // If the property accessed is a function, return a noop function
        if (typeof target[prop] === 'function') {
            return function() {}; // noop function
        }
        // If the property is not a function, just return it
        return Reflect.get(...arguments);
    }
};

// Create a Proxy around console.tron object
console.tron = new Proxy(console.tron, handler);
``` 

We'd need to add this into the `else` block in our App entry point instead of creating the Reactotron config.

joshuayoes: This is also something we could ship in `reactotron` so that way we could be like
```js
if (__DEV__) {
  require('ReactotronConfig')
} else {
  require('reactotron-react-native/console-tron-stub') // add a source file to the package with the above stub
}

```

boilerplate/app/services/reactotron/index.ts
```diff
@@ -1 +1,174 @@
-export * from "./reactotron"
```
jamonholmgren, line 1: Given there's only one file, I'd recommend moving this and renaming it.

```
boilerplate/app/services/reactotron.ts
```

In general, I dislike `index.ts` (ever had 12 `index.ts` files open in the same VS Code instance...?). And if we're only doing a single file then a folder is redundant.

jamonholmgren (approved):
I'm in favor!

boilerplate/app/services/reactotron/index.ts
joshuayoes: I can get down with this. 

I think I may actually name this to `ReactotronConfig` since that's the convention in the docs: https://github.com/infinitered/reactotron/blob/master/docs/quick-start-react-native.md#configure-reactotron-with-your-project

boilerplate/app/app.tsx
joshuayoes: Push up a commit with a comment for this:
https://github.com/infinitered/ignite/pull/2465/commits/1037d3ce290debf7e98cdbb203b443f51efe77c8

boilerplate/app/services/reactotron/index.ts
joshuayoes: Created a new folder: `devtools` and put it into `app/devtools/ReactotronConfig` with this: https://github.com/infinitered/ignite/pull/2465/commits/1037d3ce290debf7e98cdbb203b443f51efe77c8

boilerplate/app/screens/DemoDebugScreen.tsx
frankcalise: The proxy is pretty cool and getting it for free from reactotron seems like a cool win as well (and can always come later)

PR #2467, opened 2023-06-22
Fix: Error Boundary Infinite Loop

morganick:
Co-authored-by: yulolimum <yulian@infinite.red>

## Describe your PR

When you use the `ErrorBoundary` with the `catchErrors` prop set to `never`, this will cause the application to infinitely render. This is due to setting the `error` and `errorInfo` state, but not checking whether catch is enabled. This causes the render function to re-render the children that will then trigger another exception, the `componentShouldUpdate` passes because it is a new error, and start the process again.

Special thanks to @yulolimum for helping me come up with a simple solution.


frankcalise (approved):
Nice find!

PR #2479, opened 2023-07-12
Fix: Error while converting `null` to `RootStore`

midrizi:
Fixes the following issue:

[mobx-state-tree] Error while converting `null` to `RootStore`: value `null` is not assignable to type: `RootStore` (Value is not a plain object).


mazenchami (approved):
@midrizi this LGTM! Thank you very much for your contribution!

PR #2487, opened 2023-08-01
feat(boilerplate): config plugin scaffolding

frankcalise:

## Describe your PR
- Adds scaffolding for writing TypeScript Expo Config Plugins (via `ts-node`) within your project without needing to transpile to JS
- Adds documented Config Plugin, `withSplashScreen`, to address Android 12 issues with `expo-splash-screen`
- Updates some documentation related to Expo and Ignite

### DIFF
--- a/boilerplate/app.config.ts
+++ b/boilerplate/app.config.ts
@@ -1,0 +1,25 @@
+import { ExpoConfig, ConfigContext } from "@expo/config"
+
+/**
+ * Use ts-node here so we can use TypeScript for our Config Plugins
+ * and not have to compile them to JavaScript
+ */
+require("ts-node/register")
+
+/**
+ * @param config ExpoConfig coming from the static config app.json if it exists
+ * 
+ * You can read more about Expo's Configuration Resolution Rules here:
+ * https://docs.expo.dev/workflow/configuration/#configuration-resolution-rules
+ */
+module.exports = ({ config }: ConfigContext): Partial<ExpoConfig> => {
+  const existingPlugins = config.plugins ?? []
+
+  return {
+    ...config,
+    plugins: [
+      ...existingPlugins,
+      require("./plugins/withSplashScreen").withSplashScreen,
+    ],
+  }
+}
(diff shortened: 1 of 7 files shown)


boilerplate/app.config.ts
```diff
@@ -0,0 +1,25 @@
...
+require("ts-node/register")
+
+/**
+ * @param config ExpoConfig coming from the static config app.json if it exists
+ * 
+ * You can read more about Expo's Configuration Resolution Rules here:
+ * https://docs.expo.dev/workflow/configuration/#configuration-resolution-rules
+ */
+module.exports = ({ config }: ConfigContext): Partial<ExpoConfig> => {
+  const existingPlugins = config.plugins ?? []
```
frankcalise, line 16: @mazenchami TS didn't like when I just spread `...plugins` inline since it could be undefined - any other way you like achieving this or does this look fine?

mazenchami: this looks fine to me. no need in over complicating it with more checks

kateinkim (approved):
looks awesome!

keith-kurak:
@frankcalise This looks great to me! The v9/ Ex[press]o (ah, did I get that right?) explanation is spot on IMO. I'll try that config plugin in a bit and let you know if I see any issues.

Not part of this PR, but I just noticed that, since the **postinstall** script now more-or-less only runs patch-package, you may be able to move that directly into the NPM scripts `postinstall`, and then Ignite should be fully compatible with Windows/ Powershell (at least that was the case for my workshop when I did that with v8).

frankcalise:
> The v9/ Ex[press]o (ah, did I get that right?) explanation is spot on IMO. 

@keith-kurak haha yeah, `Expresso` is fine but I put brackets in the event people don't get the pun of `Expo` with `ress` injected in the middle 😀 (and it's quick and to the point, etc - I think it works nicely!)

Thanks for looking! Point noted on the post install as well.

frankcalise:
@joshuayoes flat and flexible enough for you? I think it's pretty straight forward

robinheinze (approved):
Impemented these plugin changes on our current client app and it works great! No more compiling to js!

joshuayoes:
@frankcalise sorry I didn't get a look at this sooner, this is great work

PR #2488, opened 2023-08-07
Correct cd command when targetPath has changed

bradherman:

## Describe your PR

The `cd` output on a new project uses the `projectName`, but you can override the default `targetPath`, making that command incorrect.  This changes to use the `targetPath` so you can copy/paste if you changed the path and it will work


frankcalise:
Good catch! Thanks for the contribution, much appreciated :)

PR #2497, opened 2023-09-11
feat(cli): add bun support

frankcalise:

## Describe your PR

- Adds support for the `bun` package manager 🎉 

image


morganick (approved):
Works as expected. I wonder if the commands are similar enough in all situations to where we could extract out a mapping of the commands for the package manager so the functions can be generic and the mapping can be updated per packager more easily.

morganick:
Appears that tests are failing for it.

jamonholmgren:
This PR for Gluegun will add better Bun support over there as well.

https://github.com/infinitered/gluegun/pull/783

jamonholmgren:
IMO, we should consider moving to Bun for local development of Ignite as well as CI (like I'm doing with Gluegun CLI). It'll speed up tests a _lot_.

frankcalise:
@jamonholmgren that looks like maybe you have `ios/` and `android/` dirs local left over from Ignite 8 dev. This branch is targeting Ignite 9 (although if we want to cherry pick it over we could)

jamonholmgren (approved):
Let's merge it!

PizzaBunApp
joshuayoes, line 1: What is this git submodule here?

frankcalise: Uh no idea, shouldn't be there. Good catch

frankcalise: Should be gone now

PR #2499, opened 2023-09-15
Remove dependency on strip-ansi and embed the tiny function instead

jamonholmgren:
This may not be appropriate to merge into `master`, but we should do this at some point to reduce our dependency footprint. Also: strip-ansi has been an outlier in that it was using require(...).

I don't like having it be a dependency when it's just a 20 line file.


frankcalise (approved):
Sounds good to me!

PR #2510, opened 2023-10-03
fix: `date-fns` import syntax adding almost ½ megabyte to production bundle sizes

markrickert:

Tests failed on my local machine, but the failure seems unrelated to the content I changed in this PR.

## Describe your PR

Because react-native doesn’t tree-shake, importing things from `date-fns` bundles ALL of the `date-fns` library, which is quite heavy.

Changing from the syntax `import { Locale, format, parseISO } from "date-fns”` to importing them individually like:

```ts
import type Locale from "date-fns/locale"
import format from "date-fns/format"
import parseISO from "date-fns/parseISO"
```

Saves us 0.41 megabytes in the production bundle size of an ignited app. That’s a HUGE savings for such a small syntax change.

I also added scripts to the boilerplate to use `react-native-bundle-visualizer` through `npx` (which is how i determined this bundle filesize savings).

Before this PR: `yarn bundle:visualize`: 5.13mb
After this PR: `yarn bundle:visualize`: 4.72mb

### DIFF
--- a/boilerplate/app/utils/formatDate.ts
+++ b/boilerplate/app/utils/formatDate.ts
@@ -1,6 +1,12 @@
-import { Locale, format, parseISO } from "date-fns"
 import I18n from "i18n-js"
 
+// Note the syntax of these imports from the date-fns library.
+// If you import with the syntax: import { format } from "date-fns" the ENTIRE library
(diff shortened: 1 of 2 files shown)


frankcalise (approved):
Looks good to me, thanks!

PR #2513, opened 2023-10-05
feat(boilerplate) `@shopify/flash-list` integration

markrickert:

## Describe your PR

This integrates [`@shopify/flash-list`](https://shopify.github.io/flash-list/) in all places where we used to have a react-native `FlatList` and updates the expo to compatibility object to use the correct expected version while using the latest for expo prebuild and bare workflows.

Once Ignite 9.0 is released, we should [delete the ignite-cookbook recipe for FlashList](https://github.com/infinitered/ignite-cookbook/pull/102).


src/commands/new.ts
```diff
@@ -620,9 +620,15 @@ module.exports = {
...
-        let appJsonRaw = read("app.json")
-        appJsonRaw = appJsonRaw.replace(/"newArchEnabled": false/g, '"newArchEnabled": true')
+        const appJsonRaw = read("app.json")
         const appJson = JSON.parse(appJsonRaw)
+        appJson.expo.plugins[1][1].ios.newArchEnabled = true
+        appJson.expo.plugins[1][1].android.newArchEnabled = true
+
+        // Adding the "deploymentTarget" key is required for
+        // @react-native-async-storage/async-storage to work in the new architecture
+        appJson.expo.plugins[1][1].ios.deploymentTarget = "13.4"
```
frankcalise, line 630: nice regex replace side step!

src/tools/expoGoCompatibility.ts
```diff
@@ -9,6 +9,7 @@ export const expoGoCompatExpectedVersions = {
   "expo-file-system": "~15.4.4",
   "expo-font": "~11.4.0",
   "expo-localization": "~14.3.0",
+  "@shopify/flash-list": "1.4.3",
```
frankcalise, line 12: 🔥

frankcalise (approved):
Looks great!

Thanks for the thorough new arch checks too

src/commands/new.ts
markrickert: please don't murder me for this array accessor 🫣

frankcalise: Could we access it via a find for expo-build-properties instead?

frankcalise: @markrickert do you think this helps future proof the array accessor a bit? it's not much cleaner but maybe a little more resilient

```
const expoBuildProperties = appJson.expo.plugins.find(plugin => typeof plugin === 'object' && plugin[0] === 'expo-build-properties')

// access it via expoBuildProperties.ios......
```

markrickert: Well, since we control the generation of this file, I think it's OK to leave as is. I'm actually going to break this out into it's own pull request since it doesn't have anything to do with FlashList, just the new architecture.

mazenchami (approved):
This is a great workaround for the current issue, great work @markrickert!

@ahmedrowaihi for now let's go with this solution until `FlashList` is updated

markrickert:
I'm going to merge this into `@next`. We really like FlashList for use in client projects but there are a few issues that we've tried to mitigate here:

1. `FlashList` doesn't support the new architecture (fabric) yet. [I have an open PR](https://github.com/Shopify/flash-list/pull/931) that rewrites `CellContainer` and `AutoLayoutView` to be compatible with Fabric. For now, if you choose to enable the new architecture and use `FlashList`, things may be wonky (especially on iOS. Android didn't have as many problems in my testing)
2. There are issues with RTL language support in `FlashList`. To mitigate this, we've created a higher order component in ignite called `ListView` that will detect the user's preferred language and use `FlatList` if the language is RTL, otherwise it uses `FlashList`

PR #2517, opened 2023-10-10
fix: ios compiling issue when the user specifies the new architecture

markrickert:

## Describe your PR

This is pulling out a commit from #2513 so that we can get this fix in for the new architecture with that PR stalled for now.

`@react-native-async-storage/async-storage` requires ios 13.4 as a minimum deployment target when using the new architecture.


frankcalise (approved):
Thank you sir

PR #2519, opened 2023-10-10
fix: `Localization.isRTL` is deprecated.

markrickert:

## Describe your PR

`Localization.isRTL` is deprecated.

Screenshot 2023-10-10 at 3 26 04 PM

The deprecation notice for `expo-localization`'s `Localization.isRTL` is the synchronous function call `Localization.getLocales()`. This gets an array of the user’s preferred languages with the first element in the array being their preferred language. This array object contains everything we need to determine their `i18n.locale` and if the language is `rtl` or `ltr`.

As you can see in this video, I'm switching back and forth between English and Arabic (with a little debug notice alert that is not included in this PR, just so we can see what's going on) and the `isRTL` functionality remains the same.

Debug code for my little alert:

```tsx
Alert.alert(
  "Preferred Languages:",
  locales.map((l) => `${l.languageTag} (${l.textDirection})`).join(", "),
)
```
9F29BB84-DC31-499A-9257-86787891FE90-54927-0002E20E3E751C3F

### DIFF
--- a/boilerplate/app/i18n/i18n.ts
+++ b/boilerplate/app/i18n/i18n.ts
@@ -17,7 +17,18 @@ i18n.fallbacks = true
 i18n.translations = { ar, en, "en-US": en, ko, fr }
 
 const locales = Localization.getLocales() // This method is guaranteed to return at least one array item.
-const preferredLanguage = locales[0] // The preferred language is the first element in the array.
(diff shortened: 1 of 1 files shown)


mazenchami (approved):
great fix on this @markrickert

PR #2534, opened 2023-10-22
fix(cli): support for kebab-case names

### LINKED ISSUE #1391: Feature Request: Allow generating kebab-case projects
---
What's going on?
This is more of a feature request than an issue report. `Ignite new` requires an app name to be provided in camel case. For example: `Please use camel case for your project name. Ex: FooBar`. This generates a containing directory using the same project name, such as `FooBar`

For some users, it would be preferable to generate the containing directory in kebab case, such as 'foo-bar'. Furthermore, when generating a new app, the name of the package (i.e. the `name prop in package.json`) is generated in camel case, and Prettier gripes about the format: `String does not match the pattern of "^(?:@[a-z0-9-~][a-z0-9-._~]*/)?[a-z0-9-~][a-z0-9-._~]*$".`

To summarize, this is a request for an option for the user to have the Ignited app directory name in kebab case and to have the package name prop is kebab case as well.

---
Steps to reproduce

* Ignite a new project: `ignite new FooBar`
* Observe that the project directory name is `FooBar`. That's expected, but it would be great if the user could provide an option so the project directory name was in kebab case: `foo-bar`
* Observe the "name" prop in package.json. It is `FooBar`. That's expected, but it would be nice if the package name was in kebab case. At the very least, how about if the user could provide a kebab case option to Ignite so the package name was generated in kebab case?

---
`ignite doctor` results:

```
System
  platform           darwin
  arch               x64
  cpu                8 cores                                              Intel(R) Core(TM) i7-4980HQ CPU @ 2.80GHz
  directory          /Users/derekgreenberg/Documents/ignite-experiments

JavaScript
  node               8.10.0       /usr/local/bin/node
  npm                5.6.0        /usr/local/bin/npm
  yarn               1.10.1       /usr/local/bin/yarn

React Native
  react-native-cli   2.0.1

Ignite
  ignite             2.2.3        /usr/local/bin/ignite

Android
  java               1.8.0_73     /Library/Java/JavaVirtualMachines/jdk1.8.0_73.jdk/Contents/Home/bin/java
  android home       -            /usr/local/opt/android-sdk

iOS
  xcode              10.1
```

frankcalise:

## Describe your PR

- Closes #1391
- Fixes `bun` install (needs `bunfig.toml` to not install peer deps in order for `yarn ios` to work after a fresh install)
- Remove `bunfig.toml` when not using `bun` as package manager
- Remove validation check for `kebab-case` since Expo prebuild will handle the Xcode project generation and strip dashes
- Formats bundle identifier default to strip dashes from project name if passed in as `kebab-case`
- Add dash to the list of allowed `isNotAlphaNumeric` validation

## Screenshots

image


markrickert (approved):
i-tested-these changes-with kebab-case and-found-no-issues!

PR #2536, opened 2023-10-23
demo screen fix drawer opened by default

### LINKED ISSUE #2535: Demo screen : drawer opens unexpectedly on mount due to route.params check always truthy inside useEffect
### Describe the bug

When the screen demo loads, the drawer opens by default. I've traced this behaviour to the handleScroll function being triggered by a React.useEffect block. Within this block, toggleDrawer gets called, changing the drawer's open/close state. The root cause seems to be the condition if (route.params) in the useEffect, which evaluates to true even when route.params is an empty object.

### Ignite version

8.9.3

### Additional info

Inside the useEffect  changing the condition to be `Object.keys(route.params).length > 0` fix the problem

ludocourbin:

## Describe your PR

Fix demo screen : drawer opens unexpectedly on mount due to route.params check always truthy inside useEffect

close #2535

### DIFF
--- a/boilerplate/app/screens/DemoShowroomScreen/DemoShowroomScreen.tsx
+++ b/boilerplate/app/screens/DemoShowroomScreen/DemoShowroomScreen.tsx
@@ -97,7 +97,7 @@ export const DemoShowroomScreen: FC<DemoTabScreenProps<"DemoShowroom">> =
 
     // handle Web links
     React.useEffect(() => {
-      if (route.params) {
+      if (Object.keys(route.params).length > 0) {
         const demoValues = Object.values(Demos)
         const findSectionIndex = demoValues.findIndex(
           (x) => x.name.toLowerCase() === params.queryIndex,
--- a/boilerplate/app/screens/DemoShowroomScreen/DemoShowroomScreen.tsx
+++ b/boilerplate/app/screens/DemoShowroomScreen/DemoShowroomScreen.tsx
@@ -97,7 +97,7 @@ export const DemoShowroomScreen: FC<DemoTabScreenProps<"DemoShowroom">> =
 
     // handle Web links
     React.useEffect(() => {
-      if (Object.keys(route.params).length > 0) {
+      if (Object.keys(params).length > 0) {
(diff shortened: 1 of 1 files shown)


boilerplate/app/screens/DemoShowroomScreen/DemoShowroomScreen.tsx
```diff
@@ -97,7 +97,7 @@ export const DemoShowroomScreen: FC<DemoTabScreenProps<"DemoShowroom">> =
 
     // handle Web links
     React.useEffect(() => {
-      if (route.params) {
+      if (Object.keys(route.params).length > 0) {
```
frankcalise, line 100: let's add a `Platform.OS` check for web here because it still remains open on mobile from what I tested.

frankcalise (changes requested):
Hey thanks for the contribution! Left one comment after some testing.

Also if you wouldn't mind updating line 100 to params and switching that in the dependency array that would be helpful I think - or change everything to route.params and wipe out 96 for consistency.

boilerplate/app/screens/DemoShowroomScreen/DemoShowroomScreen.tsx
ludocourbin: It appears to work as intended in mobile for me, can you guide me to how you tested ? i'm relatively new to using expo and react-native

frankcalise: @ludocourbin I was testing with `npx ignite-cli@next`, let me try it again with the master version and if that works I can make the necessary updates for next as a follow-up

ludocourbin: Perfect, thank you

frankcalise (changes requested):
We probably need to update the dependency array here now to match `params`

frankcalise (approved):
Thank you for your contribution!

ludocourbin:
> Thank you for your contribution!

Thank you for your help

Issue #2137, opened 2022-09-09
Add fatal error handling to `ignite-cli new`


joshuayoes:
Currently, when an error gets thrown on `npx ignite-cli new`, it will throw the fatal error.

It would be nice to display the error to the user, along with the following information
- A CLI command with the options they used
- A link to the Ignite github issues template
- A suggestion to run `npx ignite-cli doctor`

jamonholmgren:
🤔 I wonder if there's a way to detect similar issues based on the errors generated and suggest issues to look at. (Just TOL, for future)

frankcalise:
This issue is closely related to https://github.com/infinitered/ignite/issues/2509

frankcalise:
@jvm-cognition With your other fixes to #2539, would you mind adding a link to the console output to the bug template? I think that will cover this issue as well. We can output a message along the lines of:

```
Consider opening an issue with the following information at:
https://github.com/infinitered/ignite/issues/new?template=bug_report.yml&labels=bug
```

PR #2539, opened 2023-10-25
feat/add-doctor-check

jvm-cognition:

## Describe your PR
feat: add error handling and doctor checks to failed task when running new command.
feat: add new function to pretty tool for stopping last spinner.

Extra: Added some indentation space to doctor command to align with spinners in terminal(reverted this).

### DIFF
--- a/src/commands/new.ts
+++ b/src/commands/new.ts
@@ -133,7 +133,7 @@ export default {
     const { kebabCase } = strings
     const { exists, path, removeAsync, copy, read, write, homedir } = filesystem
     const { info, colors, warning } = print
-    const { gray, cyan, yellow, underline, white } = colors
+    const { gray, cyan, yellow, white, underline, bold } = colors
     const options: Options = parameters.options
 
     const yname = boolFlag(options.y) || boolFlag(options.yes)
@@ -385,6 +385,25 @@ export default {
     p()
     // #endregion
 
+    // #region run doctor
+    let doctorHasRun = false
+    const runDoctor = async () => {
+      startSpinner(" Gathering system and project details")
+      try {
+        const IGNITE = "node " + filesystem.path(__dirname, "..", "..", "bin", "ignite")
+        const doctorResults = await system.run(`${IGNITE} doctor`)
+        p(`\n\n${doctorResults}`)
+        doctorHasRun = true
+      } catch (e) {
+        p(yellow("Unable to gather system and project details."))
+      }
+      stopSpinner(" Gathering system and project details", "🛠️")
+      p()
+      p(white(` Note: For additional information try re-running the command with the ${bold("`--debug`")} flag.`))
+      p()
+    }
+    // #endregion
+
     // #region Overwrite
     if (exists(targetPath) === "dir" && overwrite === true) {
       const msg = ` Tossing that old app like it's hot`
@@ -473,11 +492,16 @@ export default {
     if (shouldUseCache) {
       const msg = `Grabbing those ${packagerName} dependencies from the back`
       startSpinner(msg)
-      await cache.copy({
-        fromRootDir: cachePath,
-        toRootDir: targetPath,
-        packagerName,
-      })
+      try {
+        await cache.copy({
+          fromRootDir: cachePath,
+          toRootDir: targetPath,
+          packagerName,
+        })
+      } catch (e) {
+        log(e)
+        p(yellow("Unable to copy dependecies from cache."))
+      }
       stopSpinner(msg, "📦")
     }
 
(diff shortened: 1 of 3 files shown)


jvm-cognition:
@frankcalise sorry this took so long to release, I kind bamboozled myself because I pulled your `feat/bun-doctor` branch and used it to reproduce the bun bug on `new` command, I did all the work there only to realise I can't merge those changes into master as some things like new arc doesn't exist on it yet  so if those changes ever get merged into the master branch, I could apply those changes as well. That said not sure if this is wat you imagined the solution to be so let me know if I missed the mark.

jvm-cognition:
@frankcalise last thing should I have bumped the version on this commit or when do you guys exactly bump a version on any updates.

frankcalise (changes requested):
So I think what we're looking for here is more if the process ends in a failure, then we run the doctor command following that failure.

This would help because when a user goes to post an issue report, they'd likely copy the error and the doctor ouput, which woudl help us diagnose with the versions of everything installed, etc.

And instead of doing it each catch, we can likely just flag that we need to run doctor at the end, one time.

I'd still look to flag it during dependency install because there have been quite a few issues in the past where that step fails for people.

jvm-cognition:
@frankcalise So what I've done is wrap the whole cli executing process (excluding the prompts) in a try catch this allows for us to catch:

> If the process ends in a failure.

This also fulfills the fact that:

> We need to run doctor at the end, one time.

And with this solution it also catches the pesky dependency install failures

src/commands/new.ts
```diff
@@ -364,319 +367,335 @@ export default {
...
+      startSpinner(" Gathering system and project details")
+      try {
+        const IGNITE = "node " + filesystem.path(__dirname, "..", "..", "bin", "ignite")
+        const doctorResults = await system.run(`${IGNITE} doctor`)
+        p(`\n\n${doctorResults}`)
+      } catch (e) {
+        p(yellow("Unable to gather system and project details."))
+      }
+      stopSpinner(" Gathering system and project details", "🛠️")
+      process.exit(0)
```
jamonholmgren, line 697: This should be `process.exit(1)`, since it was unsuccessful. (`0` is successful, nonzero is failure)

jamonholmgren (approved):
I think this will be helpful!

src/commands/new.ts
```diff
@@ -364,319 +367,335 @@ export default {
...
+      // #endregion
+
+      // this is a hack to prevent the process from hanging
+      // if there are any tasks left in the event loop
+      // like I/O operations to process.stdout and process.stderr
+      // see https://github.com/infinitered/ignite/issues/2084
+      process.exit(0)
+    } catch (e) {
+      stopLastSpinner("❌")
+      p2(red(`\nThe following error occured:\n${e}`))
```
frankcalise, line 687:
```suggestion
      p2(red(`\nThe following error occurred:\n${e}`))
```

frankcalise: Spelling fix

src/commands/new.ts
```diff
@@ -364,319 +367,335 @@ export default {
...
+      startSpinner(" Gathering system and project details")
+      try {
+        const IGNITE = "node " + filesystem.path(__dirname, "..", "..", "bin", "ignite")
+        const doctorResults = await system.run(`${IGNITE} doctor`)
+        p(`\n\n${doctorResults}`)
+      } catch (e) {
+        p(yellow("Unable to gather system and project details."))
+      }
+      stopSpinner(" Gathering system and project details", "🛠️")
```
frankcalise, line 696: Let's just remove this since the process exits, the spinner stops and this ends up printing after the doctor output. It looks cleaner without it and we don't need people adding it to the error report.

frankcalise (changes requested):
Tried it out, working well! Thanks for your help here.

A spelling fix, line removal and process exit update and should be good to go.

src/commands/new.ts
jvm-cognition: @frankcalise Just for clarity you want me to remove `startSpinner(" Gathering system and project details")` and `stopSpinner(" Gathering system and project details", "🛠️")`.

frankcalise: @jvm-cognition definitely the `stopSpinner` line. I think the start one is ok, on an m1 it's fast and maybe not necessary but other machines it might be good to inform. Your call

joshuayoes (approved):
Nice work here!

jvm-cognition:
@frankcalise I applied the requested info to the console output in `issue: #2137` and then adhered to any other comment above.

frankcalise (approved):
Looks good! Thanks for the contribution.

Will look to merge after I pull it down and test

Issue #2084, opened 2022-08-30
New command hangs after process complete


cdanwards:
### Describe the bug

When spinning up a new app through cli, after all of the processes run the node instance will continue to run, hanging the terminal.

<img width="672" alt="Screen Shot 2022-08-30 at 4 11 35 PM" src="https://user-images.githubusercontent.com/8878532/187534209-6025f46c-1625-41ab-8e63-a2971ee4d5c6.png">

### Ignite version

maverick `713919b6`

### Additional info

<img width="677" alt="Screen Shot 2022-08-30 at 4 15 25 PM" src="https://user-images.githubusercontent.com/8878532/187534476-2f435687-4095-420f-bb90-34345c81dd3e.png">

joshuayoes:
It seems related to the usage of `prompt.ask`, if it ever gets called, it sometimes doesn't resolve.

joshuayoes:
This wasn't the case in earlier commits, so I may try to do a git-bisect to see when it go introduced

joshuayoes:
Apparently this has been around since the initial commit in maverick

joshuayoes:
Dan found this using [`npx why-is-node-running`](https://github.com/mafintosh/why-is-node-running#cli)

<img width="1197" alt="Screen Shot 2022-08-31 at 12 07 54 PM" src="https://user-images.githubusercontent.com/37849890/187796330-21a1786c-257c-4296-8dc0-a9f246dfbca2.png">

It seems that either the `prompt` tool or `crossSpawn` tool is never resolving out of the event loop

cdanwards:
I think we can close this issue with #2092 being merged.

Issue #1735, opened 2021-07-14
npx ignite-cli new fooapp hangs on "Cleaning up Expo install"


growlycode:
What's going on?

Attempting to initialize a new expo project, it just hangs on "Cleaning up Expo install". Left it for hours, and it never completed.
When I look at the package.json, none of it is expo-related, it's all react native stuff.

I can can successfully create and run an app using the vanilla expo-cli

```
· · · · · · · · · · · · · · · · · · 🔥 Ignite 🔥 · · · · · · · · · · · · · · · · · ·

    █ Creating fooapp using Ignite 7.0.7
    █ Powered by Infinite Red - https://infinite.red
    █ Using expo-cli
    ────────────────────────────────────────────────

 [object Object]
    - Igniting app npx expo-cli@4 init fooapp --template C:\Users\blahblah\AppData\Local\npm-cache\_npx\e31027f3785124a8\node_modules\ignite-cli\boilerplate --non-interactive        
    🔥 Igniting app
 - Downloading and extracting project files.

 √ Downloaded and extracted project files.

 [10:25:13] 🧶 Using Yarn to install packages. Pass --npm to use npm instead.

    - Summoning Expo CLI - Installing JavaScript dependencies.

    | Summoning Expo CLI > yarn install

    - Summoning Expo CLI yarn install v1.13.0

 info No lockfile found.

 [1/4] Resolving packages...

    - Summoning Expo CLI [2/4] Fetching packages...

    \ Summoning Expo CLI info fsevents@2.3.2: The platform "win32" is incompatible with this module.
info "fsevents@2.3.2" is an optional dependency and failed compatibility check. Excluding it from installation.

 info fsevents@1.2.13: The platform "win32" is incompatible with this module.
info "fsevents@1.2.13" is an optional dependency and failed compatibility check. Excluding it from installation.

 [3/4] Linking dependencies...

    - Summoning Expo CLI [4/4] Building fresh packages...

    - Summoning Expo CLI success Saved lockfile.

 $ node ./bin/postInstall

    / Summoning Expo CLI patch-package 6.2.2

 Applying patches...

    / Summoning Expo CLI 
✔︎ Solidarity checks valid

    / Summoning Expo CLI Jetifier found 1263 file(s) to forward-jetify. Using 12 workers...

    \ Summoning Expo CLI Done in 114.09s.

 √ Installed JavaScript dependencies.

 - Installing CocoaPods...

[...]

Issue #1365, opened 2019-01-07
Error: stderr maxBuffer exceeded


dischord01:
---
What's going on?
Getting an error on fresh install

```
⠴ configuring git an error occured while installing ignite-andross boilerplate.
Error: stderr maxBuffer exceeded
```

---
Steps to reproduce

```
~ » ignite new PizzaApp                                                                                                                                       
-----------------------------------------------
  (                  )   (
  )\ )   (        ( /(   )\ )    *   )
 (()/(   )\ )     )\()) (()/(  ` )  /(   (
  /(_)) (()/(    ((_)\   /(_))  ( )(_))  )\
 (_))    /(_))_   _((_) (_))   (_(_())  ((_)
 |_ _|  (_)) __| | \| | |_ _|  |_   _|  | __|
  | |     | (_ | | .` |  | |     | |    | _|
 |___|     \___| |_|\_| |___|    |_|    |___|
-----------------------------------------------

An unfair headstart for your React Native apps.
https://infinite.red/ignite

-----------------------------------------------

🔥 igniting app PizzaApp
? Which boilerplate would you like to use? Andross (React Navigation, Redux, & Redux Saga)
✔ using the Infinite Red boilerplate v2 (code name 'Andross')
✔ added React Native 0.57.7 in 21.47s
? Would you like Ignite Development Screens? Yes
? What vector icon library will you use? react-native-vector-icons
? What internationalization library will you use? react-native-i18n
? What animation library will you use? react-native-animatable
? Would you like to include redux-persist? Yes
✔ added ignite-andross in 28.42s
✔ added ignite-vector-icons in 17.04s
✔ added ignite-i18n in 17.46s
✔ added ignite-animatable in 14.01s
✔ added ignite-dev-screens in 28.56s
✔ added ignite-redux-persist in 30.99s
✔ added ignite-standard in 36.51s
⠼ configuring gitan error occured while installing ignite-andross boilerplate.
Error: stderr maxBuffer exceeded
⠹ configuring git
```

This last part hangs the terminal.

---
`ignite doctor` results:

```
~ » ignite doctor                                                                                                                                             dischord01@dischord01
System
  platform           darwin
  arch               x64
  cpu                4 cores             Intel(R) Core(TM) i5-6500 CPU @ 3.20GHz
  directory          /Users/dischord01

JavaScript
  node               9.11.1       /Users/dischord01/.nvm/versions/node/v9.11.1/bin/node
[...]

Issue #1287, opened 2018-06-16
App hangs up on splash screen


punktzrzutu:
Hi,
What's going on?
Newly created app hangs up on splash screen, also doesn't react to shake gesture.
It happens both in iOS simulator and Android device.
<img width="395" alt="screen shot 2018-06-16 at 13 54 11" src="https://user-images.githubusercontent.com/6332985/41498394-c6102eba-716c-11e8-9b48-9675cc1edc10.png">

---
Steps to reproduce

$ ignite new PizzaApp
  ( Choose `Bowser (alpha)` when prompted )
$ cd PizzaApp
$ ignite generate screen pizza-location-list
  ( Choose `example` domain when prompted. This determines where your new screen will go in the directory structure. )
$ ignite generate component pizza-location
  ( Choose `example` domain when prompted. This determines where your new screen will go in the directory structure. )
$ react-native run-ios

---
`ignite doctor` results:

```
System
  platform           darwin                                                                                              
  arch               x64                                                                                                 
  cpu                8 cores                                        Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz  
  directory          /Users/bartek/code/reactnative/ignite/Pizza3                                                        

JavaScript
  node               8.11.1       /Users/bartek/.nvm/versions/node/v8.11.1/bin/node  
  npm                6.0.0        /Users/bartek/.nvm/versions/node/v8.11.1/bin/npm   
  yarn               1.7.0        /usr/local/bin/yarn                                

React Native
  react-native-cli   2.0.1       
  app rn version     0.55.4      

Ignite
  ignite             2.1.0        /Users/bartek/.nvm/versions/node/v8.11.1/bin/ignite  

Android
  java               1.8.0_92     /usr/bin/java                      
  android home       -            /Users/bartek/Library/Android/sdk  

iOS
  xcode              9.2         
```

lustigdev:
👍

lustigdev:
@punktzrzutu,

Try using this instead: `ignite new PizzaApp -b ir-boilerplate`

Worked for me. Still familiarizing myself with ignite though, but I don't think this is the bowser boilerplate.

mrgoodrich:
+1

kevinvangelder:
[...]

PR #2546, opened 2023-11-08
Add timeout to avoid long-term hangs

jamonholmgren:
We've had multiple issues over the years with Ignite hanging forever and people (patiently!) waiting for hours for it to finish.

Examples: #1249, #2084, #1735, #1365, #1287, #172, and many more.

This adds a simple timeout that won't allow Ignite to take longer than 10 minutes once we start spinning up a new app.

This might be a bit too aggressive; I don't know what our longest installs take these days, but I just tried one using `npm` as the package manager and choosing the `DIY` option on Ignite v9 and it took roughly 72 seconds, which is 12% of this timeout.

This does NOT need to be merged prior to v9, but can be; it's a relatively small change.


bun.lockb
frankcalise, line 1: Think you can cut this from the PR, maybe belongs in the top level gitignore if we missed that

.circleci/config.yml
```diff
@@ -10,17 +10,12 @@ defaults: &defaults
   working_directory: /mnt/ramdisk/repo
 
 version: 2.1
-orbs:
-  # https://circleci.com/developer/orbs/orb/cmgriffing/bun-orb
-  bun-orb: cmgriffing/bun-orb@0.0.26
```
frankcalise, line 15: Did you mean to remove this orb and bun from CI?

jamonholmgren: Yes but I forgot to commit the fix I did instead. 😅

frankcalise (approved):

frankcalise:
@jamonholmgren one way to avoid the timer being too aggressive, maybe a cli flag to disable it for certain situations? like terrible car/plane wifi or something?

jamonholmgren:
Added a `--no-timeout` flag!

jamonholmgren:
@frankcalise I'll let you merge when you're ready, if you're ready!

src/commands/new.ts
```diff
@@ -775,6 +797,7 @@ module.exports = {
         useCache,
         y: yname,
         yes: yname,
+        timeout: !noTimeout,
```
frankcalise, line 800: I think this should be `noTimeout` or `no-timeout: noTimeout`. This builds the CLI command to re-run the input the user just gave, so currently this will output:

image


But the CLI switch isn't `timeout`

frankcalise: Passing `--timeout` won't achieve `noTimeout: true`

frankcalise: Fixed!

PR #2568, opened 2023-11-22
Update Screen.tsx

JenimDacket:
Updated the <KeyboardAvoidingView> component prop "behavior" as it was setting the non-ios value to "undefined" instead of "height". 

This was causing malfunctions on android devices and emulators.

- [ x] `yarn test` **jest** tests pass with new tests, if relevant
- [x ] `README.md` has been updated with your changes, if relevant

## Describe your PR
Updated the <KeyboardAvoidingView> boilerplate component prop "behavior" as it was setting the non-ios value to "undefined" instead of "height". 

This was causing malfunctions on android devices and emulators.

### DIFF
--- a/boilerplate/app/components/Screen.tsx
+++ b/boilerplate/app/components/Screen.tsx
@@ -204,7 +204,7 @@ export function Screen(props: ScreenProps) {
       <StatusBar style={statusBarStyle} {...StatusBarProps} />
 
       <KeyboardAvoidingView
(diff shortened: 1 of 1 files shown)


frankcalise:
@JenimDacket thanks for the contribution! Can you explain some test steps to me so I can verify it is working versus the old behavior for testing purposes?

JenimDacket:
That's correct! Sorry for the late response!

PR #2572, opened 2023-11-29
feat(cli): remove unused deps upon demo removal

### LINKED ISSUE #2544: [cli] --remove-demo should clean up unused deps
## Problem

The `--remove-demo` cli switch successfully removes demo mark up, files and assets (in v9). It would be nice if it also removed dependencies that are no longer used so the dev doesn't have to figure this out.

## Packages to Remove

- `react-native-drawer-layout`
- `expo-application`
- `@react-navigation/bottom-tabs`

frankcalise:
## Describe your PR
- Closes #2544
- If the user decides to remove the demo application, there are a handful of dependencies that we can clean up out of the way that go unused with just the basic Welcome screen

### DIFF
--- a/src/commands/new.ts
+++ b/src/commands/new.ts
@@ -30,6 +30,7 @@ import {
   expoGoCompatExpectedVersions,
   findAndUpdateDependencyVersions,
 } from "../tools/expoGoCompatibility"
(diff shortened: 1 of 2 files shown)


joshuayoes (approved):
Haven't run these changes locally yet but this looks like an elegant solution!

PR #2593, opened 2023-12-27
feat: add props for disabled styles in Button component

coolsoftwaretyler:

- [x] `yarn test` **jest** tests pass with new tests, if relevant (one failure, I'll post details in the PR description)
- [x] `README.md` has been updated with your changes, if relevant (I don't think it's relevant for this one)

## Describe your PR

I was copying Ignite's `Button` component into my own codebase and wanted to have a disabled `Button` visually indicated with some custom styles. This change was really helpful for me, so I wanted to propose it to y'all while I wait for my CI to pass haha.

No worries if you're not interested, I imagine folks could style this from outside by syncing up some conditional styles with any conditions they use to set the `disabled` prop, but I kind of like that this reads the `disabled` prop directly at author-time and behaves as expected.

Let me know what ya think. Happy to rework anything you'd like. Thanks one way or another!

## Test failures

I had one set of test failures, but I don't think they're related to this change. I get the same results on the `master` branch in my fork.

<details>
Output: 

```sh
in ~/forked-ignite on feat/add-disabled-styles-to-ignite-button λ yarn test
yarn run v1.22.11
$ TS_JEST_DISABLE_VER_CHECKER=true jest
 PASS  src/tools/demo.test.ts
 PASS  src/tools/packager.test.ts
 PASS  test/vanilla/ignite-help.test.ts (9.151 s)
 PASS  test/vanilla/ignite-remove-demo.test.ts (9.492 s)
 PASS  test/vanilla/ignite-generate.test.ts (31.918 s)
 FAIL  test/vanilla/ignite-new.test.ts (277.184 s)
  ● ignite new › ignite new Foo --debug --packager=bun --yes › should be able to use `generate` command and have pass output pass bun run test, bun run lint, and bun run compile scripts

    Command failed: cd /private/var/folders/4y/6gtftfw936v8f4p7tzjdxl900000gn/T/ignite-e58e9255968265d5869ac97b0bed02d4/Foo && bun run test && cd /Users/tylerwilliams/forked-ignite
    $ jest
     FAIL  test/i18n.test.ts
      ● Test suite failed to run

        TypeError: Cannot redefine property: performance

          at Object.<anonymous> (node_modules/react-native/jest/setup.js:410:20)

     FAIL  app/models/Episode.test.ts
      ● Test suite failed to run

        TypeError: Cannot redefine property: performance

          at Object.<anonymous> (node_modules/react-native/jest/setup.js:410:20)

     FAIL  app/utils/storage/storage.test.ts
      ● Test suite failed to run

        TypeError: Cannot redefine property: performance

          at Object.<anonymous> (node_modules/react-native/jest/setup.js:410:20)

     FAIL  app/models/ModTest.test.ts
      ● Test suite failed to run

        TypeError: Cannot redefine property: performance

          at Object.<anonymous> (node_modules/react-native/jest/setup.js:410:20)

     FAIL  app/services/api/apiProblem.test.ts
      ● Test suite failed to run

        TypeError: Cannot redefine property: performance

          at Object.<anonymous> (node_modules/react-native/jest/setup.js:410:20)

    Test Suites: 5 failed, 5 total
    Tests:       0 total
    Snapshots:   0 total
    Time:        5.374 s
    Ran all test suites.
    error: script "test" exited with code 1 (SIGHUP)



Test Suites: 1 failed, 5 passed, 6 total
Tests:       1 failed, 33 passed, 34 total
Snapshots:   31 passed, 31 total
Time:        278.093 s
Ran all test suites.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
tylerwilliams in ~/forked-ignite on feat/add-disabled-styles-to-ignite-button λ 
```
</details>

### DIFF
--- a/boilerplate/app/components/Button.tsx
+++ b/boilerplate/app/components/Button.tsx
@@ -70,6 +70,10 @@ export interface ButtonProps extends PressableProps {
    * https://reactnative.dev/docs/pressable#disabled
    */
   disabled?: boolean
+  /**
+   * An optional style override for the disabled state
+   */
+  disabledStyle?: StyleProp<ViewStyle>
 }
 
 /**
@@ -91,6 +95,7 @@ export function Button(props: ButtonProps) {
     RightAccessory,
     LeftAccessory,
     disabled,
+    disabledStyle,
     ...rest
   } = props
 
@@ -100,7 +105,7 @@ export function Button(props: ButtonProps) {
       $viewPresets[preset],
       $viewStyleOverride,
       !!pressed && [$pressedViewPresets[preset], $pressedViewStyleOverride],
-      disabled && { opacity: 0.5 },
+      disabled && disabledStyle,
     ]
   }
   function $textStyle({ pressed }: PressableStateCallbackType) {
--- a/boilerplate/app/components/Button.tsx
+++ b/boilerplate/app/components/Button.tsx
@@ -123,7 +123,13 @@ export function Button(props: ButtonProps) {
   }
 
   return (
-    <Pressable style={$viewStyle} accessibilityRole="button" {...rest} disabled={disabled}>
+    <Pressable
+      style={$viewStyle}
+      accessibilityRole="button"
+      accessibilityState={{ disabled: !!disabled }}
+      {...rest}
+      disabled={disabled}
+    >
       {(state) => (
         <>
           {!!LeftAccessory && <LeftAccessory style={$leftAccessoryStyle} pressableState={state} />}
--- a/boilerplate/app/components/Button.tsx
+++ b/boilerplate/app/components/Button.tsx
@@ -65,6 +65,10 @@ export interface ButtonProps extends PressableProps {
    * Children components.
    */
   children?: React.ReactNode
+  /**
+   * disabled prop, accessed directly for styling reasons.
+   */
+  disabled?: boolean
 }
 
 /**
@@ -85,6 +89,7 @@ export function Button(props: ButtonProps) {
     children,
     RightAccessory,
     LeftAccessory,
+    disabled,
     ...rest
   } = props
 
@@ -94,6 +99,7 @@ export function Button(props: ButtonProps) {
       $viewPresets[preset],
       $viewStyleOverride,
       !!pressed && [$pressedViewPresets[preset], $pressedViewStyleOverride],
+      disabled && { opacity: 0.5 },
     ]
   }
   function $textStyle({ pressed }: PressableStateCallbackType) {
@@ -105,7 +111,7 @@ export function Button(props: ButtonProps) {
   }
 
   return (
-    <Pressable style={$viewStyle} accessibilityRole="button" {...rest}>
+    <Pressable style={$viewStyle} accessibilityRole="button" {...rest} disabled={disabled}>
       {(state) => (
         <>
           {!!LeftAccessory && <LeftAccessory style={$leftAccessoryStyle} pressableState={state} />}
--- a/boilerplate/app/components/Button.tsx
+++ b/boilerplate/app/components/Button.tsx
@@ -15,6 +15,7 @@ type Presets = keyof typeof $viewPresets
 export interface ButtonAccessoryProps {
   style: StyleProp<any>
   pressableState: PressableStateCallbackType
+  disabled?: boolean
 }
 
 export interface ButtonProps extends PressableProps {
--- a/boilerplate/app/components/Button.tsx
+++ b/boilerplate/app/components/Button.tsx
@@ -48,6 +48,10 @@ export interface ButtonProps extends PressableProps {
    */
   pressedTextStyle?: StyleProp<TextStyle>
   /**
+   * An optional style override for the button text when in the "disabled" state.
+   */
+  disabledTextStyle?: StyleProp<TextStyle>
+  /**
    * One of the different types of button presets.
    */
   preset?: Presets
@@ -91,6 +95,7 @@ export function Button(props: ButtonProps) {
     pressedStyle: $pressedViewStyleOverride,
     textStyle: $textStyleOverride,
     pressedTextStyle: $pressedTextStyleOverride,
+    disabledTextStyle: $disabledTextStyleOverride,
     children,
     RightAccessory,
     LeftAccessory,
@@ -113,6 +118,7 @@ export function Button(props: ButtonProps) {
       $textPresets[preset],
       $textStyleOverride,
       !!pressed && [$pressedTextPresets[preset], $pressedTextStyleOverride],
+      disabled && $disabledTextStyleOverride,
     ]
   }
 
--- a/boilerplate/app/components/Button.tsx
+++ b/boilerplate/app/components/Button.tsx
@@ -100,7 +100,7 @@ export function Button(props: ButtonProps) {
     RightAccessory,
     LeftAccessory,
     disabled,
-    disabledStyle,
+    disabledStyle: $disabledViewStyleOverride,
     ...rest
   } = props
 
@@ -110,7 +110,7 @@ export function Button(props: ButtonProps) {
       $viewPresets[preset],
       $viewStyleOverride,
       !!pressed && [$pressedViewPresets[preset], $pressedViewStyleOverride],
-      disabled && disabledStyle,
+      !!disabled && $disabledViewStyleOverride,
     ]
   }
   function $textStyle({ pressed }: PressableStateCallbackType) {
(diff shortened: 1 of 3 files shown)


abtonc:
Good call! Instead of just giving option to change the opacity, button component could provide `disabled` bool and `disabledStyle` props(with disabled preset's). I think this is needed and I am sure that Infinitered team also uses a similar approach to handle disabled button styles for their clients.

coolsoftwaretyler:
I really like the idea of using the style prop. I'll update this PR tomorrow to that effect and see how it goes.

coolsoftwaretyler:
Hey folks, I took @abtonc's idea and this PR now uses a `disabledStyle` prop if `disabled` is truthy. Thanks for your consideration!

frankcalise (commented):
Looks good, thanks for the contribution!

Is there a typical default disabled style we could add here that would work with the predefined styles? And then we could override it with the passed in `disabledStyle` prop?

coolsoftwaretyler:
Doesn't look like React Native does anything by default on either `Pressable` components, or `TouchableOpacity` components.

Disabled `TouchableOpacity` components just don't change opacity, but there's no visible difference on `Pressables`.

Here's a Snack: https://snack.expo.dev/@coolsoftwaretyler/pressable-disabled-styles-demo.

I think a default option is nice, but it might actually be closer to the React Native implementation to leave it alone.

boilerplate/app/components/Button.tsx
```diff
@@ -65,6 +65,15 @@ export interface ButtonProps extends PressableProps {
...
   children?: React.ReactNode
+  /**
+   * disabled prop, accessed directly for declarative styling reasons.
+   * https://reactnative.dev/docs/pressable#disabled
+   */
+  disabled?: boolean
+  /**
+   * An optional style override for the disabled state
+   */
+  disabledStyle?: StyleProp<ViewStyle>
```
yulolimum, line 76: If we're visually affecting one part of the component for an interaction, we should probably provide the user with the option to configure all part. Let's add a `disabledTextStyle`. To the props.

boilerplate/app/components/Button.tsx
```diff
@@ -105,7 +117,7 @@ export function Button(props: ButtonProps) {
   }
 
   return (
-    <Pressable style={$viewStyle} accessibilityRole="button" {...rest}>
+    <Pressable style={$viewStyle} accessibilityRole="button" {...rest} disabled={disabled}>
```
yulolimum, line 120: I'm not an accessibility expert, but do we need a `accessibilityState={{disabled}}` here (similar to the `Toggle` component)?

boilerplate/app/components/Button.tsx
```diff
@@ -105,7 +117,7 @@ export function Button(props: ButtonProps) {
...
       {(state) => (
         <>
           {!!LeftAccessory && <LeftAccessory style={$leftAccessoryStyle} pressableState={state} />}
```
yulolimum, line 123: Similar to the `disabledTextStyle` comment, for consistency, we should pass the `disabled` state into these accessories so the user has a choice to style them via these props rather than higher ordered state. 

`{!!LeftAccessory && <LeftAccessory style={$leftAccessoryStyle} pressableState={state} disabled={disabled} />}`

Example:
```tsx
<Button
  disabled={someComplex.any(logic => !!logic)}
  LeftAccessory={(props) => <Icon style={[props.disabled && $disabledIconStyle]} />}
```

yulolimum (commented):
@coolsoftwaretyler 

Impressive work! This is a great change!

I added a couple comments regarding supporting customizability of all Button parts. Additionally, if you could drop a couple notes about this change in the [Documentation](https://github.com/infinitered/ignite/blob/master/docs/boilerplate/components/Button.md), as well as add usage example to the [Showroom under "Styling"](https://github.com/infinitered/ignite/blob/master/boilerplate/app/screens/DemoShowroomScreen/demos/DemoButton.tsx), that would be a bonus!

coolsoftwaretyler:
Thanks for the review @yulolimum! I appreciate your time. I will address your feedback in the next few days.

boilerplate/app/components/Button.tsx
coolsoftwaretyler: Ah yes, makes sense!

Should be good in https://github.com/infinitered/ignite/pull/2593/commits/a880064758b7d01b47848b12a160b2bfe5e2b231

I also noticed my disabled style diverged from the `$[something]StyleOverride` pattern, so I refactored to match y'all a little more closely in https://github.com/infinitered/ignite/pull/2593/commits/3b99aca24920c0800f19ac37e9cd2bb29f4dc2f3

boilerplate/app/components/Button.tsx
coolsoftwaretyler: Great catch! Fixed in https://github.com/infinitered/ignite/pull/2593/commits/e6b72d46c1985c86ed4831e8251afe799c247549

boilerplate/app/components/Button.tsx
coolsoftwaretyler: Ah yes, love it. Updated in https://github.com/infinitered/ignite/pull/2593/commits/d329a0cf8928498278e4de3b2703048623ce8090

coolsoftwaretyler: Whoops, forgot to update the interface. I'll do that along with the docs/examples updates in a push later on.

coolsoftwaretyler:
Hey @yulolimum - I have incorporated your feedback, and [added some docs](https://github.com/infinitered/ignite/pull/2593/commits/e280c516a094540b3cda3cf3643a3c7c744adfff), and [updated the showroom](https://github.com/infinitered/ignite/pull/2593/commits/b61e1916c1c99fd9dc095de751fea81a6c37c4b8).

I'm not exactly sure what y'all would want in the Showroom, so I'm happy to take additional feedback. Here's a screenshot from my simulator for quick reference:



Thanks again for all your time!

abtonc:
I think supporting presets for disabled state would be a helpful feature to maintain a design system out-of-the-box.

coolsoftwaretyler:
> Hey @yulolimum - I have incorporated your feedback, and [added some docs](https://github.com/infinitered/ignite/pull/2593/commits/e280c516a094540b3cda3cf3643a3c7c744adfff), and [updated the showroom](https://github.com/infinitered/ignite/pull/2593/commits/b61e1916c1c99fd9dc095de751fea81a6c37c4b8).
> 
> I'm not exactly sure what y'all would want in the Showroom, so I'm happy to take additional feedback. Here's a screenshot from my simulator for quick reference:
> 
> 
> 
> Thanks again for all your time!

Realizing now that you asked for this in a specific place. I will adjust that tomorrow. Sorry about that!

yulolimum:
> I think supporting presets for disabled state would be a helpful feature to maintain a design system out-of-the-box.

@abtonc We have a default disabled style in the works from the UX team. I'd like to get this one merged meanwhile.

> Realizing now that you asked for this in a specific place. I will adjust that tomorrow. Sorry about that!

@coolsoftwaretyler I think what you have done is probably better since there are quite a few things to show for the disabled state. Great work!

yulolimum (approved):
@coolsoftwaretyler great work man and thanks for addressing everything so quickly!

cc @frankcalise

coolsoftwaretyler:
Happy to contribute! Thanks for the review.

docs/boilerplate/components/Button.md
```diff
@@ -116,6 +124,14 @@ The `pressedStyle` prop is optional. This can be used to style the `Pressable` c
 <Button pressedStyle={{ backgroundColor: "red" }} />
 ```
 
+### `disabledStyle`
+
+The `disabledStyle` prop is optional. This can be used to style the `Pressable` component of the `Button` when the `disabled` prop is truthy. Values passed here will voerride anything set in the preset.
```
frankcalise, line 129: Small typo here, voerride -> override

frankcalise (approved):
Looks good to me! I'll merge it after the doc typo fix if you don't mind.

Thanks for the very complete PR and @yulolimum for the review!

docs/boilerplate/components/Button.md
coolsoftwaretyler: Whoops! Thank you. Fixed in https://github.com/infinitered/ignite/pull/2593/commits/2336fef1fdfb55c3848ed18349d5b9d94507b9c9

coolsoftwaretyler:
@frankcalise - typo fixed. Thanks again for everyone's time!

PR #2602, opened 2024-01-05
feat(only-use-internal-diff)

morganick:

- [ ] ~`README.md` has been updated with your changes, if relevant~

## Describe your PR
If you have your `git diff` configured with an external diff tool, the tests will not pass because it relies on the output of the internal `git diff` command. This updates  the command lines arguments for consistency to force `git` to use the internal diff tool always for the spec. (Docs: https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---no-ext-diff)


frankcalise (approved):
Thanks for looking into this!

PR #2622, opened 2024-02-01
fix(localization): fix default locale issue for non en-US 

bi3ri:
Context: https://infiniteredcommunity.slack.com/archives/C41PK1LAY/p1706234516339489
https://infiniteredcommunity.slack.com/archives/C41PK1LAY/p1706940969197049

Resolves #2613

PR #2631, opened 2024-02-09
fix (localization): Guard against i18n errors for older Android platforms

douglowder:

## Describe your PR

On certain older platforms, the `systemLocale` returned by `expo-localization` might be invalid, leading to a crash when `systemLocale.languageTag` or `systemLocale.textDirection` are referenced. This PR just adds optional chaining for these references and defaults to `en-US` if there is a problem.

Also changed the default locale from `en` to `en-US` to fix an Apple TV issue, but this may have already been dealt with in #2622 .

Reviewers: if desired, we could add `console.warn()` calls if this condition is encountered, but `expo-localization` is already writing Android logcats if there are errors.


mazenchami (approved):
Thanks for putting this together! Much appreciated 🚀 🎉

PR #2636, opened 2024-02-13
feat(Theming): Adds theming support with documentation.

markrickert:

## Describe your PR

This PR adds theming support to the ignite boilerplate as well as relevant documentation on how to use the theming system.

Note that this PR is onto the `v10` branch and not `main`.


jamonholmgren:
Before we merge this, I want to make sure we get some real-world experience. @markrickert you're starting a new project soon, so this will be a good opportunity for that.

docs/boilerplate/app/theme/Theming.md
```diff
@@ -31,3 +31,86 @@ Timings are defined in `app/theme/timing.ts`. They can be used for consistent an
 Spacing is a first class citizen in Ignite. We use a spacing scale to define the spacing between elements in the app. This allows us to have a consistent spacing scale across the app, and also allows us to change the spacing easily. It is recommended to use the spacing scale for all spacing in the app if possible.
 
 [Spacing](./spacing.ts.md)
+
+# Taking Advantage of Multiple Themes (a.k.a "Dark Mode")
+
+The ignite boilerplate ships with color pallette definitions and support for multiple themes! By default we define two themes, but you can easily add more using our generic theming system.
```
jamonholmgren, line 37:
```suggestion
The Ignite boilerplate ships with color palette definitions and support for multiple themes! By default we define two themes, but you can easily add more using our generic theming system.
```

docs/boilerplate/app/theme/Theming.md
```diff
@@ -31,3 +31,86 @@ Timings are defined in `app/theme/timing.ts`. They can be used for consistent an
 Spacing is a first class citizen in Ignite. We use a spacing scale to define the spacing between elements in the app. This allows us to have a consistent spacing scale across the app, and also allows us to change the spacing easily. It is recommended to use the spacing scale for all spacing in the app if possible.
 
 [Spacing](./spacing.ts.md)
+
+# Taking Advantage of Multiple Themes (a.k.a "Dark Mode")
```
jamonholmgren, line 35:
```suggestion
# Multiple Themes (a.k.a "Dark Mode")
```

docs/boilerplate/app/theme/Theming.md
```diff
@@ -31,3 +31,86 @@ Timings are defined in `app/theme/timing.ts`. They can be used for consistent an
...
+    </View>
+  )
+}
+```
+
+For more information on the `useAppTheme()`, [check out its documentation](../utils/useAppTheme.tsx.md).
+
+## Switching Between Themes
+
+Theme support would be useless if there wasn't a built-in way to switch the theme at will. Ignite's theming system will automatically pick the theme that matches the user's system configuration but you can override this using the `
```
jamonholmgren, line 81:
```suggestion
Theme support would be useless if there wasn't a built-in way to switch the theme at will. Ignite's theming system will automatically pick the theme that matches the user's system configuration (light mode or dark mode) but you can override this using the `
```

jamonholmgren (commented):
This is really looking good. Just want to wait for real-world experience.

markrickert:
@jamonholmgren Absolutely. I'll keep this branch up to date with `main` and use this as the starting point for my next project.

boilerplate/app/components/Button.tsx
```diff
@@ -181,53 +184,54 @@ const $baseViewStyle: ViewStyle = {
...
 }
 
-const $textPresets: Record<Presets, StyleProp<TextStyle>> = {
-  default: $baseTextStyle,
-  filled: $baseTextStyle,
-  reversed: [$baseTextStyle, { color: colors.palette.neutral100 }],
+const $textPresets: Record<Presets, ThemedStyleArray<TextStyle>> = {
+  default: [$baseTextStyle],
+  filled: [$baseTextStyle],
+  reversed: [$baseViewStyle, ({ colors }) => ({ color: colors.palette.neutral100 })],
```
Jpoliachik, line 224: found a bug in our sample app - this should be changed: 
```suggestion
  reversed: [$baseTextStyle, ({ colors }) => ({ color: colors.palette.neutral100 })],
```

markrickert: Fixed in 9fe7419

boilerplate/app.json
```diff
@@ -7,6 +7,7 @@
     "scheme": "helloworld",
     "version": "1.0.0",
     "orientation": "portrait",
+    "userInterfaceStyle": "automatic",
```
frankcalise, line 10: I think we need `expo-system-ui` installed to support this on Android

image

markrickert: I haven't actually tested this fix explicitly in android. I'll do that and update this branch as necessary. Thanks for surfacing this!

markrickert: @frankcalise I rebased on the v10 branch and fixed a few things and I'm not seeing this when i run `npx expo prebuild --clean` in the boilerplate directory.

markrickert:
Rebased on the `v10` branch and fixed a few issues caused by the Switch/Toggle rewrite

frankcalise:
TODO: need to check over the showroom issue: https://infiniteredcommunity.slack.com/archives/C41PK1LAY/p1716828905367989?thread_ts=1716662953.807429&cid=C41PK1LAY

boilerplate/app/theme/index.ts
```diff
@@ -1,4 +1,73 @@
...
+ */
+export type ThemedStyle<T> = (theme: Theme) => T
+export type ThemedStyleArray<T> = (
+  | ThemedStyle<T>
+  | StyleProp<T>
+  | (StyleProp<T> | ThemedStyle<T>)[]
+)[]
+
+// Export the two Theme objects:
+export { colorsLight, colorsDark }
```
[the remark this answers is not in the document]
markrickert: I just merged new changes in from v10 and also updated this for the exports to work with the old theming system so people can gradually adopt the new theming system.

boilerplate/app/theme/index.ts
```diff
@@ -1,4 +1,73 @@
...
+ */
+export type ThemedStyle<T> = (theme: Theme) => T
+export type ThemedStyleArray<T> = (
+  | ThemedStyle<T>
+  | StyleProp<T>
+  | (StyleProp<T> | ThemedStyle<T>)[]
+)[]
+
+// Export the theme objects with backwards compatibility for the old theme structure.
+export { colorsLight as colors, colorsDark, spacingLight as spacing, typography }
```
frankcalise, line 71: Should `spacingDark` be exported here as well?

frankcalise: Actually I guess there are no `spacingDark` calls to be back compat with, just would be a consistency thing since we're exporting `colorsDark`

markrickert:
This is ready for another review.

boilerplate/app.json
markrickert: Fixed in https://github.com/infinitered/ignite/pull/2636/commits/2850c86b755a91713b28bf421b57c30a06699bc1

boilerplate/package.json
```diff
@@ -81,6 +81,9 @@
     "eslint": "8.17.0",
     "eslint-config-prettier": "8.5.0",
     "eslint-config-standard": "17.0.0",
+    "eslint-plugin-import": "2.26.0",
```
lindboe, line 85: That's so odd, do we know why we could have been depending on these but they weren't installed by the package.json?

frankcalise: It's possible I removed some too aggressively at some point

lindboe:
Looking into root cause of the screen jumping in this video (running on Android device), as I don't think it's themeing, but it does point out to me that I think we should be setting the actual app background as a dark background color so we don't get any white strips accidentally. Since we have expo-system-ui, I think https://docs.expo.dev/versions/latest/sdk/system-ui/#systemuisetbackgroundcolorasynccolor should do it?

https://github.com/user-attachments/assets/f463b03b-aaa4-421b-9083-bf91d132666c

boilerplate/app/screens/DemoShowroomScreen/DemoShowroomScreen.tsx
```diff
@@ -177,24 +187,28 @@ export const DemoShowroomScreen: FC<DemoTabScreenProps<"DemoShowroom">> =
 
           <SectionList
             ref={listRef}
-            contentContainerStyle={$sectionListContentContainer}
+            contentContainerStyle={themed($sectionListContentContainer)}
             stickySectionHeadersEnabled={false}
-            sections={Object.values(Demos)}
+            keyExtractor={(item, index) => item.name + index}
```
lindboe, line 192: I don't think this is working as intended for me. I'm getting a unique key warning, and when I log the values it's giving undefined for `item.name` and 0 for index. Which is very weird.

boilerplate/app/components/EmptyState.tsx
```diff
@@ -108,22 +110,28 @@ interface EmptyStatePresetItem {
...
 export function EmptyState(props: EmptyStateProps) {
+  const {
+    themeContext,
+    themed,
+    theme: { spacing },
+  } = useAppTheme()
+
+  const EmptyStatePresets = {
+    generic: {
+      imageSource: themeContext === "dark" ? sadFaceDark : sadFace,
```
lindboe, line 128: Good catch! I think we should probably also do this for the face on the Welcome screen.

I wonder if we can add this as a general image pattern where the swapping can be handled automatically? not necessary, just if you want

boilerplate/package.json
markrickert: Yeah, i couldn't run `yarn lint` in an ignited app from my branch so i had to add those back in yesterday when i was working on this

boilerplate/app/components/EmptyState.tsx
markrickert: While fixing the welcome screen, i realized i can just use `tintColor` on these images since there's hardly any color data. In dark mode, just tint the images white! It converts all non transparent pixels to that color. I'll remove some of the image assets i added to this PR!

```jsx
<Image
  style={$welcomeFace}
  source={welcomeFace}
  resizeMode="contain"
  tintColor={themeContext === "dark" ? "#FFFFFF" : undefined}
/>
```

boilerplate/package.json
lindboe: @frankcalise part of what I'm surprised about is that tests didn't catch this. should we run commands like lint in an ignited app on CI?

boilerplate/app/components/EmptyState.tsx
markrickert: Here's what the welcome screen looks like now: 
2EF7AEFF-E099-40C6-ACF3-34BEE7592E17-52253-00023A4B247E5E68

boilerplate/package.json
frankcalise: hmm well it is inside the `ignite-new.test.ts` so maybe that isn't failing like we'd expect?


```
      // #region Assert package.json Scripts Can Be Run
      // run the tests; if they fail, run will raise and this test will fail
      await run(`bun run test`, runOpts)
      await run(`bun run lint`, runOpts)
      await run(`bun run compile`, runOpts)
      expect(await run("git diff HEAD --no-ext-diff", runOpts)).toContain("+  Bowser: undefined")
      // #endregion
```

lindboe: It looks like these are required as part of `eslint-config-standard`, which I guess is what `standard` in the eslintConfig refers to?

so I'm not sure why that wouldn't come up for that test, at a glance.

boilerplate/app/theme/spacingDark.ts
```diff
@@ -0,0 +1,14 @@
+const SPACING_MULTIPLIER = 0.6
+
+// This is an example of how you can have different spacing values for different themes.
+export const spacing = {
```
lindboe, line 4: I think some spacing on light mode has been messed up by these changes, specifically the drawer menu and the visible icon for the password field (this is wrong on dark mode too).

I think the spacing on the demo screen between the header and first section also looks wrong, but I'm less sure it wasn't that way before.

Screenshot 2024-08-05 at 2 23 51 PM


https://github.com/user-attachments/assets/432d9416-1734-465b-84ab-41e33713ad34

Screenshot 2024-08-05 at 2 27 05 PM

boilerplate/app/screens/DemoShowroomScreen/DrawerIconButton.tsx
```diff
@@ -3,7 +3,8 @@ import { Pressable, PressableProps, ViewStyle, Platform } from "react-native"
 import Animated, { interpolate, interpolateColor, useAnimatedStyle } from "react-native-reanimated"
 import { useDrawerProgress } from "react-native-drawer-layout"
 import { isRTL } from "../../i18n"
-import { colors, spacing } from "../../theme"
+import { useAppTheme } from "app/utils/useAppTheme"
+import type { ThemedStyle } from "app/theme"
```
lindboe, line 7: This icon actually isn't consistently updating with the rest of the app. Haven't looked at why yet. note it took a few attempts to reproduce, but eventually the icon fails to turn white when moving into dark mode (that's the only way I've seen it fail so far, I've never seen it fail to turn dark when moving into light mode).

https://github.com/user-attachments/assets/80e8f84d-ab78-49b7-82c4-42c0d4de51d3

docs/Guide.md
```diff
@@ -80,15 +80,28 @@ We instead use a strategy of constants, colocated with our components, camelCase
...
+// This is a themed style that you must wrap with `themed()` to pass the style object.
+const $container: ThemedStyle<ViewStyle> = ({ colors }) => ({
   flex: 1,
-  backgroundColor: palette.bgColor,
+  backgroundColor: colors.palette.bgColor,
+})
+
+// This is a non-themed style
+const $innerView: ViewStyle{
+  backgroundColor: '#fff,
```
lindboe, line 94: seems like a typo here?

docs/boilerplate/app/theme/Theming.md
```diff
@@ -31,3 +31,151 @@ Timings are defined in `app/theme/timing.ts`. They can be used for consistent an
...
+
+Head on over to the [Ignite Cookbook](https://ignitecookbook.com/) to find recipes for how to integrate Ignite's theming system with other popular styling and component libraries!
+
+:::
+
+In Ignite `<10` we relied on pure style objects typed as `ViewStyle`, `TextStyle`, `ImageStyle`, etc. This didn't quite fit circumstances where you wanted to dynamically change the overall app's theme during runtime. You could always reload the javascript bundle to see changes, but that was a less than ideal solution.
+
+Ignite now has the concept of `ThemedStyle`s. You can stick with the traditional pure style objects but if you want to take advantage of themed global colors, spacing, typography, and timings, you've got to add a little more code...
+
+Here's an example of how to use the new `useAppTheme()` hook to style your components and dynamically use the correct theme:
```
lindboe, line 49:
```suggestion
**New in Ignite 10**: the `useAppTheme` hook allows you to create dynamically-themed styles, right out of the box. Here's an example:
```

Rationale for suggestion: most people are probably not going to be using Ignite multiple times or paying attention to versions; I think it's sufficient to explain that there's now a new feature and how to use it.

docs/boilerplate/app/theme/Theming.md
```diff
@@ -31,3 +31,151 @@ Timings are defined in `app/theme/timing.ts`. They can be used for consistent an
...
+    useThemeProvider()
+  return (
+    <ThemeProvider value={{ themeScheme, setThemeContextOverride }}>
+      <Component />
+    </ThemeProvider>
+  )
+}
+```
+
+For more information on the `useAppTheme()`, [check out its documentation](../utils/useAppTheme.tsx.md).
```
lindboe, line 92:
```suggestion
For more information on the `useAppTheme()` hook, [check out its documentation](../utils/useAppTheme.tsx.md).
```

I also wouldn't mind just linking the docs in the mention of the hook above the example code.

docs/boilerplate/app/utils/Utils.md
```diff
@@ -13,6 +13,24 @@ Every app needs a junk drawer. Here you can find a library of utilities that are
...
+const $themedStyle: ThemedStyle<ViewStyle> = (theme) =({
+  backgroundColor: theme.colors.background
+})
+
+<View style={themed($themedStyle)}>
+  <Text>{themeContext}</Text>
+</View>
+```
+
+[Full `useAppTheme`](./useAppTheme.ts.md)
```
lindboe, line 32:
```suggestion
[Full `useAppTheme`](./useAppTheme.tsx.md)
```

Make sure to test the docs out with ir-docs to make sure they work, this red-screened for me

docs/boilerplate/app/utils/useAppTheme.tsx.md
```diff
@@ -0,0 +1,131 @@
...
+<View style={themed([$themedStyle, $plainStyle])} />
+<View style={themed([
+  $themedStyle,
+  $plainStyle,
+  // A plain style object
+  { width: 7, backgroundColor: 'yellow'}
+])} />
+```
+
+Make sure you don't pass any `Animated` styles to `themed()`. It will not work as expected! Keep them in separate style array objects: `<Animatd.View style={[$animatedStyle, themed($myThemedStyle)]}>`.
```
lindboe, line 131:
```suggestion
Make sure you don't pass any `Animated` styles to `themed()`. It will not work as expected! Keep them in separate style array objects: `<Animated.View style={[$animatedStyle, themed($myThemedStyle)]}>`.
```

boilerplate/app/theme/spacingDark.ts
markrickert: Yeah, so the `theme/spacing.ts` is the ORIGINAL spacing file from the theme. `theme/spacingDark.ts` is a modified theme that has 1/2 the spacing of the light theme.

Current: https://github.com/infinitered/ignite/blob/master/boilerplate/app/theme/spacing.ts
Themed: https://github.com/infinitered/ignite/blob/mrickert/theming/boilerplate/app/theme/spacing.ts
Themed Dark (smaller): https://github.com/infinitered/ignite/blob/mrickert/theming/boilerplate/app/theme/spacingDark.ts

So the light mode is what ignite used to be and dark mode is a little tighter.

The problem of the submenus not showing in your first video has been fixed in a new push.

docs/Guide.md
markrickert: Fixed! Thank you!

boilerplate/app/screens/DemoShowroomScreen/DrawerIconButton.tsx
markrickert: I haven't been able to replicate this myself. The animated styles are using the colors returned by the useAppTheme() to determine the color of the bars.

boilerplate/package.json
lindboe: @frankcalise FYI I figured this out; CI works because it's using bun as the package manager.

> Running bun install will: Install all dependencies, devDependencies, and optionalDependencies. Bun will install peerDependencies by default.

https://bun.sh/docs/cli/install

If yarn's more likely to fail at steps like that, should we switch the test to use yarn instead of bun? or test both?

lindboe: I also don't love this behavior for bun, people don't pay close attention to `peerDependencies` because it's often useless and especially in RN libs, I see out-of-date versions referenced.

frankcalise: I'm not in love with bun either, I think it got switched along the way to speed up the tests. But if they're not working correctly with it than I'm open to switching it back as it only makes sense.

Bun version on the CI is probably pinned from a while ago, so it might also be worth trying with a more updated version (but if I recall, we were following the version found on [EAS Servers](https://docs.expo.dev/build-reference/infrastructure/))

docs/boilerplate/app/utils/useAppTheme.tsx.md
```diff
@@ -0,0 +1,130 @@
+---
+sidebar_position: 10
+---
+
+# useAppTheme()
```
lindboe, line 5:
```suggestion
# useAppTheme
```

nit, I just noticed it looks inconsistent in the sidebar:

Screenshot 2024-08-06 at 4 06 41 PM

boilerplate/app/components/TextField.tsx
```diff
@@ -259,21 +266,22 @@ const $inputStyle: TextStyle = {
...
+})
 
-const $helperStyle: TextStyle = {
+const $helperStyle: ThemedStyle<TextStyle> = ({ spacing }) => ({
   marginTop: spacing.xs,
-}
+})
 
-const $rightAccessoryStyle: ViewStyle = {
+const $rightAccessoryStyle: ThemedStyle<ViewStyle> = ({ spacing }) => ({
```
lindboe, line 275: This was updated, but a spot that consumed it wasn't wrapped with `themed()`. This caused the login text box to still have incorrect alignment for the visiblity icon: 

Screenshot 2024-08-06 at 4 15 35 PM

Adding `themed()` fixes this, but I wonder if we should add more aggressive type-checking to avoid passing a style function to components that don't expect them. @markrickert, thoughts on this?

markrickert: I think the reason typescript didn't complain bec of this line: https://github.com/infinitered/ignite/blob/8b347a080fb2ea8972c85d4363ba7457603e9ad8/boilerplate/app/components/TextField.tsx#L16

I wonder if changing the `style: StyleProp<any>` to this would have any unintended consequences? 

```tsx
export interface TextFieldAccessoryProps {
  style: StyleProp<ViewStyle | TextStyle | ImageStyle>
  status: TextFieldProps["status"]
  multiline: boolean
  editable: boolean
}
```

lindboe: That seems like a reasonable fix to me for now. It would be ideal to try to use the style prop from the actual component used but I was just poking at that for a while and haven't figured out anything I'm happy with (without changing how TextField is architected)

markrickert: I've made and verified the fix:

Screenshot 2024-08-08 at 2 24 52 PM

lindboe:
@markrickert I do think we should address https://github.com/infinitered/ignite/pull/2636#issuecomment-2261960598. It not only makes the demo look better, but there are likely to be other use cases where somebody needs to set a theme on non-React components (i.e., imperatively on a native component).

I took a stab at it:

https://github.com/user-attachments/assets/35e79a49-9813-41df-861f-ad9876ea19c7

Here's my diff:

```
diff --git a/app/utils/useAppTheme.ts b/app/utils/useAppTheme.ts
index 8d03a0c..712bf81 100644
--- a/app/utils/useAppTheme.ts
+++ b/app/utils/useAppTheme.ts
@@ -1,4 +1,4 @@
-import { createContext, useCallback, useContext, useMemo, useState } from "react"
+import { createContext, useCallback, useContext, useEffect, useMemo, useState } from "react"
 import { StyleProp, useColorScheme } from "react-native"
 import { DarkTheme, DefaultTheme, useTheme as useNavTheme } from "@react-navigation/native"
 import {
@@ -9,6 +9,7 @@ import {
   lightTheme,
   darkTheme,
 } from "app/theme"
+import * as SystemUI from 'expo-system-ui';

 type ThemeContextType = {
   themeScheme: ThemeContexts
@@ -23,6 +24,13 @@ export const ThemeContext = createContext<ThemeContextType>({
   },
 })

+const themeContextToTheme = (themeContext: ThemeContexts): Theme => (
+    themeContext === "dark" ? darkTheme : lightTheme)
+
+const setImperativeThemeing = (theme: Theme) => {
+  SystemUI.setBackgroundColorAsync(theme.colors.background)
+}
+
 export const useThemeProvider = (initialTheme: ThemeContexts = undefined) => {
   const colorScheme = useColorScheme()
   const [overrideTheme, setTheme] = useState<ThemeContexts>(initialTheme)
@@ -34,6 +42,10 @@ export const useThemeProvider = (initialTheme: ThemeContexts = undefined) => {
   const themeScheme = overrideTheme || colorScheme || "light"
   const navigationTheme = themeScheme === "dark" ? DarkTheme : DefaultTheme

+  useEffect(() => {
+    setImperativeThemeing(themeContextToTheme(themeScheme))
+  }, [overrideTheme])
+
   return {
     themeScheme,
     navigationTheme,
@@ -78,7 +90,7 @@ export const useAppTheme = (): UseAppThemeValue => {
   )

   const themeVariant: Theme = useMemo(
-    () => (themeContext === "dark" ? darkTheme : lightTheme),
+    () => themeContextToTheme(themeContext),
     [themeContext],
   )
```

My only concern here is that we haven't documented using `useThemeProvider`, which is what would run this. Is that meant to be a required part of the theming system?

lindboe (approved):
It's looking good to me, just would like your thoughts on the addition of the native background color setting

boilerplate/app/utils/useAppTheme.ts
```diff
@@ -0,0 +1,109 @@
...
+  )
+
+  const themeVariant: Theme = useMemo(
+    () => (themeContext === "dark" ? darkTheme : lightTheme),
+    [themeContext],
+  )
+
+  const themed = useCallback(
+    <T>(styleOrStyleFn: ThemedStyle<T> | StyleProp<T> | ThemedStyleArray<T>) => {
+      const flatStyles = [styleOrStyleFn].flat(3) as (ThemedStyle<T> | StyleProp<T>)[]
```
lindboe, line 87: I just checked if this cast was necessary, I removed it and am not seeing any issues, okay to remove @markrickert ?

markrickert: Nice find! Lets get rid of it

lindboe:
Pushed commit with changes to set native background color after discussion with @markrickert https://github.com/infinitered/ignite/pull/2636/commits/4fb4d6127d63aba9790fa396fbc3011a38898dfe

frankcalise (approved):
🚢 

nice work everyone!

PR #2647, opened 2024-02-21
Remove MobX-State-Tree flag for non-demo projects

Jpoliachik:
This PR adds support for removing MobX-State-Tree code from a non-demo project. 
Retargeted to v10 branch. 

Changes: 
- `// @mst` inline comments, similar to how demo code removal works
- Refactored `demo.ts` removal logic into more generic `markup.ts` to be shared with `@mst` comment types too, and make it easy to add additional other types in the future too. 
- Added new `replace-next-line` markup command type, to help swap out lines of code easily without hot patching in scripts (keeps code replacements inline next to other code)

The option shows up --removeDemo=true is set
image


src/commands/new.ts
```diff
@@ -144,6 +146,13 @@ export interface Options {
    * @default false
    */
   noTimeout?: boolean
+  /**
+   * Whether or not to remove MobX-State-Tree boilerplate code
+   *
+   * Input Source: `prompt.ask` | `parameter.option`
+   * @default false
+   */
+  removeMST?: boolean
```
jamonholmgren, line 155: Despite all other evidence above, we've been trying to go away from boolean options (see the `experimental` flag).

So, maybe consider doing something like:

```
state?: "mst" | "none"
```

Open to other opinions, though. [GAFO](https://capwatkins.com/blog/the-sliding-scale-of-giving-a-fuck) level is 3/10.

src/commands/remove-mst.ts
```diff
@@ -0,0 +1,120 @@
...
+
+export const mstDependenciesToRemove = [
+  "mobx",
+  "mobx-react-lite",
+  "mobx-state-tree",
+  "reactotron-mst",
+]
+
+module.exports = {
+  alias: ["rmst", "remove-mst"],
```
jamonholmgren, line 17: maybe `rm-mst`

src/commands/new.ts
jamonholmgren: Actually looking at the rest of the code, perhaps `removeMST` is fine?

Jpoliachik: makes sense, yeah I'll roll with that!

Jpoliachik: Yeah I had a similar thought. `state` sounds like it opens the door for potentially supporting other state options, which we don't want due to maintenance effort. 
And the code itself is similar enough to the `removeDemo` code that I thought it was a decent pattern to copy. 
I might vote for `removeMST` but also low gafo

src/commands/new.ts
```diff
@@ -372,6 +381,23 @@ module.exports = {
...
 
+    // #region Prompt to Remove MobX-State-Tree code
+    // only ask if we're removing the demo code
+    const defaultRemoveMST = false
+    let removeMST = useDefault(options.removeMST) ? defaultRemoveMST : boolFlag(options.removeMST)
+    if (removeDemo && removeMST === undefined) {
+      const removeMSTResponse = await prompt.ask<{ removeMST: boolean }>(() => ({
+        type: "confirm",
+        name: "removeMST",
+        message: "Remove MobX-State-Tree code? We recommend leaving it in",
```
frankcalise, line 392: 🙌 I see you `MobX`

src/commands/remove-mst.ts
```diff
@@ -0,0 +1,120 @@
...
+    ]
+    const transformPath = pathlib.join(__dirname, "../tools/remove-mst-transformer.ts")
+    const paths = filesToProcess.map((file) => pathlib.join(TARGET_DIR, file))
+    const options = {
+      dry: dryRun,
+      print: false,
+      importsToRemove: mstDependenciesToRemove,
+    }
+    const res = await jscodeshift(transformPath, paths, options)
+    console.log(res)
```
frankcalise, line 53: ✂️

src/commands/remove-mst.ts
```diff
@@ -0,0 +1,120 @@
...
+    const { parameters } = toolbox
+
+    const CWD = process.cwd()
+    const TARGET_DIR = parameters.first ?? CWD
+    const dryRun = boolFlag(parameters.options.dryRun) ?? false
+
+    // TODO: full dryRun support
+
+    p()
+    p(`Removing MobX-State-Tree code from '${TARGET_DIR}'${dryRun ? " (dry run)" : ""}`)
```
frankcalise, line 30: What about a try/catch here and if running into any issues, let's warn the user the automated removal could not be completed and refer them to the recipe to do so manually?

src/commands/new.ts
frankcalise: We could also just call the flag `--mst` and default it to true

This way it reads `--mst=false` when you want to opt out via cli (or you hit the question and `--mst=false` will be output on the generated command to run the project generation again).

And then you don't have anything called `state` misleading the user that we'll be open to supporting others out of the box, etc

frankcalise (commented):
First pass, it's taking shape nice work!

frankcalise:
@Jpoliachik can you target this against `v10` as the base branch?

Jpoliachik:
@frankcalise sure thing, I'll work on this early next week 👍

src/commands/new.ts
Jpoliachik: I ended up using `--mst` flag and defaulting to true 👍

boilerplate/app/models/helpers/useStores.ts
```diff
@@ -63,6 +63,7 @@ export const useInitialRootStore = (callback?: () => void | Promise<void>) => {
 
       // reactotron integration with the MST root store (DEV only)
       if (__DEV__) {
+        // @ts-ignore
         console.tron.trackMstNode(rootStore)
```
Jpoliachik, line 67: not sure why this wasn't needed before, but I was getting typescript errors without.

src/tools/markup.ts
```diff
@@ -0,0 +1,299 @@
...
+        })
+
+        if (!dryRun) {
+          await update(path, (contents: string) => {
+            let result = contents
+            if (removeMarkupOnly) {
+              result = sanitize(result)
+            } else {
+              result = updateFile(result, markupPrefix)
+              result = sanitize(result)
```
Jpoliachik, line 287: I don't know if this extra `sanitize()` was happening before - what behavior do we want here, I assumed we'd always want to get rid of all the markup regardless

docs/cli/Ignite-CLI.md
```diff
@@ -98,6 +98,7 @@ Starts the interactive prompt for generating a new Ignite project. Any options n
 - `--overwrite` overwrite the target directory if it exists
 - `--targetPath` string, specify a target directory where the project should be created
 - `--removeDemo` will remove the boilerplate demo code after project creation
+- `--mst` flag to specify whether to include MobX-State-Tree in project
```
lindboe, line 101: I think we should document that this only works when also using `--removeDemo`, I was confused by this at first

src/commands/help.ts
```diff
@@ -60,6 +60,18 @@ module.exports = {
...
     )
     p()
+    command(
+      "remove-mst (rm-mst)",
+      "Removes MobX-State-Tree code from the project (add --dry-run to list changes but not execute)",
+      ["npx ignite-cli remove-mst", "npx ignite-cli remove-mst --dry-run"],
+    )
+    p()
+    command(
+      "remove-mst-markup (rmstm)",
```
lindboe, line 70: From what I can tell, removal of MST, and removal of MST markup can only be done when generating a new project, since either way the markup won't exist after generation, right?

If that's the case, should we consider these commands "internal" and not document them here, since they can't be used by end users? Or at least add that information in to the docs for context.

src/commands/remove-mst.ts
lindboe: A suggestion: if we want to refer people to document officially, I think that document should be docs in Ignite, and not a cookbook recipe.

boilerplate/app/app.tsx
```diff
@@ -73,6 +73,7 @@ function App(props: AppProps) {
 
   const [areFontsLoaded] = useFonts(customFontsToLoad)
 
+  // @mst replace-next-line React.useEffect(() => {
   const { rehydrated } = useInitialRootStore(() => {
     // This runs after the root store has been initialized and rehydrated.
```
lindboe, line 78: This comment is no longer true after removing MST, should probably also be removed.

test/vanilla/__snapshots__/ignite-remove-demo.test.ts.snap
```diff
@@ -3,40 +3,66 @@
 exports[`ignite-cli remove-demo should print the expected response 1`] = `
 "   
    Removing demo code from '/user/home/ignite'
-   Found '@demo remove-file' in /user/home/ignite/.maestro/FavoritePodcast.yaml
-   Found '@demo remove-file' in /user/home/ignite/.maestro/Login.yaml
+removing file /user/home/ignite/.maestro/FavoritePodcast.yaml
```
lindboe, line 6: I think we should have a test like this for the addition of the --mst=false flag as well just to ensure that continues working

boilerplate/app/app.tsx
```diff
@@ -73,6 +73,7 @@ function App(props: AppProps) {
 
   const [areFontsLoaded] = useFonts(customFontsToLoad)
 
+  // @mst replace-next-line React.useEffect(() => {
```
lindboe, line 76: nit: something got a little funky here, once this run I see the two spaces on a separate line from the replaced line here:

Screenshot 2024-07-30 at 2 36 12 PM

There might be a way to fix this, I don't know this markup system yet, but I was also going to suggest we also acll prettier at the end of `remove-mst` like we do for `remove-mst-markup`

src/commands/help.ts
lindboe: I can't think of any reason, no, @frankcalise ?

boilerplate/app/app.tsx
Jpoliachik: good call! I noticed a couple of these spots, so I added a prettier step 👍

src/commands/remove-mst.ts
Jpoliachik: I added a link to the recipe in the existing try/catch as part of the `new` flow 👍 
I think since it's a tutorial and intended as a backup if this fails, it makes sense to keep it in Cookbook, IMO. 
image

Jpoliachik:
This is ready for additional review. 

As far as CLI documentation, I added the `--mst` flag option to the list, and added a blurb under the MST page. I didn't think much else was necessary, since this flag should be straightforward for users, but i'm open to opinions if we think other docs are needed!

src/commands/remove-mst.ts
lindboe: The main reason I say documentation and not a cookbook recipe was that the original intention of cookbook was it to have less obligation to keep it up-to-date (and therefore encourage contributions).

Linking to things officially from here makes it seem more like documentation, to me, which also implies it should continue to be tested with Ignite updates and stay accurate. Drawing a clear line on what we want to keep up-to-date and what's "this worked at one point" seems valuable to me.

Medium GAFO on this, definitely open to disagreement, but that's the thinking behind it.

lindboe (approved):
LGTM, just a small conflict now after the v10 branch has been updated

Thanks for this! 🌟 🥳

src/commands/help.ts
frankcalise: I think the reasoning was if you had an existing project where the markup wasn't removed (for the demo markup that is). We didn't always remove the markup out of the box.

That was probably around v8 though and fixed in or around some patch of 9 - we could probably look to remove it as never leaving mark up is what's current.

If we want to address the demo one as a separate PR that's fine

src/commands/remove-mst.ts
Jpoliachik: I'm going to leave as a Cookbook link for now for simplicity, but if we find that the cookbook recipe falls out of date or if we want to integrate the MobX Removal instructions into official Ignite docs in the future, we can make those changes separately

PR #2718, opened 2024-07-30
Add a contributing home page

lindboe:
This entry point is intended for quick instruction on how to accomplish development tasks, and then links out to deeper guides on contributing from there. Adding a "Contributing to Ignite" title should also make it easier to search from Algolia amidst the other "Contributing" pages.

Not sure why I had to add that full path to _category_.json when that wasn't there before, but this was the only way to make it work locally with ir-docs.

Screenshot 2024-07-30 at 4 17 41 PM


markrickert:
This is great! Do you think we should link to the contributing guide in the main readme?

frankcalise (approved):
Thanks for this, looks good!

As far the category json thing I'll try it out locally tomorrow, unsure why that would be 🤔

lindboe:
@markrickert good call, I'll add that!

lindboe:
@frankcalise I figured it out, it's case-sensitive and I just missed the difference. Found this in my git reflog: 

```
diff --git a/docs/contributing/_category_.json b/docs/contributing/_category_.json
index 61cbad9c..34a89ce7 100644
--- a/docs/contributing/_category_.json
+++ b/docs/contributing/_category_.json
@@ -3,6 +3,6 @@
   "position": 99,
   "link": {
     "type": "doc",
-    "id": "Contributing-to-Ignite"
+    "id": "ignite-cli/contributing/Contributing-To-Ignite"
   }
```

Issue #2351, opened 2023-01-10
Bug with Screen component's keyboard avoiding behaviour


dakshshah96:
### Describe the bug

I was trying to use the `Screen` component in my Ignite app and noticed that the keyboard avoiding behaviour for text inputs doesn't work perfectly. On focusing on a text input, there is a slight scroll which takes place but not enough to view the text input.

The behaviour I noticed also exists in the Ignite boilerplate app, so I'll skip adding a reproducible example here in favour of a screen recording of the boilerplate app.

https://user-images.githubusercontent.com/7896438/211507890-0aeabca5-f3a3-43b3-a8f2-f74218c7aa36.MP4

### Ignite version

8.4.5

### Additional info

```
System
  platform           darwin
  arch               arm64
  cpu                10 cores      Apple M1 Max
  directory          boilerplate   /Users/daksh/Documents/Projects/IgniteTest/boilerplate

JavaScript (and globally-installed packages)
  node                  16.15.0                                   /Users/daksh/Library/pnpm/node
  npm                   7.24.2                                    /Users/daksh/Documents/Projects/IgniteTest/node_modules/.bin/npm
    corepack            0.10.0
    npm                 8.18.0
  yarn                  1.22.19                                   /opt/homebrew/bin/yarn
    apk-mitm            1.1.0
    aws-cdk             2.15.0
    cordova-res         0.15.4
    create-next-app     12.0.8
    create-vue          3.4.0
    dependents          0.2.0
    eas-cli             3.1.1
    eslint              8.16.0
    expo-cli            6.0.8
    lerna               4.0.0
    madge               5.0.1
    netlify-cli         9.12.3
    npm-check-updates   12.5.2
    vercel              28.2.2
  pnpm                  7.22.0                                    /opt/homebrew/bin/pnpm
                        <no pnpm global package info available>
  expo                  46.0.19                                   managed

Ignite
  ignite-cli         8.4.5        /Users/daksh/.npm/_npx/e31027f3785124a8/node_modules/.bin/ignite
  ignite src         build        /Users/daksh/.npm/_npx/e31027f3785124a8/node_modules/ignite-cli/build

Android
  java               19.0.1       /usr/bin/java
  android home       -            /Users/daksh/Library/Android/sdk

iOS
  xcode              14.2
  cocoapods          1.11.3       /opt/homebrew/bin/pod

Tools
  git                git version 2.39.0   /opt/homebrew/bin/git
```

flexbox:
[...]

PR #2722, opened 2024-08-01
Improve keyboard avoid behavior for section list and Screen component with "fixed" preset variant

fpena:

## Describe your PR

This PR solves an issue with text inputs inside a list (e.g. `SectionList`, `Flatlist`, etc). It was first reported [here](https://github.com/infinitered/ignite/issues/2351).

## Overall behaviour 

Here are also videos with examples of variants for `Screen.tsx` and its `preset` prop.

| `preset="fixed"` - items at top   | `preset="fixed"` - items at bottom |  `preset="scroll"` - items at top | `preset="scroll"` - items at bottom | 
|----------|:-------------:|------:|------:|
| Simulator Screen Recording - iPhone SE (2nd generation) - 2024-08-05 at 09 02 54 |  Simulator Screen Recording - iPhone SE (2nd generation) - 2024-08-05 at 09 16 30 | Simulator Screen Recording - iPhone SE (2nd generation) - 2024-08-05 at 09 17 35 | Simulator Screen Recording - iPhone SE (2nd generation) - 2024-08-05 at 09 18 51


boilerplate/app/components/Screen.tsx
```diff
@@ -269,6 +269,10 @@ const $outerStyle: ViewStyle = {
   width: "100%",
 }
 
+const $justifyFlexEnd: ViewStyle = {
```
fpena, line 272: It's not in scope for [the bug](https://github.com/infinitered/ignite/issues/2351), but when it's non-scrolling, it's better to add `justify-content: flex-end` to make keyboard avoid behaviour work properly like in this video. 

https://github.com/user-attachments/assets/d9360f5c-e73c-4fff-a342-200f7ff10f54

boilerplate/app/screens/DemoShowroomScreen/DemoShowroomScreen.tsx
```diff
@@ -181,6 +182,12 @@ export const DemoShowroomScreen: FC<DemoTabScreenProps<"DemoShowroom">> =
             stickySectionHeadersEnabled={false}
             sections={Object.values(Demos)}
             renderItem={({ item }) => item}
+            renderScrollComponent={({ contentContainerStyle }) => (
+              <KeyboardAwareScrollView
+                contentContainerStyle={contentContainerStyle}
+                bottomOffset={50}
+              />
+            )}
```
fpena, line 190: By adding this, we get normal keyboard avoid behavior within a `SectionList`. From here, we can do a couple of things: 

* Extract 50 to a default constant for bottom offset
* Create a new component (e.g. `SectionListWithKeyboardAwareScrollView` that brings this behavior by default
* Add more documentation overall in Ignite about keyboard avoid behavior for SectionList, FlatList FlashList, etc.

fpena: Here's also a demo:


https://github.com/user-attachments/assets/48ec3cca-967b-4dd9-99de-3baaea676edb

lindboe:
Tried this out on Android and iOS and it looks great 👍

boilerplate/app/screens/DemoShowroomScreen/DemoShowroomScreen.tsx
lindboe: Do you know if we can use this on the `LoginScreen` as well to ensure the "Tap to sign in!" button stays visible? or is that maybe too complicated for now?

fpena: @lindboe yes, it's possible. Check this video [here](https://github.com/infinitered/ignite/pull/2722#discussion_r1700460133).

lindboe: Oh I see, I missed that's what that video was doing, I saw the multiline input and thought it was still the demo screen 👍

boilerplate/package.json
```diff
@@ -79,6 +80,9 @@
     "eslint": "8.17.0",
     "eslint-config-prettier": "8.5.0",
     "eslint-config-standard": "17.0.0",
+    "eslint-plugin-import": "2.26.0",
+    "eslint-plugin-n": "^15.0.0",
+    "eslint-plugin-promise": "6.0.0",
```
fpena, line 85: From [Mark's commit](https://github.com/infinitered/ignite/pull/2636/commits/6476e06ecda7f161c006a06d473fca6544f316f9).

frankcalise (commented):
Nice work! This looks good.

Is the plan to also use the `KeyboardAwareScrollView` from `react-native-keyboard-controller` in `Screen.tsx` for `ScreenWithScrolling`? Or is that going to be a separate PR?

Here is an example we did at the intermediate CR workshop that Ellie came up with: https://github.com/infinitered/cr-2024-intermediate-workshop-template/blob/02-inputs-solution/src/components/Screen.tsx#L181-L223

Probably useful to bring that into the boilerplate if it fits (of course the Showroom needed it's own section list, so that makes sense how you implemented it).

If you need a screen to test, you can utilize [this one](https://github.com/infinitered/cr-2024-intermediate-workshop-template/blob/02-inputs-solution/src/app/(app)/(tabs)/profile.tsx) (but it has some extra dependencies you'll want to comment out or just make a screen in place of DebugScreen with a bunch of text fields and Screen preset of `scroll`)

Then would help alleviate pain in not only the `fixed` preset but `scroll` as well.

fpena:
> Is the plan to also use the KeyboardAwareScrollView from react-native-keyboard-controller in Screen.tsx for ScreenWithScrolling? Or is that going to be a separate PR?

@frankcalise Not necessarily. I started by modifying `Screen.tsx` but stopped because I was naive to the `react-native-keyboard-controller` behaviour. Then I realized the solution was to add the `renderScrollComponent` prop in `SectionList`.

I think a more generic solution is needed for `Screen.tsx` if we want to support not only `SectionList`, but also `FlashList` and `FlatList`.

Let me know your thoughts!

frankcalise:
Hmm. I think it's possible Screen (with scrolling) and handling the lists are actually separate since you wouldn't want to nest scrollviews (one being screen, the other coming from the list).

But to that point, yeah maybe something can be done in app/components/ListView for the lists aspect of the fix

boilerplate/app/components/Screen.tsx
```diff
@@ -194,7 +197,8 @@ function ScreenWithScrolling(props: ScreenProps) {
   useScrollToTop(ref)
 
   return (
-    <ScrollView
+    <KeyboardAwareScrollView
+      bottomOffset={DEFAULT_BOTTOM_OFFSET}
```
frankcalise, line 201: I think this should come in from a prop and we can set this default if it makes sense, but right now this will just always be 50 correct?

frankcalise (commented):
Thanks for updating screen! Question about the hardcoded offset, but I may be missing something so LMK :)

boilerplate/app/components/Screen.tsx
fpena: @frankcalise added new prop!

lindboe:
I'm seeing a weird issue on iOS, I reproduced on both simulator and physical device. It happens whether keyboard is open or not:

1. On demo showroom screen, open hamburger menu
2. Scroll all the way down the hamburger menu and select the last item ("Styling")
3. After screen animates to end, try to scroll up again. it will glitch and keep trying to scroll you back down.


https://github.com/user-attachments/assets/b25808cf-87b7-46df-a507-ddc3d227dfee

boilerplate/app/components/Screen.tsx
frankcalise: @fpena Great thanks for doing that!

frankcalise (commented):
This is looking good! Nice work!

What are your thoughts on extending `src/components/ListView.tsx` similar to the way you did the demo section list?

Does that make sense or only if the list in not embedded in a `<Screen />` at this point? Which I would say I typically reach for when letting the list drive the entire screen area.

fpena:
> This is looking good! Nice work!
> 
> What are your thoughts on extending `src/components/ListView.tsx` similar to the way you did the demo section list?
> 
> Does that make sense or only if the list in not embedded in a `<Screen />` at this point? Which I would say I typically reach for when letting the list drive the entire screen area.

@frankcalise I did a quick update to `ListView` to add `renderScrollComponent` and I'm getting an error related to MobX. I feel the scope of this PR is increasing. Would you mind if I create an extra issue to tackle that? Let me know what do you think!

boilerplate/package.json
```diff
@@ -56,8 +56,9 @@
     "react-native": "0.74.3",
     "react-native-drawer-layout": "4.0.0-alpha.9",
     "react-native-gesture-handler": "~2.16.1",
+    "react-native-keyboard-controller": "^1.12.7",
     "react-native-mmkv": "^2.12.2",
-    "react-native-reanimated": "~3.10.1",
+    "react-native-reanimated": "~3.15.0",
```
fpena, line 61: Upgrade `react-native-reanimated` to fix [the issue](https://github.com/infinitered/ignite/pull/2722#issuecomment-2272362409) Lizzi reported.

frankcalise:
@fpena yeah I agree, we can take care of that separately

lindboe:
Preliminary impression: the bug no longer occurs, but the scrolling animation after menu selection on my Android device (Galaxy S21) is not smooth, and I think it was before. I want to try it out in release mode to see if it's any better, and again on the older versions to compare.

I can try that out after lunch.

frankcalise:
Fixed feels good on iOS (sim, haven't tried device just yet) and Android device.

The section list feels good on the iOS sim but not working as great on Android device? (Pixel 4).

As you can see my test for this is bringing the TextInput very low and some portion of behind the tab bar (although it makes no difference, as long as it is in the bottom half of the screen where the keyboard would be upon being shown) to make sure it has to adjust.

https://github.com/user-attachments/assets/61959b67-dc23-41b9-8e5b-e46259be3224

lindboe:
Okay, I just compared with a plain v10 repro and I don't see a difference. Would be ideal to get this smoother but is fine for now since this list is so big 👍  attaching videos for reference

https://github.com/user-attachments/assets/6012638a-52db-4c94-9ba1-5e0c534d7e04


https://github.com/user-attachments/assets/29880cd4-d878-4e9c-ac5e-20092775dacd

boilerplate/app/screens/DemoShowroomScreen/SectionListWithKeyboardAwareScrollView.tsx
```diff
@@ -0,0 +1,60 @@
...
+  name: string
+  description: string
+  data: ItemType[]
+}
+
+type SectionListWithKeyboardAwareScrollViewProps<ItemType> = SectionListProps<ItemType> & {
+  /* Optional function to pass a custom scroll component */
+  renderScrollComponent?: (props: ScrollViewProps) => React.ReactNode
+  /* Optional bottom offset for the keyboard avoid behavior */
+  bottomOffset?: number
```
lindboe, line 16: Since this passing along a react-native-keyboard-controller prop, I think linking to the docs in the comment is useful, something like:

```
  /* Optional additional offset between TextInput bottom edge and keyboard top edge. See https://kirillzyusko.github.io/react-native-keyboard-controller/docs/api/components/keyboard-aware-scroll-view#bottomoffset */
  bottomOffset?: number

```

lindboe (approved):
Looking good! I suggested a few minor things. Thanks for this @fpena 👏 🥳

frankcalise:
Adding a video from a Pixel 7 device as well. Similar to the [Pixel 4](https://github.com/infinitered/ignite/pull/2722#issuecomment-2276541293) video above, except the Login screen also showed some weird side effects (where it didn't, on the smaller screen size)

frankcalise (approved):
Great work! Android is functioning well now!!

boilerplate/app/screens/DemoShowroomScreen/DemoShowroomScreen.tsx
```diff
@@ -181,22 +183,30 @@ export const DemoShowroomScreen: FC<DemoTabScreenProps<"DemoShowroom">> =
...
+        <Screen
+          preset="fixed"
+          safeAreaEdges={["top"]}
+          contentContainerStyle={$styles.flex1}
+          {...(isAndroid ? { KeyboardAvoidingViewProps: { behavior: undefined } } : {})}
+        >
           <DrawerIconButton onPress={toggleDrawer} />
 
-          <SectionList
+          <SectionListWithKeyboardAwareScrollView<ReactElement[]>
```
lindboe, line 194: I just noticed this, why does this tag appear to have a type annotation? I've never seen this with JSX before

lindboe: Addressed in https://github.com/infinitered/ignite/pull/2722/commits/b1c1525bc37afc14950aed71d13c8ca8dc8d6419.

After investigation, I found that it's typical for devs consuming `SectionList` to allow `ItemT` to remain as `any` and narrow types in `renderItem`. To support that, I updated the component to default to `any` for that type.

PR #2714, opened 2024-07-25
feat(cli): experimental expo-router flag

frankcalise:
## Please verify the following:

- [ ] `yarn test` **jest** tests pass with new tests, if relevant
- [ ] `yarn lint` **eslint** checks pass with new code, if relevant
- [ ] `yarn format:check` **prettier** checks pass with new code, if relevant
- [ ] `README.md` (or relevant documentation) has been updated with your changes

## Describe your PR
- Adds `expo-router` value to the `--experimental` switch
- Ignites a new app (no demo app available, though) and makes all the updates as per the [Cookbook Recipe](https://ignitecookbook.com/docs/recipes/ExpoRouter/) automatically
- Quick and dirty right now, but I got it out of my brain and it works

## TODO
- [ ] Tests? xd
- [ ] Deploy it to `next` label for in the wild testing
- [ ] Remove the one Reactotron command about resetting nav state
- [ ] Update generators (`model` and `component` as `navigator` and `screen` will vanish) - that being said the cli probably still needs [some more support for sub dir paths on `generate` command](https://github.com/infinitered/ignite/issues/2708)

PR #2726, opened 2024-08-04
feat(cli): generator options --dir and --case 

### LINKED ISSUE #2708: revise generator to use different paths
When the generator is run, it references certain paths which are different after converting to Expo Router ([link](https://ignitecookbook.com/docs/recipes/ExpoRouter/)).

Relevant frontmatter in the ejs file
```
---
patches:
- path: "app/screens/index.ts"
  append: "export * from \"./<%= props.pascalCaseName %>Screen\"\n"
  skip: <%= props.skipIndexFile %>
- path: "app/navigators/AppNavigator.tsx"
  replace: "// IGNITE_GENERATOR_ANCHOR_APP_STACK_PARAM_LIST"
  insert: "<%= props.pascalCaseName %>: undefined\n\t// IGNITE_GENERATOR_ANCHOR_APP_STACK_PARAM_LIST"
- path: "app/navigators/AppNavigator.tsx"
  replace: "{/* IGNITE_GENERATOR_ANCHOR_APP_STACK_SCREENS */}"
  insert: "<Stack.Screen name=\"<%= props.pascalCaseName %>\" component={Screens.<%= props.pascalCaseName %>Screen} />\n\t\t\t{/* IGNITE_GENERATOR_ANCHOR_APP_STACK_SCREENS */}"
  skip: <%= props.skipIndexFile %>
---
```

When I brought it up, modifying the generator to allow for changing the paths used was brought up.

Is that something you are wanting to consider? If so, I'm happy to help. Otherwise, we can close this issue.

frankcalise:

## Goal
The generator has a path hardcoded in it, `app/`. While this is the current file structure of most Ignite projects, Expo Router becoming more popular (and some client project desired structures even without it) and uses this directory for the file-based routing system (that and `src/app/`).

The goal of these changes is to leverage Ignite generators but be flexible enough to put something in `src/app/(auth)/(tabs)/` for example, since there isn't really the pattern of a flat `screens/` directory in such projects. This will especially be nice when the functionality in #2714 lands as well, which lets you start an Ignite project with Expo Router.

## Describe your PR
- Closes #2708 
- Adds two new cli options for generators
  - `--dir` specifies where the output files should be generated (will override default app/ path and front matter if specified)
  - `--case=none` specifies that the filename should be exactly as typed in
- Updates documentation for the new options
- Updates help on generate command to display `--case` and `--dir`
- Adds tests to account for these options

### DIFF
--- a/boilerplate/ignite/templates/component/NAME.tsx.ejs
+++ b/boilerplate/ignite/templates/component/NAME.tsx.ejs
@@ -1,4 +1,5 @@
 ---
+destinationDir: app/components
 patch:
   path: "app/components/index.ts"
   append: "export * from \"./<%= props.subdirectory %><%= props.pascalCaseName %>\"\n"
--- a/boilerplate/ignite/templates/component/NAME.tsx.ejs
+++ b/boilerplate/ignite/templates/component/NAME.tsx.ejs
@@ -1,5 +1,4 @@
 ---
-destinationDir: app/components
 patch:
   path: "app/components/index.ts"
   append: "export * from \"./<%= props.subdirectory %><%= props.pascalCaseName %>\"\n"
--- a/boilerplate/ignite/templates/model/NAME.test.ts.ejs
+++ b/boilerplate/ignite/templates/model/NAME.test.ts.ejs
@@ -1,3 +1,6 @@
+---
+destinationDir: app/models
+---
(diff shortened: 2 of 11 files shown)


mazenchami (approved):
🔥

PR #2729, opened 2024-08-06
Update deps for expo-doctor

lindboe:

## Describe your PR

Updates dependencies according to errors from `EXPOD_DOCTOR_ENABLE_DIRECTORY_CHECK=1 npx expo-doctor@latest`.

Note this command will still complain about Flashlist. Flashlist 1.7.0 is installed because we're pointing to `^1.6.4` and expo-doctor expects exactly `1.6.4`. Looking at the changes in 1.7.0, I think we want to keep this: https://github.com/Shopify/flash-list/releases/tag/v1.7.0.

That means this error will continue to exist. I [looked at silencing the dependency warning](https://docs.expo.dev/more/expo-cli/#configuring-dependency-validation), but right now we can't do it in a way that's dependent on the version, so it would be silenced for Flashlist forever until somebody remembers to check, which I think could lead to missed upgrades in the future, so I'd prefer to leave it as-is.


mazenchami (approved):

fpena (approved):
🚀

Issue #2638, opened 2024-02-15
[Enhancement] v10 - Remove Expo Go compatibility


frankcalise:
## Overview
- Remove Expo Go compatibility in favor of CNG workflow
- Make CNG the default (only?) experience (bare is just a gitignore change to allow ios/android dirs to be checked in)
- Add Cookbook recipe to revert back to Expo Go
- Target branch `v10`

frankcalise:
Closed by #2641

PR #2742, opened 2024-08-13
Monorepo setup doc

fpena:


docs/Setting-up-monorepo.md
```diff
@@ -0,0 +1,356 @@
...
+📝 Do you want to install dependencies?: No
+```
+
+3. Open `metro.config.js` file
+
+```
+touch mobile/metro.config.js
+```
+
+4. Replace Metro configuration with the following content
```
jamonholmgren, line 91: Instead of `replace`, show a diff instead. You can do a diff with three backticks & `diff`, like this:

```diff
diff --git a/old-metro.config.js b/new-metro.config.js
index abcdefg..hijklmn 100644
--- a/old-metro.config.js
+++ b/new-metro.config.js
@@ -1,17 +1,16 @@
-// Learn more https://docs.expo.io/guides/customizing-metro
 const { getDefaultConfig } = require('expo/metro-config');
+const path = require('path');

-/** @type {import('expo/metro-config').MetroConfig} */
-const config = getDefaultConfig(__dirname);
+// Find the project and workspace directories
+const projectRoot = __dirname;
+// This can be replaced with `find-yarn-workspace-root`
+const monorepoRoot = path.resolve(projectRoot, '../..');
+const config = getDefaultConfig(projectRoot);

-config.transformer.getTransformOptions = async () => ({
-  transform: {
-    // Inline requires are very useful for deferring loading of large dependencies/components.
-    // For example, we use it in app.tsx to conditionally load Reactotron.
-    // However, this comes with some gotchas.
-    // Read more here: https://reactnative.dev/docs/optimizing-javascript-loading
-    // And here: https://github.com/expo/expo/issues/27279#issuecomment-1971610698
-    inlineRequires: true,
-  },
-});
+// 1. Watch all files within the monorepo
+config.watchFolders = [monorepoRoot];
+// 2. Let Metro know where to resolve packages and in what order
+config.resolver.nodeModulesPaths = [
+  path.resolve(projectRoot, 'node_modules'),
+  path.resolve(monorepoRoot, 'node_modules'),
+];
 
-// This helps support certain popular third-party libraries
-// such as Firebase that use the extension cjs.
-config.resolver.sourceExts.push("cjs")
```

Speaking of which, this seems pretty drastic; are you sure you are using the latest Ignite here?

docs/Setting-up-monorepo.md
```diff
@@ -0,0 +1,356 @@
...
+cd apps/mobile
+yarn android
+```
+
+3. Run web app
+
+```
+cd apps/web
+yarn start
+```
```
jamonholmgren, line 352: We should probably address live reloading across web and mobile, when not to share logic / code, bundling considerations, and more. We should also make sure this guide, if appropriate, conforms to @trevor-coleman's ideas in [this internal wiki](https://infinitered.slab.com/posts/sharing-config-in-monorepos-u2av3f61) (open to disagreement with how he presented it though, use discernment please.)

jamonholmgren (changes requested):
Left comments, thanks Felipe!

docs/Setting-up-monorepo.md
```diff
@@ -0,0 +1,356 @@
...
+
+In a monorepo setup, multiple apps often share common functionality. For example, a mobile (React Native) and a web app (React). This tutorial demonstrates how to set up and use shared utilities within a monorepo. For instance, if you have several apps that need to validate user inputs like email addresses, passwords, or text fields, you can create a single validation utility that can be reused across all your apps.
+
+By centralizing these utilities, you reduce code duplication and simplify maintenance, ensuring that any updates or bug fixes are immediately available to all your apps.
+
+In this setup, we’ll create a React Native app, a React app and a form-validator utility that includes functions to validate email addresses, passwords, and generic text fields. This utility will then be integrated to both apps.
+
+## Step 1: Setting up the monorepo
+
+First, follow the [Expo documentation on setting up monorepos](https://docs.expo.dev/guides/monorepos/) to initialize your own monorepo. This will include setting up your `packages/` and `apps/` directories and configuring Yarn workspaces.
```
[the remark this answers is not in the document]
lindboe: @jamonholmgren that's a surprising question to me in this context. @frankcalise's comment on the CNG default issue is the only effective difference in using Ignite with "bare" is that the android and iOS directories won't be gitignored: https://github.com/infinitered/ignite/issues/2638.

So to me in effect that sounds like everyone using Ignite is using Expo (in the ways in which it matters for this context, which would would be using Expo libraries and Expo's metro config.)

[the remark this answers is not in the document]
jamonholmgren: @lindboe You're correct, technically, but we will still get this question and should address it here, IMO.

docs/Setting-up-monorepo.md
lindboe: Just a tip, we found on Cookbook that using `success-line` and `error-line` markup resulted in a better-looking diff with docusaurus:

https://github.com/infinitered/ignite-cookbook/blob/aca9fc9e6765ea8e5731ce375dc3231355558cfe/docs/recipes/EASUpdate.md?plain=1#L90-L95 for example

docs/Setting-up-monorepo.md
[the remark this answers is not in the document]
jamonholmgren: As an example (please don't just copy & paste this, I've given it very little thought 😅 ):

> **What if I'm not using Expo?** Generally speaking, it shouldn't make any significant difference for this guide how you're using Expo (or even if you're using any Expo libraries at all). You can still follow the Expo monorepo guide up to the point before creating the app itself.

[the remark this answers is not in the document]
jamonholmgren: BTW that Expo monorepo guide is very close to what I had in mind. It's very good, and we can draw inspiration from it (including all the little ways they educate and give hints along the way.)

CleanShot 2024-08-13 at 16 12 37@2x

docs/Setting-up-monorepo.md
jamonholmgren: That's very nice -- definitely cool!

CleanShot 2024-08-13 at 16 13 39@2x

docs/Setting-up-monorepo.md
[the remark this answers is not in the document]
lindboe: Okay thanks, that helps me understand what you're looking for 👍 

I think the main thing to call out would be that `expo/metro-config` has automatic monorepo support, as the guide notes:

> Expo's Metro config has monorepo support for bun, npm, and yarn.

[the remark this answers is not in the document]
jamonholmgren: I'm pretty sure we have non-Expo (in any way) projects that are in monorepos, and could provide guidance for those as well; this could be a todo for later though.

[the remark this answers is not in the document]
lindboe: > I'm pretty sure we have non-Expo (in any way) projects that are in monorepos, and could provide guidance for those as well; this could be a todo for later though.

@jamonholmgren I think we might need to clarify intended scope, then. My impression is we were wanting to document monorepo configuration for future Ignite projects, and that's why this going in the Ignite docs. If we want a more general-purpose guide, are the Ignite docs the right place for that?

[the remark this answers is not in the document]
jamonholmgren: Yeah, those could be a future Cookbook recipe @lindboe; we could link to them from here once they're published. It's always good to provide people with references to options, especially since those people might be us in the future!

docs/Monorepos-Overview.md
```diff
@@ -0,0 +1,97 @@
...
+| ----------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------- |
+| Yarn Workspaces [Recommended] | Built-in support for monorepos. <br>Simplifies dependency management across packages. <br>Excellent community support and documentation.  | Can be complex to configure initially. <br>Some issues with peer dependencies.                                                   |
+| Lerna                         | Great for managing multi-package repositories. <br>Supports independent versioning of packages. <br>Integrates well with Yarn Workspaces. | Slower for large codebases. <br>Requires additional configuration for CI/CD.                                                     |
```
trevor-coleman, line 34: Lerna is no longer actively supported. I would either not mention it or say that it's obsolete / not recommended.

docs/Monorepos-Overview.md
```diff
@@ -0,0 +1,97 @@
...
+| Turborepo                     | High performance with built-in caching and parallel task execution. <br>Simplifies monorepo management with minimal configuration.        | Less mature ecosystem compared to Yarn or Lerna. <br>Limited tooling support for some workflows.                                 |
+| Pnpm                          | Efficient disk space usage with unique package storage. <br>Fast installation times. <br>Built-in support for monorepos.                  | Less mainstream adoption, leading to potential issues with community support. <br>Some tooling might not fully support Pnpm yet. |
+
+## Common monorepo setups for software apps
+
+### Frontend and backend in a single monorepo
+
+Structure:
+
+- `/frontend`: Contains the React Native app and other frontend components.
```
trevor-coleman, line 45: I think between 
 "## Common monorepo setups for software apps" and "### Frontend and backend in a single monorepo"  it would be good to have a short paragraph on how to choose your structure:

> The way that you organize your monorepo is entirely up to you. Often, teams will fall back to a "default" organization that looks like this:
>
> - `apps/my-app`
> - `packages/package-one`
> - `packages/package-two`
>
> But just because it's common, doesn't mean that its right.  You should think carefully and pick a structure that makes sense for your project and your team. Here are some general rules:
>
> 1. Aim to keep your structure as simple as possible -- if you can use a single `/packages` folder that's sometimes the easiest and best. Each package is entirely independent, so if you have a small number of packages, then you may not need a complicated structure.
> 2. Consider CI and tooling -- jobs are often easiest to target by directory, so if a group of packages needs special tooling, or will be released together, keep them in the same folder. 
>
> Below are some examples of structures that might work for you and your team.

docs/Setting-up-monorepo.md
```diff
@@ -0,0 +1,357 @@
...
+
+In a monorepo setup, multiple apps often share common functionality. For example, a mobile (React Native) and a web app (React). This guide demonstrates how to set up and use shared utilities within a monorepo. For instance, if you have several apps that need to validate user inputs like email addresses, passwords, or text fields, you can create a single validation utility that can be reused across all your apps.
+
+By centralizing these utilities, you reduce code duplication and simplify maintenance, ensuring that any updates or bug fixes are immediately available to all your apps.
+
+In this setup, we’ll create a React Native app, a React app and a form-validator utility that includes functions to validate email addresses, passwords, and generic text fields. This utility will then be integrated to both apps.
+
+## Step 1: Setting up the monorepo
+
+First, follow the [Expo documentation on setting up monorepos](https://docs.expo.dev/guides/monorepos/) to initialize your own monorepo. This will include setting up your `packages/` and `apps/` directories and configuring Yarn workspaces.
```
trevor-coleman, line 22: In the other doc we spend a lot of time discussing alternate structures. Should we reference that here, maybe with a link?

docs/Setting-up-monorepo.md
```diff
@@ -0,0 +1,357 @@
...
+```
+
+2. Configure workspaces in `package.json`:
+
+```json
+{
+  "private": true,
+  "workspaces": {
+    "packages": [
+      "packages/*"
```
trevor-coleman, line 41: maybe include a note that you can/should customize this to match your monorepo structure if you don't use `packages` and `apps`

docs/Setting-up-monorepo.md
```diff
@@ -0,0 +1,357 @@
...
+  path.resolve(projectRoot, 'node_modules'),
+  path.resolve(monorepoRoot, 'node_modules'),
+];
+
+module.exports = config;
+```
+
+## Step 3: Create web app
+
+Let's create a second app now. This app is meant to be run on web browsers.
```
trevor-coleman, line 117: It might be more useful to have something like a design system package in `/packages` that exports some components as the example, as that's far more common in my experience.

docs/Setting-up-monorepo.md
```diff
@@ -0,0 +1,357 @@
...
+Let's make sure all of our dependendencies are installed for both apps.
+
+1. Run `yarn` at the root of the project
+
+```shell
+cd ..
+cd ..
+yarn
+```
+
```
trevor-coleman, line 183: It might be good to talk a bit about what this did -- i.e. point out where the `node_modules` folders are, with a link to something about hoisting.

Also, you may want a note for users of yarn 2+ about checking their settings in `yarnrc.yml` in case things are hoisted that shouldn't be. Not like full instructions just a pointer to the issue.

docs/Setting-up-monorepo.md
trevor-coleman: Adding a shared eslint config package or a shared tsconfig package would be a great addition to this guide, as that's the place where most people screw up.

trevor-coleman:
One other thing, which I'm not sure where to put, but I think we should include is [yarn global scripts](https://yarnpkg.com/features/workspaces#global-scripts)! ([yarn classic link](https://classic.yarnpkg.com/lang/en/docs/workspaces#global-scripts))

in monorepos you always see the root package.json with a bunch of scripts like:

```
build:my-app
build:package-one
lint:my-app
lint:package-one
```

but these are redundant!

From any package in the repo, a script with a prefix will be **globally** available.

So if you have:

```json
// packages/package-one/package.json

{
   //...
   scripts: {
       "build:package-one" : "yarn tsc --watch"
       //...
  }
}
```

then you can do `yarn build:package-one` from _anywhere_ and yarn will execute it in the package-one scope! 

Scripts like this in the root just add indirection and confusion about what's happening:
```
"build:package-one" : "yarn workspace package-one build" 
```

Maybe this is more of an advanced tips-and-tricks thing. But I think not enough people know about it.

docs/Setting-up-monorepo.md
fpena: Added one

CleanShot 2024-08-15 at 08 16 37@2x

docs/Setting-up-monorepo.md
fpena: Hey @trevor-coleman, thanks! 

Yes, did add a section for your first point. 

For the second one, this guide is intended for Yarn 3.x. Do you think it would still be valid?

docs/Setting-up-monorepo.md
fpena: Added a section for this. 

CleanShot 2024-08-18 at 07 27 52@2x

docs/Setting-up-monorepo.md
trevor-coleman: Ah no if it's yarn 3 then you wouldn't need it.

docs/Setting-up-monorepo.md
trevor-coleman: nice

docs/Setting-up-monorepo.md
trevor-coleman: nice

fpena:
Closing this one in favour of moving this doc to the [Ignite cookbook](https://github.com/infinitered/ignite-cookbook).

PR #2744, opened 2024-08-14
Fix screen jump from Welcome screen to Demo screen

lindboe:

## Describe your PR

To fix Android build issue, depends on https://github.com/infinitered/ignite/pull/2745.

Android and iOS were both jumping when navigating from the Welcome screen to the Demo Showroom screen (although in different directions, and Android was less severe).

I found this issue was prevented when we use the `StatusBar` component on the WelcomeScreen as well as DemoShowroom, so converted WelcomeScreen to use the built-in `Screen` component, which includes `StatusBar`.

However, that raised another issue: when navigating from the Login screen to the Welcome screen, the "fixed" `Screen` component on Welcome screen would render the wrong height; it would be slightly too tall.

The `useHeader` hook was using useEffect, which means that header height wouldn't be set until the second render, so when the KeyboardAvoidingView on on Welcome screen tried to render, it had the wrong height information and rendered the wrong size.

By using `useLayoutEffect`, we make sure the header height is set before any of the screen components try to render, fixing the issue.

| Platform  | Before | After |
| ------------- | ------------- | ------------- |
| Android  | <video src="https://github.com/user-attachments/assets/e6aa4385-08fd-425d-b838-e4012b00556d" />  | <video src="https://github.com/user-attachments/assets/48e2e5fb-287f-46e9-88bc-01371bb68827" />  |
| iOS  | <video src="https://github.com/user-attachments/assets/fe6cd6a8-b0cc-47f6-9f04-c8d797a1d24f" />   | <video src="https://github.com/user-attachments/assets/fa76aa66-62b5-427e-8b7b-fcbb96b0ca84" />  |

### DIFF
--- a/boilerplate/app/screens/WelcomeScreen.tsx
+++ b/boilerplate/app/screens/WelcomeScreen.tsx
@@ -4,6 +4,7 @@ import { Image, ImageStyle, TextStyle, View, ViewStyle } from "react-native"
 import {
   Button, // @demo remove-current-line
   Text,
+  Screen,
 } from "app/components"
 import { isRTL } from "../i18n"
 import { useStores } from "../models" // @demo remove-current-line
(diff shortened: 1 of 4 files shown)


frankcalise (approved):
Great investigation! So much smoother

Issue #2737, opened 2024-08-09
Audit and fix linting/transpiling dependencies


lindboe:
Lint dependencies should be audited and updated, at least for v10.

A few known areas in need of attention:
1. These babel plugins are probably no longer needed, this behavior should be included: https://github.com/infinitered/ignite/blob/master/boilerplate/package.json#L67-L69
2. The project linting still requires us to import React, which shouldn't be necessary anymore with React 17. See https://github.com/infinitered/ignite/blob/master/boilerplate/package.json#L67-L69 for more information, we want "plugin:react/jsx-runtime"
3. Consider reconstructing the lint configuration on top of Expo's eslint configuration ([setup docs](https://docs.expo.dev/guides/using-eslint/), [preset code](https://github.com/expo/expo/tree/main/packages/eslint-config-expo)). This might help simplify our own configuration a good deal.

frankcalise:
re 1: you're probably right, we just got rid of a few but would be nice to clean more of those up: https://github.com/infinitered/ignite/pull/2701/files#diff-594f75b3ec0c76210a39b8c7ce238eea438454a93d830bd968a282f89e30a8b0

PR #2761, opened 2024-08-29
Update lint-related dependencies and configuration

lindboe:

## Describe your PR

Addresses #2737.

This PR modernizers our linting setup by updating Babel, ESLint, and Prettier, and our rule configuration.

To keep things manageable, I tried to keep most existing opinions in place, but had to make several opinionated decisions. These decisions are outlined below, and can be reviewed commit-by-commit. Every decision here is up for discussion.

## Decisions

1. Update out-of-date packages
    1. Importance: high, we should keep deps up-to-date
    2. Remove unnecessary babel transformations: these are handled by the existing babel preset
    3. Move to eslint 8, update other plugins, and fixes warning from typescript-eslint about unsupported TS version
    4. Updated root CLI project dependencies to match boilerplate where needed to get consistent formatting between the root project and a generated Ignite project
        1. didn’t update ts-jest in CLI — that’s had an out-of-date TS peer dep for a while, but will require updating jest from 27 → 29, which might be big
2. Use expo’s eslint config: https://docs.expo.dev/guides/using-eslint/
    1. Importance: Medium. There’s value here, but how to get it is debatable.
    2. What
        1. Keeps a basic set of lint dependencies up-to-date for us:
        2. Eslint runs prettier (optional, see later decision on this)
        3. adds react-hooks linter (see later decision on this)
        4. handles globals for us
    3. Why
        1. Keeps us in line with other expo projects, easier to stay up-to-date with globals or other RN-specific config the linter needs
    4. Arguments against
        1. Doesn’t configure many rules, but does configure some, including some we don’t prefer, and that might get confusing for maintenance
    5. Downstream decisions
        1. ESlint/prettier config files, not package.json
            1. Why:
                1. Expo default config doesn’t do package.json config
                2. ESlint 9 will not support package.json config
        2. ESlint runs prettier as a plugin
            1. Why: Because that’s what Expo recommends by default, and I don’t see a reason not to.
        3. Use react-hooks linter
            1. Why:
                1. It’s included with expo-config-eslint
                2. it includes the `rules-of-hooks` and `exhaustive-deps` lint rules, which can help avoid bugs, and most projects use them
        4. More files included by default
            1. Importance: low.
            2. Why: We were excluding a lot of files from formatting rules. This is reasonable for files we don’t want to manage and pull directly from a template. It’s not clear to me which that really applies to, so I left the config open to more of these files for now. (https://github.com/infinitered/ignite/commit/61747804f330f0054e96ba1fcc52c065f7abd663 for examples)
            3. Downstream decision: configure prettier use use `quote-props: 'consistent'`  ([[docs](https://prettier.io/docs/en/options.html)](https://prettier.io/docs/en/options.html)), since mixed string/non-string keys in config files looked odd
3. Don’t use react-native linter
    1. Importance: low
    2. Why:
        1. we’re ignoring no-inline-styles constantly
        2. development on this plugin has slowed (see [[project update](https://github.com/intellicode/eslint-plugin-react-native)](https://github.com/intellicode/eslint-plugin-react-native) in README), and it doesn’t seem like we would get much out of [[the available rules](https://github.com/intellicode/eslint-plugin-react-native?tab=readme-ov-file#configuration)](https://github.com/intellicode/eslint-plugin-react-native?tab=readme-ov-file#configuration). Open to disagreement on that though.
4. Configure JSX runtime lint rules, don’t need to import React when using JSX (do need to still import when using React API)
    1. Importance: Low, but I don’t see why we shouldn’t.
    2. Why: modern React transpiling with Babel hasn’t needed `React` to be imported when writing JSX for a while now. The linter config should match that too.
    3. Downstream decisions
        1. Require importing from `React` consistently
            1. Why: I noticed we’re very inconsistent about whether we use `React.thing` versus importing named `thing`. [[The blog post about the new JSX transform encourages use of named imports](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html)](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html), so I decided to just default to that for this.
            2. Importance: low, but I think it’s nice to be consistent
5. remove-demo/remove-mst-markup are introducing formatter churn because they’re not using the same formatting as our linter setup
    1. there is a formatting step in CLI already, that should cover installDeps use case; move it to after the remove-* steps, and add a branch to run prettier if we haven’t used installDeps, which should be enough for most cases

## Decisions left for later

1. Updating tsconfig and/or `@typescript-eslint` configuration
    1. should audit rules and streamline TS config
        1. there’s a `test-tsconfig.json` that extends the base boilerplate config but re-adds a lot of the same rules
        2. some potentially useful configuration to consider adding:
            1. `noUncheckedIndexAccess` ([[link](https://www.typescriptlang.org/tsconfig/#noUncheckedIndexedAccess)](https://www.typescriptlang.org/tsconfig/#noUncheckedIndexedAccess)): when you have an object type with arbitrary keys, TS can often wrongly assume that the value you looked up is defined. This sets it so that there's always a default `undefined` you have to check for
            2. `exactOptionalPropertyTypes` ([[link](https://www.typescriptlang.org/tsconfig/#exactOptionalPropertyTypes)](https://www.typescriptlang.org/tsconfig/#exactOptionalPropertyTypes)): will prevent you from setting an optional prop to `undefined` if that's not part of the type, since that can mess up the `in` operator.
            3. `useUnknownInCatchVariables` ([[link](https://www.typescriptlang.org/tsconfig/#useUnknownInCatchVariables)](https://www.typescriptlang.org/tsconfig/#useUnknownInCatchVariables)): without this, catch variables will be `any`, and using `unknown` is pretty much always preferable for type-checking IMO
            4. `noFallthroughCasesInSwitch` ([[link](https://www.typescriptlang.org/tsconfig/#noFallthroughCasesInSwitch)](https://www.typescriptlang.org/tsconfig/#noFallthroughCasesInSwitch)): was kind of surprised to learn this isn't the default.
        3. should audit typescript-eslint config rules; do we want rules that include typechecking?
2. VS Code configuration in-project
    1. It’s important to configure your editor to run linting and formatting correctly, otherwise you can get conflicting behavior. However, putting this configuration into the project isn’t simple: 1) there’s no way to override those committed project-specific settings locally without committing them for everyone, 2) because of the way VS Code merges settings, you may have to repeat configuration several times under different language settings, and 3) we’re currently ignoring `.vscode` and that decision may need further discussion. I recommend putting this off for a subsequent PR.
3. `Perfectionist` sorting or other new opinionated rules
    1. Went with consistent `React` import as a test drive for a new rule. don’t want to introduce too many right now, just enough to test that lint config is working correctly. (It is hard to test new lint config against already linted code, can’t tell if it’s missing rules or just that issues are fixed!)
4. Consistent linting between root CLI project and generated project: to achieve this, we either need a monorepo-like setup to share lint deps b/n CLI and boilerplate, or to have boilerplate linted by its own deps in-place (may not be feasible because of comment markup that gets removed)


robinheinze (commented):
This is an incredible update! Much needed ❤️ 

My only hesitation is the react-native linter `no-inline-styles` removal. My understanding is that we were overriding it in many of the demo files simply because for the purposes of demonstrating the versatility of the components, it's more helpful to be able to see the exact styles as you look at the code. As far as I know, we still try to avoid inline styles in general in our projects. Definitely open to hearing from others about how important that is to them. 

Otherwise, I'm a-ok with all of this 👍

lindboe:
@robinheinze good call out, I can add `react-native` config back in then

boilerplate/app/components/Card.tsx
```diff
@@ -271,8 +271,8 @@ const $alignmentWrapper: ViewStyle = {
 }
 
 const $alignmentWrapperFlexOptions = {
-  top: "flex-start",
-  center: "center",
+  "top": "flex-start",
```
frankcalise, line 274: Do these quotes match the other properties because they have hyphens in them? that's pretty cool

boilerplate/package.json
frankcalise, line 1: So much cleaner 💯

src/commands/new.ts
```diff
@@ -799,24 +799,21 @@ module.exports = {
       stopSpinner(" Configuring app.json", "⚙️")
       // #endregion
 
-      // #region Run Format
+      // #region Run Prebuild
       // we can't run this option if we didn't install deps
       if (installDeps === true) {
         // Check if we need to run prebuild to generate native dirs based on workflow
         // Prebuild also handles the packager install
-        const prebuildMessage = ` Generating native template via Expo Prebuild`
```
frankcalise, line 807: Hm I think some of these spaces were intended to make the terminal output look nicer, let's see if things don't looked mashed up against the spinner / icon bullet

frankcalise (approved):
Awesome job! Thanks for doing this.

Had one question on how things turned out in the new command

boilerplate/app/components/Card.tsx
lindboe: Yep: that's what `"quoteProps": "consistent"` in the prettier config is for

src/commands/new.ts
lindboe: Oh yeah I removed that because it looked odd to me: I'll try to dig up screenshots to compare with

lindboe: What I observed that led to me editing out the space:

Screenshot 2024-08-27 at 4 07 12 PM

What it looks like now on this branch (didn't realize the MST icon had changed):
Screenshot 2024-09-05 at 2 41 16 PM

I notice I missed space for "Configuring app.json", so that's still inconsistent.

What was the intended design here?

frankcalise: Ok cool, no real intended design just have them consistent, it looks fine without the space. It must have just been left over from elsewhere at some point

lindboe: 👍 I'll remove the one last remaining space then

lindboe:
Updated 👍

PR #2770, opened 2024-09-10
[2754] Swap out i18n-js for react-18next

fpena:

## Describe your PR

This PR replaces the `i18n-js` library with `i18next` to enhance internationalization in the project. The migration ensures better support for translations, including improved handling of locale-specific formats, pluralization, and more efficient loading of translation resources.

### Key Changes:

- Replaced all instances of i18n-js with i18next across the codebase.
- Updated translation initialization logic to use initReactI18next from the i18next library.
- Refactored translation key format: e.g., from "common.ok" to "common:ok".
- Updated all relevant unit tests reflect changes in the translation library.

Note: We will also create a cookbook recipe later that references this PR for Ignite users who want to migrate from `i18n-js` to `i18next` and are using  Ignite 9.x or less. 

## Video

https://github.com/user-attachments/assets/641f33f4-dec0-4fc2-98b3-0aaf57f5101b


boilerplate/package.json
```diff
@@ -97,6 +98,9 @@
     "ts-node": "^10.9.2",
     "typescript": "~5.3.3"
   },
+  "resolutions": {
+    "eslint-plugin-react": "7.35.2"
```
frankcalise, line 102: What was the reason for adding this?

If we're going to add a yarn resolution we also need to add the proper keys for other package managers, such as [overrides for npm](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#overrides)

frankcalise (commented):
Looking good!

What was the motivation behind the `screen:some.key` naming scheme change?

Not opposed to it, just made a bunch of extra changes because of it.

boilerplate/package.json
lindboe: This I ended up adding in the expo-router PR. But this is the commit: https://github.com/infinitered/ignite/commit/b1cb6c3dd47a236baf2aa2ec375ce7bf19746a3a

Today there were new test failures because they published a new version of this package that has a bug in a rule, but the fix wasn't published yet: https://github.com/jsx-eslint/eslint-plugin-react/pull/3821.

I was trying to pin it this way since it's not a direct dependency, but didn't realize that didn't work for all package managers. My bad.

@fpena we'll want to remove this specific change from you PR; you added this in a different spot in package.json and now if you look at the file there's two separate `resolutions` fields.

lindboe: Actually it looks like we can remove both `resolutions` sections, they published a new version that fixes the issue: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/CHANGELOG.md#7361---20240912

fpena: @lindboe @frankcalise  Yes, removed them from this PR.

fpena:
> Looking good!
> 
> What was the motivation behind the `screen:some.key` naming scheme change?
> 
> Not opposed to it, just made a bunch of extra changes because of it.

@frankcalise it is the namespace separator for `i18next`. There's more context about it [here](https://infinitered.slack.com/archives/C0X0P5LP7/p1725982358375359).

frankcalise (approved):
Great work!

boilerplate/package.json
frankcalise: Cool thanks!

PR #2772, opened 2024-09-11
Fixes for expo-router app generation and issues with master merge up

lindboe:

## Describe your PR

We found some issues with paths when generating a new app with expo-router enabled. To make this easier to manage, I added `@` as an alias to consistently refer to the root app directory, whether it's `app` or `src` (depending on react-navigation vs. expo-router).

This still requires us to transform some files, but much fewer. I also left test files out of the path alias configuration for now, because we have a separate test-tsconfig file and we don't immediately need to fix that, but I'd love to remove that separate configuration in the future.
 
I also found:
1. Merged up i18n work did not remove all lines needed for demo removal, and generated app had compile errors
2. Tests on this branch and another were failing inconsistently; to address, I updated the tests to log all of stdout when tests fail so we are able to directly see the error message of what went wrong. This could probably be prettier, but it's an improvement for now.
3. Generative tests that were using bun were failing because the react-native-drawer-layout patch wasn't getting removed, and this I'm pretty confused by. It did originally work and fixed tests that were failing prior, so I'm not sure what went wrong. Fixed in this PR here: https://github.com/infinitered/ignite/pull/2772/commits/03bbc7ad309687f0ce46b9015707e8a4cc46d59e.

I'm concluding that it's _extra_ important to make PRs for merging up changes from `master` at this point because v10 now has a lot more tests for making sure the generated app has no errors.

### DIFF
--- a/src/tools/react-native.ts
+++ b/src/tools/react-native.ts
@@ -311,6 +311,7 @@ export function updateExpoRouterSrcDir(toolbox: GluegunToolbox) {
     "tsconfig.json",
     // has its own tsconfig, needs updating separately
     "test/i18n.test.ts",
+    "test/setup.ts",
     "ignite/templates/model/NAME.ts.ejs",
     "ignite/templates/component/NAME.tsx.ejs",
   ]
--- a/src/tools/react-native.ts
+++ b/src/tools/react-native.ts
@@ -309,10 +309,8 @@ export function updateExpoRouterSrcDir(toolbox: GluegunToolbox) {
   const TARGET_DIR = filesystem.path(process.cwd())
   const expoRouterFilesToFix = [
     "tsconfig.json",
+    // has its own tsconfig, needs updating separately
     "test/i18n.test.ts",
-    "src/devtools/ReactotronConfig.ts",
-    "src/components/Toggle/Switch.tsx",
-    "src/components/ListView.tsx",
     "ignite/templates/model/NAME.ts.ejs",
     "ignite/templates/component/NAME.tsx.ejs",
   ]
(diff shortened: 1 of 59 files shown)


lindboe:
@frankcalise I'm getting this type error in `App.tsx` with the changes you made to the `app` component.

I'm not sure what the desired interface here looks like to reconcile this and still work for expo-router. Do you know how you'd like to fix this?

Screenshot 2024-09-11 at 10 59 34 AM

lindboe:
@fpena I think I've fixed every reproducible error I can find, check out this branch and let me know how it goes for you

src/tools/react-native.ts
```diff
@@ -309,10 +309,9 @@ export function updateExpoRouterSrcDir(toolbox: GluegunToolbox) {
   const TARGET_DIR = filesystem.path(process.cwd())
   const expoRouterFilesToFix = [
     "tsconfig.json",
+    // has its own tsconfig, needs updating separately
     "test/i18n.test.ts",
-    "src/devtools/ReactotronConfig.ts",
-    "src/components/Toggle/Switch.tsx",
-    "src/components/ListView.tsx",
+    "test/setup.ts",
```
fpena, line 314: Curious: Why are we removing the 3 lines on the left?

fpena (approved):
Looks good! Just left 1 non-blocking question.

src/tools/react-native.ts
frankcalise: Because the alias was updated so there are less specific changes to files that are necessary. Before the changes were very specific and didn't scale well, now @lindboe has adjusted the path alias in `tsconfig` and we don't have to do so many transformations across the source files (whether they're in `app/` or `src/`) - less maintenance for us.

frankcalise (approved):
Much cleaner and looks to be more future proof, well done!

src/tools/react-native.ts
fpena: Thanks!

lindboe: In addition, the test files are still in there because there's a separate TS config, but if we can unify those better we can also update the test files automatically. But I'm planning on circling back to that in a later change.

PR #2784, opened 2024-09-24
Stream test command output to file for debugging purposes

lindboe:

### What

We needed `ignite new` output to be printed by failing tests when developing; otherwise you have to generate the app at least twice to see why a given test is failing.

### Approach

The way to get jest to print the output we want is by throwing it when the command fails. However, when we tried this, we found it was possible for the output to be so large for certain errors that it can hit node’s max buffer size, which *also* means you don’t get to view the error (and get a red herring error on top of that.)

That means we should stream the output with `spawn`, to manage output that's too large. In this PR, we stream it to a file so there’s always at least a partial record of the output, even if we fail to be able to print it to the console.

Existing tests expect a string of command output to test against, and so the simplest implementation for now was to just read the output from the file again. There’s potential for changing  this in the future but it should work just fine for now, since the happy path doesn’t involve output too large for a buffer.

### Testing

Throw an error somewhere in `new.ts`. it should be printed to the console for tests that run `ignite new`.

### Future improvements to consider

- Use with more commands? right now I focused on just `ignite new` since that's where we tend to have issues
    - it is getting a little silly to make `verbIgnite` versions of all these commands in order to account for appending Ignite in there. Might want to handle that differently. (As an options flag? Let the caller handle it?)
- Log cleanup?
    - I don’t think it’s necessary right now; they’re gitignored and re-written each test run. Shouldn’t be too much stuff.
- testing against streamed data instead of loading it all in as one file
    - complicated! Chunks of stream are probably not the entire string we want to test against, and what a given chunk looks like can change as things are added before it. Have to build a test utility library for this, practically.
- Avoid logging command run on normal test runs
    - Was hoping to use jest's `--verbose` flag to control this, but it doesn't appear to work like that. I think it’s more useful to log the command run than not, and it doesn’t take up too much space.
- automated log naming when using the spawn fns? tried look at auto-naming logs based on tests, jest doesn’t appear to have an affordance for that

### DIFF
--- a/test/vanilla/ignite-new.test.ts
+++ b/test/vanilla/ignite-new.test.ts
@@ -1,6 +1,7 @@
 import { filesystem } from "gluegun"
 import * as tempy from "tempy"
-import { runIgnite, runError, run } from "../_test-helpers"
+import { spawnIgnite, runError, run, runIgnite } from "../_test-helpers"
+import { stripANSI } from "../../src/tools/strip-ansi"
 
 const APP_NAME = "Foo"
 const originalDir = process.cwd()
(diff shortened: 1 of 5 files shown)


test/vanilla/ignite-new.test.ts
```diff
@@ -287,54 +281,10 @@ describe("ignite new", () => {
     beforeAll(async () => {
       tempDir = tempy.directory({ prefix: "ignite-" })
 
-      result = await runIgnite(`new ${APP_NAME} --debug --packager=yarn --workflow=cng --yes`, {
```
lindboe, line 290: This diff is a little confusing, sorry; I noticed while updating this that this test for yarn was duplicated.

fpena:
@lindboe this looks good, thanks for creating it. Let me know when you would like a formal review (I see it's still a draft)

lindboe:
The main reason I was leaving this in draft is I wasn't sure whether to consider it complete if it only covers `ignite new`. @frankcalise do you have thoughts on this? Would command output for other commands be useful?

frankcalise (commented):
I like this a lot. Since the output can get verbose, it's nice to be able to analyze the file entirely and get that feedback, otherwise even though we have these tests we'd end up doing more manual checks anyway!

As far as other commands to check, I'd say generate is the only other one that would get  any volume (but no where near the complexity of new) - but I think the snapshot test there should be fine.

lindboe:
@frankcalise Agreed, I think snapshot tests are probably better for generate.

@frankcalise @fpena any thoughts on approve/request changes?

frankcalise (approved):

frankcalise:
@lindboe good to go!

PR #2785, opened 2024-09-24
fix(cli): improper MST removal when using CLI prompt, new arch compat

frankcalise:

## Describe your PR
- As described by @markrickert, when using the prompts to set the following scenario (yarn, no demo, git repo, install deps, no MST, expo router, new arch) - MST was not removed properly
- With proper MST removal, the `lint` step failed inside of the ReactotronConfig leaving in `clear` from `@/utils/storage` which we were no longer needing, so I added the proper MST removal directives around that)
- Also occurs in the main `_layout` file when in the `expo-router` scenario
- When opting into the new architecture, some libraries need to be updated (for example `react-native-mmkv` from 2 to 3), the `new` command has updates for this

### DIFF
--- a/boilerplate/src/app/_layout.tsx
+++ b/boilerplate/src/app/_layout.tsx
@@ -27,7 +27,8 @@ export default function Root() {
 
   const [fontsLoaded, fontError] = useFonts(customFontsToLoad)
 
-  const loaded = fontsLoaded && rehydrated
+  const loaded = fontsLoaded 
+                         && rehydrated // @mst remove-current-line
 
   useEffect(() => {
     if (fontError) throw fontError
--- a/src/commands/new.ts
+++ b/src/commands/new.ts
@@ -31,7 +31,11 @@ import type { ValidationsExports } from "../tools/validations"
 import { boolFlag } from "../tools/flag"
 import { cache } from "../tools/cache"
 import { mstDependenciesToRemove } from "../tools/mst"
-import { findAndRemoveDependencies } from "../tools/dependencies"
+import {
+  findAndRemoveDependencies,
+  findAndUpdateDependencyVersions,
+  newArchCompatExpectedVersions,
+} from "../tools/dependencies"
 import { demoDependenciesToRemove, findDemoPatches } from "../tools/demo"
 
 type Workflow = "cng" | "manual"
@@ -656,6 +660,14 @@ module.exports = {
       if (stateMgmt === "none") {
         log(`Removing MST dependencies... ${mstDependenciesToRemove.join(", ")}`)
         packageJsonRaw = findAndRemoveDependencies(packageJsonRaw, mstDependenciesToRemove)
+      }
+
+      if (experimentalNewArch) {
+        log(`Swapping new architecture compatible dependencies...`)
+        packageJsonRaw = findAndUpdateDependencyVersions(
+          packageJsonRaw,
+          newArchCompatExpectedVersions,
+        )
       }
 
       // Then write it back out.
--- a/src/commands/new.ts
+++ b/src/commands/new.ts
@@ -366,7 +366,7 @@ module.exports = {
           format: prettyPrompt.format.boolean,
           prefix,
         }))
-        stateMgmt = includeMSTResponse.includeMST
+        stateMgmt = includeMSTResponse.includeMST ? "mst" : "none"
       }
     }
 
--- a/src/tools/dependencies.ts
+++ b/src/tools/dependencies.ts
@@ -24,3 +24,25 @@ export function removePackageJSONDependencies(
   const updatedPackageJson = findAndRemoveDependencies(packageJsonRaw, dependenciesToRemove)
   filesystem.write(packageJsonPath, updatedPackageJson)
 }
+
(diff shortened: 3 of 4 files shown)


lindboe:
Should we add more tests to prevent this from breaking again? Like fully generating and running lint on these permutations.

I know it adds more time, but it also seems useful since it's easy to accidentally break. (Side note: I realized today we should probably break out the tests that generate projects into different files, since that allows Jest to run them in parallel.)

boilerplate/src/app/_layout.tsx
```diff
@@ -27,7 +27,8 @@ export default function Root() {
 
   const [fontsLoaded, fontError] = useFonts(customFontsToLoad)
 
-  const loaded = fontsLoaded && rehydrated
+  const loaded = fontsLoaded 
+                         && rehydrated // @mst remove-current-line
```
lindboe, line 31: I did a double-take at this and wondered if it would pass lint check if we don't remove MST, and then I remembered we run a lint fix after that processing. All good 👍

src/commands/new.ts
```diff
@@ -366,7 +370,7 @@ module.exports = {
           format: prettyPrompt.format.boolean,
           prefix,
         }))
-        stateMgmt = includeMSTResponse.includeMST
+        stateMgmt = includeMSTResponse.includeMST ? "mst" : "none"
```
lindboe, line 373: Was this coming being set as the wrong type of data? According to TS this should always be `"mst" | "none"` already. if that wasn't the case do we have a type error in prompt.ask?

src/tools/dependencies.ts
```diff
@@ -24,3 +24,25 @@ export function removePackageJSONDependencies(
...
+export function findAndUpdateDependencyVersions(
+  packageJsonRaw: string,
+  dependencies: Record<string, string>,
+): string {
+  let updatedPackageJson = packageJsonRaw
+
+  Object.keys(dependencies).forEach((depName) => {
+    const desiredVersion = dependencies[depName]
+    const regex = new RegExp(`"${depName}"\\s*:\\s*"[^"]+"`, "g")
+    updatedPackageJson = updatedPackageJson.replace(regex, `"${depName}": "${desiredVersion}"`)
```
lindboe, line 37: Not important to fix here since we're already doing  this, but this did make me wonder, why are we often operating on the package.json as a string instead of structured data? It seems easier to operate on it as an object for these kinds of changes, to me?

lindboe (approved):
Looks good to merge, just left some comments

boilerplate/src/app/_layout.tsx
frankcalise: Yeah! Looks awkward but I didn't want to have to make some eslint type directive of
 `const loaded = fontsLoaded /* @mst remove-start */ &&  rehydrated /* @mst remove-end */`
 
 Although we do have that already right? can you swap `//` for `/*`? I'll have to check that now

src/commands/new.ts
frankcalise: Yeah probably, it was being set to a boolean value

src/tools/dependencies.ts
frankcalise: I can definitely look into it, that's a good call out, we did that for the `app.json`.

boilerplate/src/app/_layout.tsx
frankcalise: I tried using `  const loaded = fontsLoaded /* @mst remove-block-start */ && rehydrated /* @mst remove-block-end */`

but the entire line gets removed

src/tools/dependencies.ts
frankcalise: Ok so I started to work on this, adding

```
// Record<string, string> could be more complex for dependencies
// For example, we could type them with versions, but also
// GH forks or other custom URLs could be provided there technically, so it gets complex
export interface PackageJson {
  name: string
  version: string
  private: boolean
  main: string
  description: string
  scripts: Record<string, string>
  dependencies: Record<string, string>
  devDependencies: Record<string, string>
  engines: Record<string, string>
}
```

Thoughts on continuing with this? Should I make it a separate PR, it'll impact everything in this range: https://github.com/infinitered/ignite/blob/bc6caf9e4b167a21ae97d0b68aa4dc3f52d5ff7b/src/commands/new.ts#L619-L675

src/tools/dependencies.ts
```diff
@@ -24,3 +24,25 @@ export function removePackageJSONDependencies(
...
+
+  // Make sure `expo-dev-client` is removed
+  updatedPackageJson = updatedPackageJson.replace(/"expo-dev-client"\s*:\s*"[^"]+",?/g, "")
+
+  return updatedPackageJson
+}
+
+export const newArchCompatExpectedVersions = {
+  "react-native-mmkv": "3.0.1",
+}
```
jamonholmgren, line 48: I kinda like this idea in general.

src/tools/dependencies.ts
lindboe: Yeah a separate PR seems reasonable as a refactor

PR #2788, opened 2024-09-26
Fix language switching and update date-fns to v4

lindboe:

## Describe your PR

While updating date-fns to v4, I ran into a few issues with language switching.

### Languages were not being loaded correctly

The previous way we had of loading languages with i18n-js doesn't work with i18next. `hasOwnProperty` does not work with `i18n.languages`, as it's an array. `i18n.languages` is also empty on init, as it's only an array of the currently configured languages to use (preferred, then fallback), not all supported languages.

Instead I created a function to pick the first locale from device locales that we support, and configure i18n with that.

### Fast refresh would crash the app

This we fix by referring to i18n from the library, we don't need to store it in a variable in `i18n.ts`.

### Updates for date-fns v4

Here I just needed to update some import paths, but while I was there, I also added support for more of our supported locales, and conditionally load only the locale actually in use.

### Future improvements

I recommend we decide whether and how we want to follow up on these items later:

1. A better `Episode` test. It's not clear what this test wants to assert. If it wants to assert that the models are accounting for params correctly, we may have to find a way to manually test that, probably spying on `t()`. If it wants to assert that we're translating accessibility labels correctly, we should make it an E2E test. Etc.
2. Live reloading of locale while app is open on Android. Theoretically possible, but wasn't straightforward when I tried, particularly because we're using the `translate` function everywhere which doesn't know to react when language is changed (unlike if we were using the `useTranslation` hook.)
3. More foolproof toggling of RTL settings (i.e., always verifying if RTL setting matches what we expect, and if not, restart the app).
4. The drawer isn't rendering correctly when RTL is set, but it was also not rendering correctly with the i18n-js config. We need to circle back to this.

| Platform | This PR | With i18n-js |
|--------|--------|--------|
| iOS Device|  | don't have |
| iOS Sim | don't have |  |
| Android device |  |  |

### DIFF
--- a/boilerplate/app/i18n/i18n.ts
+++ b/boilerplate/app/i18n/i18n.ts
@@ -1,6 +1,6 @@
 import * as Localization from "expo-localization"
 import { I18nManager } from "react-native"
-import * as i18next from "i18next"
+import i18n from "i18next"
 import { initReactI18next } from "react-i18next"
 import "intl-pluralrules"
 
@@ -14,8 +14,6 @@ import ja from "./ja"
 import hi from "./hi"
 
 const fallbackLocale = "en-US"
-
-export let i18n: i18next.i18n
 
 const systemLocales = Localization.getLocales()
 
@@ -46,7 +44,7 @@ if (locale?.languageTag && locale?.textDirection === "rtl") {
 }
 
 export const initI18n = async () => {
-  i18n = i18next.use(initReactI18next)
+  i18n.use(initReactI18next)
 
   await i18n.init({
     resources,
--- a/boilerplate/app/i18n/i18n.ts
+++ b/boilerplate/app/i18n/i18n.ts
@@ -13,55 +13,52 @@ import fr from "./fr"
 import ja from "./ja"
 import hi from "./hi"
 
-// to use regional locales use { "en-US": enUS } etc
 const fallbackLocale = "en-US"
 
 export let i18n: i18next.i18n
 
-const systemLocale = Localization.getLocales()[0]
-const systemLocaleTag = systemLocale?.languageTag ?? fallbackLocale
+const systemLocales = Localization.getLocales()
+
+const resources = { ar, en, ko, es, fr, ja, hi }
+const supportedTags = Object.keys(resources)
+
+// Checks to see if the device locale matches any of the supported locales
+// Device locale may be more specific and still match (e.g., en-US matches en)
+const systemTagMatchesSupportedTags = (deviceTag: string) => {
+  const primaryTag = deviceTag.split("-")[0]
+  return supportedTags.includes(primaryTag)
+}
+
+const pickSupportedLocale: () => Localization.Locale | undefined = () => {
+  return systemLocales.find((locale) => systemTagMatchesSupportedTags(locale.languageTag))
+}
+
+const locale = pickSupportedLocale()
+
+export let isRTL = false
+
+// Need to set RTL ASAP to ensure the app is rendered correctly. Waiting for i18n to init is too late.
+if (locale?.languageTag && locale?.textDirection === "rtl") {
+  I18nManager.allowRTL(true)
+  isRTL = true
+} else {
+  I18nManager.allowRTL(false)
+}
 
 export const initI18n = async () => {
   i18n = i18next.use(initReactI18next)
 
   await i18n.init({
-    resources: {
-      ar,
-      en,
-      "en-US": en,
-      ko,
-      es,
-      fr,
-      ja,
-      hi,
-    },
-    lng: fallbackLocale,
+    resources,
+    lng: locale?.languageTag ?? fallbackLocale,
     fallbackLng: fallbackLocale,
     interpolation: {
       escapeValue: false,
     },
   })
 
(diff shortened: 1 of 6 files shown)


boilerplate/app/models/Episode.test.ts
```diff
@@ -27,14 +27,14 @@ const episode = EpisodeModel.create(data)
 test("publish date format", () => {
   expect(episode.datePublished.textLabel).toBe("Jan 20, 2022")
   expect(episode.datePublished.accessibilityLabel).toBe(
-    'demoPodcastListScreen:accessibility.publishLabel {"date":"Jan 20, 2022"}',
+    "demoPodcastListScreen:accessibility.publishLabel",
```
fpena, line 30: Can you explain please why are we removing the date? Read from your PR description, we want to still improve this test later.

boilerplate/app/i18n/i18n.ts
```diff
@@ -13,56 +13,51 @@ import fr from "./fr"
...
-const systemLocaleTag = systemLocale?.languageTag ?? fallbackLocale
+const resources = { ar, en, ko, es, fr, ja, hi }
+const supportedTags = Object.keys(resources)
+
+// Checks to see if the device locale matches any of the supported locales
+// Device locale may be more specific and still match (e.g., en-US matches en)
+const systemTagMatchesSupportedTags = (deviceTag: string) => {
+  const primaryTag = deviceTag.split("-")[0]
+  return supportedTags.includes(primaryTag)
+}
```
fpena, line 28: What it be worth writing a test for this one?

fpena (commented):
Looks good overall. Left 2 comments.

boilerplate/app/models/Episode.test.ts
lindboe: I don't really understand why this changed, but I also didn't want to block this on this test, because the behavior is correct when I actually test it in the app.

boilerplate/app/i18n/i18n.ts
lindboe: I was thinking splitting on `-` and checking `includes` were simple enough it didn't need a test, if we did something more complex maybe then?

Most of the complexity here comes in with third party systems, so an E2E test to check that we can switch languages makes the most sense to me.

boilerplate/app/models/Episode.test.ts
lindboe: The single quotes changing to double quotes was required by the linter, and I think might be related to tests having a separate TS config? but not sure, I thought that would be handled by prettier.

Actually I'm even more confused by this because apparently it failed on the CLI project's prettier check (https://app.circleci.com/pipelines/github/infinitered/ignite/3631/workflows/6e7e16e3-e4c1-4a6b-949b-d1fe26da1665/jobs/4375), but the inner project isn't complaining about the double-quotes.

That config just isn't unified, so we get weird things like this sometimes.

boilerplate/app/i18n/i18n.ts
fpena: Yeah, I agree this is not complex, but I'm also thinking of future Felipe doing an i18 migration again. It's all good; not a blocker.

boilerplate/app/models/Episode.test.ts
fpena: My question was not related to quotes but related to deleting the date within it.

boilerplate/app/models/Episode.test.ts
```diff
@@ -27,14 +27,14 @@ const episode = EpisodeModel.create(data)
...
   )
 })
 
 test("duration format", () => {
   expect(episode.duration.textLabel).toBe("42:58")
   expect(episode.duration.accessibilityLabel).toBe(
-    'demoPodcastListScreen:accessibility.durationLabel {"hours":0,"minutes":42,"seconds":58}',
+    "demoPodcastListScreen:accessibility.durationLabel",
```
fpena, line 37: Same as [here](https://github.com/infinitered/ignite/pull/2788/files#r1777945021).

boilerplate/app/models/Episode.test.ts
lindboe: Yes, for that part see my first comment. I'm not sure why it changed, but I'd also like to not block this PR on that since the logic is working correctly on the real app, this is just a test that I'm not sure is testing anything useful the way it's currently set up.

boilerplate/app/i18n/i18n.ts
lindboe: For sure. I still think in that case what is most useful is an E2E test instead of unit tests.

mazenchami (approved):
this looks good to me, great work!

fpena (approved):
Looks good, thanks Lizzi!

boilerplate/app/i18n/i18n.ts
lindboe: Filed issue for this: https://github.com/infinitered/ignite/issues/2789

boilerplate/app/models/Episode.test.ts
lindboe: Filed issue for this: https://github.com/infinitered/ignite/issues/2789

Issue #333, opened 2016-08-29
Problem passing props from one scene to another in ignite


sayedatif:
Hey guys i want to pass props from one scene to another using ignite as a boilerplate application but could not achieve success . Here is the code through which i am pushing it
`onPress={() => this.props.navigator.push({Routes.ModalData, passProps : {modalData: this.state.modalData},})}`

and on the another scene i am calling it

```
constructor (props) {
    super(props)
    this.state = {
      modalData: this.props.modalData,
    }
  }
```
can anybody help me how to pass props in ignite? Thanks in advance

skellock:
Looks like you're using the older style of Navigation.  In the new `react-native-router-flux` implementation it's really straight forward.

In your version, have a look at your Router.js.  See the `renderScene` function?  See the `createElement` function call?  Add this to the end of the 2nd parameter (right after the `navigator: navigator` line:

`...route.passProps`

sayedatif:
@skellock  do you need to update my onPress after adding that line currently it is

`onPress={() => this.props.navigator.push(Routes.ModalData)}`

and in the other scene i am just calling the props in the constructor

skellock:
Yes.  That will need to be changed to become a function to pass through the variable.

sayedatif:
@skellock  i am getting error while i added a line in ...route.passProps it says unexpected token

here is the snipet
```
return React.createElement(route.component, {
      navigator: navigator ...route.passProps
    })
```

skellock:
It's another key, so you'll need a comma.

sayedatif:
@skellock can you tell me why i am getting an error called  undefined is not an object on button press
here is the function through which i am passing props
```
modalButtonPress() {
    this.props.navigator.push({
      component: require('../Containers/ModalData').default,
      passProps: {modaldata: this.state.modaldata}
    });
  }
```
can you correct me where i am going wrong?

skellock:
`this` is not bound properly.  A couple of ways out, but try this in your constructor:

`this.modalButtonPress = this.modalButtonPress.bind(this)`.

sayedatif:
@skellock after passing props successfully i can see the receive data in console.log but when i am trying to print the receive props using 
 ``` 
 <View style={style.mainContainer}>
        <Text>{this.state.modalData}</Text>
      </View> 
```
[...]

PR #2794, opened 2024-10-01
Update PR description to add feature checks and possible screenshots/videos

fpena:

## Describe your PR

This PR updates Ignite's PR template with:
* Add a new check to enforce maintainers to test features
* Add new section for before/after screenshots

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -14,6 +14,6 @@
 
 If you’re submitting code changes related to a visible feature, please include before-and-after screenshots or videos. This helps the reviewer easily compare and understand the changes.
 
-| Before | After |
-|--------|-------|
+| Before                           | After                           |
+| -------------------------------- | ------------------------------- |
 | [Insert before screenshot/video] | [Insert after screenshot/video] |
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,7 +6,6 @@
 - [ ] `yarn lint` **eslint** checks pass with new code, if relevant
 - [ ] `yarn format:check` **prettier** checks pass with new code, if relevant
 - [ ] `README.md` (or relevant documentation) has been updated with your changes
-- [ ] Include issue number in title and description
 - [ ] If this affects functionality there aren't tests for, I manually tested it, including by generating a new app locally if needed ([see docs](https://docs.infinite.red/ignite-cli/contributing/Contributing-To-Ignite/#testing-changes-from-your-local-copy-of-ignite)).
 
 ## Describe your PR
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -17,7 +17,3 @@ If you’re submitting code changes related to a visible feature, please include
 | Before | After |
 |--------|-------|
 | [Insert before screenshot/video] | [Insert after screenshot/video] |
-
-
-
-
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,4 +1,4 @@
-## For issue #
+## For issue
 
 ## Please verify the following:
 
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,3 +6,15 @@
 - [ ] `README.md` (or relevant documentation) has been updated with your changes
 
 ## Describe your PR
+
+## Screenshots (if applicable)
+
+If you’re updating code related to a visible feature, please include before-and-after screenshots or videos. This helps the reviewer easily compare and understand the changes.
+
+| Before | After |
+|--------|-------|
+| [Insert before screenshot] | [Insert after screenshot] |
+
+
+
+
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,9 +1,12 @@
+## For issue #
+
 ## Please verify the following:
 
 - [ ] `yarn test` **jest** tests pass with new tests, if relevant
 - [ ] `yarn lint` **eslint** checks pass with new code, if relevant
 - [ ] `yarn format:check` **prettier** checks pass with new code, if relevant
 - [ ] `README.md` (or relevant documentation) has been updated with your changes
+- [ ] Include issue number in title and description
 - [ ] (If applicable) The feature related to this PR was tested across various use cases to ensure it works correctly.
 
 ## Describe your PR
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -16,6 +16,8 @@
 
 If you’re submitting code changes related to a visible feature, please include before-and-after screenshots or videos.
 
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,8 +1,22 @@
+## For issue
+
 ## Please verify the following:
 
 - [ ] `yarn test` **jest** tests pass with new tests, if relevant
 - [ ] `yarn lint` **eslint** checks pass with new code, if relevant
 - [ ] `yarn format:check` **prettier** checks pass with new code, if relevant
 - [ ] `README.md` (or relevant documentation) has been updated with your changes
+- [ ] Include issue number in title and description
+- [ ] (If applicable) The feature related to this PR was tested across various use cases to ensure it works correctly.
```
lindboe, line 10: I think if we get a little more specific this will be more clear, something like:

```suggestion
- [ ] If this affects functionality there aren't tests for, I manually tested it, including by generating a new app locally if needed ([see docs](https://docs.infinite.red/ignite-cli/contributing/Contributing-To-Ignite/#testing-changes-from-your-local-copy-of-ignite)).
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,8 +1,22 @@
...
 
 ## Describe your PR
+
+[Add your PR description]
+
+## Screenshots (if applicable)
+
+If you’re submitting code changes related to a visible feature, please include before-and-after screenshots or videos. This helps the reviewer easily compare and understand the changes.
```
lindboe, line 18:
```suggestion
If you’re submitting code changes related to a visible feature, please include before-and-after screenshots or videos.
```

I prefer this shorter explanation because:

1. Simple instructions
2. even with screenshots/videos, I prefer reviewers also run the code themselves on non-trivial changes. the screenshots/videos are also to help make sure the implementer tested it out.

lindboe (commented):
left some suggestions for edits, let me know what you think

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,8 +1,22 @@
+## For issue
+
 ## Please verify the following:
 
 - [ ] `yarn test` **jest** tests pass with new tests, if relevant
 - [ ] `yarn lint` **eslint** checks pass with new code, if relevant
 - [ ] `yarn format:check` **prettier** checks pass with new code, if relevant
 - [ ] `README.md` (or relevant documentation) has been updated with your changes
+- [ ] Include issue number in title and description
```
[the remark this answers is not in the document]
fpena: For 2, when I say description, I meant the first part of the template. I'm fine with the change you suggested above.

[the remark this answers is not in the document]
fpena: Also, I really like the issue number in the title. It makes it easier to search from the PR list, and it forces the creator to match with an issue.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,8 +1,23 @@
+## For issue
+
```
lindboe, line 2: (I think we don't need this section if it's in the description section, right?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,8 +1,23 @@
...
 
 ## Describe your PR
+
+<!-- If this PR addresses an issue, link to it in description: "Addresses #333" -->
+
+[Add your PR description]
+
+## Screenshots (if applicable)
+
+If you’re submitting code changes related to a visible feature, please include before-and-after screenshots or videos.
```
frankcalise, line 17: Could we include an example of HTML markup for resizing images that people could reference.

```html
<!-- If GH's auto attachment previews too large, trying resizing it with: <img src="filename-from-github-upload" width="300" /> -->
```

.github/PULL_REQUEST_TEMPLATE.md
frankcalise: Agree

frankcalise (commented):
Thanks, left some feedback for helping resize images so we don't end up doing this work each PR 😆

.github/PULL_REQUEST_TEMPLATE.md
fpena: In my experience and within a markdown table, Github resizes the image to fit the space. And works also when having multiple columns.

fpena: But also, some people can delete the table and add the image directly, which in that case could be a large one.

lindboe: for sure, maybe just including this as a comment might be useful to people who want quick access to how to do it. otherwise it can be ignored.

.github/PULL_REQUEST_TEMPLATE.md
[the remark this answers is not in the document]
fpena: Agreed, I haven't noticed that field. 

Ok so then, let's rely on that plus the comment added in this PR. 
Should we update the comment from:

`<!-- If this PR addresses an issue, link to it in description: "Addresses #333" -->`

to 

`<!-- If this PR addresses an issue, link to it in description: "Closes #333" -->`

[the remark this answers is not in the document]
frankcalise: Using the `Closes #xyz` will label it under the "Development" field for you, I use this quite a bit. Sometimes the Development field makes you choose the repo (unsure why, as you're already there) and then find the issue name. Either work, but I know that syntax works - unsure about Addresses

As far as an issue for each PR, we don't always have them. Sometimes an idea pops in your head and you just go for it (or have discussed it with the team if it needs discussing). So for IR folks, I don't think it is as much of a requirement.

For outside contributions though, it would be nice to understand what the issue is that they are trying to solve, but that of course could be done with a good PR write up as well, not really a requirement for me although low gafo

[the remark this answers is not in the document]
fpena: I agree! We leave it as a suggestion if the issue exists. Changed the working from "Addresses" to "Closes".

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -4,5 +4,20 @@
...
 
 ## Describe your PR
+
+<!-- If this PR addresses an issue, link to it in description: "Closes #333" -->
+
+[Add your PR description]
+
+## Screenshots (if applicable)
+
+If you’re submitting code changes related to a visible feature, please include before-and-after screenshots or videos.
```
frankcalise, line 17: nit: this should probably be an instructional comment so the user doesn't have to backspace it out?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -4,5 +4,20 @@
 - [ ] `yarn lint` **eslint** checks pass with new code, if relevant
 - [ ] `yarn format:check` **prettier** checks pass with new code, if relevant
 - [ ] `README.md` (or relevant documentation) has been updated with your changes
+- [ ] If this affects functionality there aren't tests for, I manually tested it, including by generating a new app locally if needed ([see docs](https://docs.infinite.red/ignite-cli/contributing/Contributing-To-Ignite/#testing-changes-from-your-local-copy-of-ignite)).
 
 ## Describe your PR
+
+<!-- If this PR addresses an issue, link to it in description: "Closes #333" -->
+
+[Add your PR description]
```
frankcalise, line 13: nit: this should probably be an instructional comment so the user doesn't have to backspace it out?

frankcalise (approved):
Some optional things to consider but looks good to me! Thanks!

PR #2799, opened 2024-10-12
Create .nvmrc to enforce a minimum version of Nodejs for nvm users

thehorse2000:
I feel like it's important to set a minimum version of nodejs for nvm users at least.Since many devs today use nvm for node version management.


lindboe (approved):
@thehorse2000 Thanks for this! I think this does make sense to have, especially because some of our tooling doesn't work correctly on lower node versions.

Our team does use multiple different version managers, so I think we should probably _also_ have a .node-version file. We can follow up on that if you don't want to add it yourself.

@frankcalise any thoughts on this?

thehorse2000:
Hey, Thanks for the fast review.
I can add the .node-version beside the .nvmrc since nvm doesn't respect the .node-version file. So i think we can keep both maybe?
but also another question is what do you think can be the recommended node version should we go with 20.18.0 or a more recent version ?

frankcalise:
We already have the ["engines"](https://github.com/infinitered/ignite/blob/master/package.json#L45-L46) key specified in `package.json`

Does `nvm` not respect that? This is what I see with `asdf` - I set it over what I had locally to induce the failure.

```bash
yarn install v1.22.22
warning ../../package.json: No license field
[1/5] 🔍  Validating package.json...
error ignite-cli@10.0.0: The engine "node" is incompatible with this module. Expected version "^18.18.0 || >=20.12.0". Got "20.10.0"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

thehorse2000:
Yea unfortunately nvm doesn't respect that aswell for some reason.
That's why I ran into this issue, I ran the project with node 14 which is what I use for other older projects, and of course it gave many errors.
Seems like it might be fine for other managers but nvm only sees .nvmrc, Hence the PR.

frankcalise:
@thehorse2000 understood, would you mind trying [this solution](https://github.com/infinitered/ignite/pull/2805) on `chore/node-check-preinstall`? 

This would hopefully address the issue but not be specific to the tooling involved

Thanks for bringing this to our attention!

frankcalise (changes requested):
Thanks! Can you just update this to be a min of 20 rather than this minor version

frankcalise (approved):
Thanks!

PR #2805, opened 2024-10-15
chore: check node version during preinstall

frankcalise:

## Describe your PR
- An alternative to #2799 which only addresses `nvm` users, checks the node version prior to installing dependencies
- `nvm` doesn't respect `package.json`'s `engines.node` key (while it seems `asdf` does), but `scripts/node-version-check.js` was added to do this across all tooling

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "preinstall": "node ./scripts/check-node-version.js"
   },
   "engines": {
-    "node": ">=20.10.0"
+    "node": ">=20.0.0"
   },
   "dependencies": {
     "deepmerge-json": "^1.1.0",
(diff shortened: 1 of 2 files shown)


lindboe:
@frankcalise I do think this is a significant downgrade from adding `.nvmrc` for nvm users; `nvmrc` can be used to really simply install the correct version immediately (and in my case, it'll switch automatically when I switch directories).

Given how much nvm doesn't seem to play nicely with other node version managers in the ecosystem, though, I've been considering switching.

frankcalise:
@lindboe asdf does that also if you add `.tool-versions` file, but do we really want to add a file for each possible node version manager? I guess supporting the two main ones (that I know of, at least) would be fine

Also, I could have sworn nvm does work against the `engines` key (I used to use nvm until recently). Can you test with `nvm use` and go outside the engine range that is defined?

lindboe:
> Can you test with nvm use and go outside the engine range that is defined?

I can run `nvm use 16` on a project that specifies >=18 just fine, and running `nvm use` with no args doesn't pick up the `engines` info or anything, it just says "No .nvmrc file found".

> asdf does that also if you add .tool-versions file, but do we really want to add a file for each possible node version manager? I guess supporting the two main ones (that I know of, at least) would be fine

I was actually thinking the two of `.nvmrc` and `.node-version`, if any, since node is the main thing we want to control, and AFAIK nvm is the main version manager that doesn't support `.node-version`. Which I think is fairly unreasonable, though.

PR #2812, opened 2024-10-17
Fix/eslint newline error in model generator

coolsoftwaretyler:

## Describe your PR

In Ignite X, the model generator would create files that did not pass Prettier config. See "before" screenshot, which came from `npx ignite-cli generate model Pizza` in a fresh Ignite X app.

This change primarily fixes the prettier error, as seen in the "after" screenshot.

However, when I attempted to update the Jest snapshots, I ran into this issue from Jest: https://github.com/jestjs/jest/issues/14305

One workaround is to disable prettier in Jest, which you'll see in cb78066ea24539569e21dcc20fbf9d5b9fe0c09c. We could drop that commit and just take the snapshot changes from 89841942359faa1c434c000a018d04e27d834777. I'm not sure if there's some other workaround or configuration in place that fixes that problem.

| Before                           | After                           |
| 2024-10-17-eslint-error-with-missing-newline| 2024-10-17-missing-newline-fixed |


package.json
```diff
@@ -159,6 +159,7 @@
       "tsx",
       "js",
       "jsx"
-    ]
+    ],
+    "prettierPath": null
```
coolsoftwaretyler, line 163: Workaround for https://github.com/jestjs/jest/issues/14305

test/vanilla/ignite-generate.test.ts
```diff
@@ -57,37 +57,39 @@ describe("ignite-cli generate", () => {
         "
       `)
       expect(read(`${TEMP_DIR}/app/models/Pizza.ts`)).toMatchInlineSnapshot(`
-        "import { Instance, SnapshotIn, SnapshotOut, types } from \\"mobx-state-tree\\"
-        import { withSetPropAction } from \\"./helpers/withSetPropAction\\"
+"import { Instance, SnapshotIn, SnapshotOut, types } from \\"mobx-state-tree\\"
+import { withSetPropAction } from \\"./helpers/withSetPropAction\\"
```
coolsoftwaretyler, line 62: I would guess this diff came from disabling Prettier in Jest, so maybe we should find a different way to fix that, or else it will cause a lot more noisy churn after the fact with other inline snapshots.

test/vanilla/ignite-generate.test.ts
```diff
@@ -57,37 +57,39 @@ describe("ignite-cli generate", () => {
...
+  .views((self) => ({})) // eslint-disable-line @typescript-eslint/no-unused-vars
+  .actions((self) => ({})) // eslint-disable-line @typescript-eslint/no-unused-vars
+
+export interface Pizza extends Instance<typeof PizzaModel> {}
+export interface PizzaSnapshotOut extends SnapshotOut<typeof PizzaModel> {}
+export interface PizzaSnapshotIn extends SnapshotIn<typeof PizzaModel> {}
+export const createPizzaDefaultModel = () => types.optional(PizzaModel, {})
+
+// @mst remove-file
+"
```
coolsoftwaretyler, line 79: This is the key diff: newline at end of file

test/vanilla/ignite-generate.test.ts
coolsoftwaretyler: Another approach would be ec682f836daea6f9db7009d0205ecfe6cd3f8e48, which [uses the workaround of a separate Prettier version for Jest](https://github.com/jestjs/jest/issues/14305#issuecomment-1668738936). I set `prettier-2` to be `npm:prettier@2.8.8` which I saw was the last version before the update in https://github.com/infinitered/ignite/pull/2761/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519. 

Didn't seem to change the whitespace diff here, so maybe there's no real issue. Figured I'd at least provide the commit and we could easily drop it if the `prettierPath: null` is acceptable.

coolsoftwaretyler:
Gonna close this in favor of https://github.com/infinitered/ignite/pull/2910. I was able to make the Prettier v3/Jest compat changes a little simpler.

Issue #2840, opened 2024-11-09
Ignite-cli Project Installation Error: CommandError: Cannot find module 'ajv/dist/compile/codegen'


spartanlync:
### Describe the bug

Hello, Was interested in exploring this framework by creating a project and trying it out for myself.

Unfortunately, creating my demo project failed with the following error and I could not find any reference to this issue being raised on GitHub.

**CommandError: Cannot find module 'ajv/dist/compile/codegen'**

Please see my screenshot below of trying to build my ignite project using the latest code.
Note: I do not have ignite-cli installed in my computer

Doing some digging, i found two articles on a solution.

- https://stackoverflow.com/questions/78743456/how-can-i-fix-this-cannot-find-module-ajv-dist-compile-codegen-require-stack
- https://github.com/ajv-validator/ajv/issues/2443

But these fixes requires a package.json dependancy fix from ^6 to ^8, which is outside my scope for ignite.

So before I abandon this exploration of ignite, I thought i would post what i experienced and give the community a chance to resolve.

Any suggestions on next steps?

**Text of installation and error below**
```
LMIT-MacBook-Air!valor [~/Documents/Projects/my-react-native-studio]>npx ignite-cli@latest new ignite-expo-ble-app
(node:97871) ExperimentalWarning: Support for loading ES Module in require() is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
  
:white_check_mark: What bundle identifier? · com.spartanlync.ignite_expo_ble_app
:white_check_mark: Where do you want to start your project? · /Users/valor/Documents/Projects/my-react-native-studio/ignite-expo-ble-app
:white_check_mark: How do you want to manage Native code? · cng
:white_check_mark: Do you want to initialize a git repository? (Y/n) · Yes
:white_check_mark: Remove demo code? We recommend leaving it in if it's your first time using Ignite (y/N) · No
:white_check_mark: Which package manager do you want to use? · npm
:white_check_mark: Do you want to install dependencies? (Y/n) · Yes
:white_check_mark: [Experimental] Expo Router for navigation? (y/N) · Yes
:white_check_mark: [Experimental] the New Architecture? (y/N) · Yes
  
  
  
  
                        .;;,.  .clll.         
                         .clll,  ,llll.      ...   
                          .,;,.   ;lllc'.,.  .';cllllc, 
   .',,.                  ...    .'',,. .:lllllllll. .;lllc:;clll,
[...]

Issue #2818, opened 2024-10-20
eslint-plugin-n, eslint-config-standard conflicting in Ignite X


StefanWallin:
### Describe the bug

```
❯ npx ignite-cli@latest new RnExpoLiveActionsLab

✅ What bundle identifier? · se.stefan_wallin.lab
✅ Where do you want to start your project? · /Users/festiz/gitstuff/RnExpoLiveActionsLab
✅ How do you want to manage Native code? · Manual
✅ Do you want to initialize a git repository? (Y/n) · Yes
✅ Remove demo code? We recommend leaving it in if it's your first time using Ignite (y/N) · No
✅ Which package manager do you want to use? · npm
✅ Do you want to install dependencies? (Y/n) · Yes
✅ [Experimental] Expo Router for navigation? (y/N) · No
✅ [Experimental] the New Architecture? (y/N) · Yes

[...] 

The following error occurred:

    npm error code ERESOLVE
npm error ERESOLVE unable to resolve dependency tree
npm error
npm error While resolving: rn-expo-live-actions-lab@0.0.1
npm error Found: eslint-plugin-n@17.11.1
npm error node_modules/eslint-plugin-n
npm error   dev eslint-plugin-n@"^17.10.2" from the root project
npm error
npm error Could not resolve dependency:
npm error peer eslint-plugin-n@"^15.0.0 || ^16.0.0 " from eslint-config-standard@17.1.0
npm error node_modules/eslint-config-standard
npm error   dev eslint-config-standard@"^17.1.0" from the root project
npm error
npm error Fix the upstream dependency conflict, or retry
npm error this command with --force or --legacy-peer-deps
npm error to accept an incorrect (and potentially broken) dependency resolution.
```

### Ignite version

10.0.1

### Additional info

```
❯ npx ignite-cli doctor
System
  platform           darwin
  arch               arm64
  cpu                10 cores     Apple M1 Pro
  directory          gitstuff     /Users/festiz/gitstuff
```

frankcalise:
I guess for now cd into your directory and follow the npm install flag instructions it provides via `--legacy-peer-deps`.

This seems to be an open issue over at https://github.com/standard/eslint-config-standard/issues/412 and yarn must treat it differently.

Will discuss with @lindboe on how to fix this here, thanks for reporting

StefanWallin:
Switching to yarn gave me more fun stuff to deal with on initialization 😅... Wondering if it has something to do with the fact I'm using asdf to manage node versions...

[...]

PR #2842, opened 2024-11-10
fix: include ajv in devDependencies for Expo Router and npm

coolsoftwaretyler:

## Describe your PR

This would resolve one of the issues in https://github.com/infinitered/ignite/issues/2840, although there are additional problems with the Expo Router template.

This is a little bit of a sledgehammer approach, but I wanted to start with the simple solution and we can enhance it if that seems necessary.

In https://github.com/infinitered/ignite/pull/2823, we added the `--legacy-peer-deps` flag for `npm install` commands. This fixed a problem upstream from https://github.com/standard/eslint-config-standard/issues/412, but I believe it has opened up a new problem for users who choose to use Expo Router and npm. Based on the [issues that Expo Router + ESLint has experienced with how bun resolves peer dependencies](https://github.com/expo/expo/issues/29150), I am guessing the legacy peer dependency resolution falls victim to the same problem.

The very simplest way to solve this is if the generated Ignite project includes `ajv` at the right version number in its own `devDependencies`. This version "wins" across the board.

The drawback to this approach is that if Expo Router changes its required `ajv` version, there may be some new errors for the project to resolve.

Another drawback to this PR is that it would add `ajv` to *all* projects, including other package managers that don't suffer from this problem, and including projects that do not opt in to Expo Router. So we could alternatively run `npm install ajv@^8 --legacy-peer-deps` in the install command right after https://github.com/infinitered/ignite/blob/e59d880c951a8ec729ca65f39ba0dae1de27a729/src/tools/packager.ts#L149, and maybe even do a check for if Expo Router is enabled.

No screenshots, but running `node ~/ignite/bin/ignite new project-name` and choosing `npm` and Expo Router successfully works with this change.

### DIFF
--- a/boilerplate/package.json
+++ b/boilerplate/package.json
@@ -74,6 +74,7 @@
     "@testing-library/react-native": "^12.5.2",
     "@types/jest": "^29.2.1",
     "@types/react": "~18.2.14",
+    "ajv": "8.12.0",
     "babel-jest": "^29.2.1",
     "eslint": "^8.57.0",
     "eslint-config-expo": "^7.1.2",
--- a/boilerplate/package.json
+++ b/boilerplate/package.json
@@ -74,7 +74,6 @@
     "@testing-library/react-native": "^12.5.2",
     "@types/jest": "^29.2.1",
     "@types/react": "~18.2.14",
-    "ajv": "8.12.0",
     "babel-jest": "^29.2.1",
     "eslint": "^8.57.0",
     "eslint-config-expo": "^7.1.2",
--- a/src/commands/new.ts
+++ b/src/commands/new.ts
@@ -750,6 +750,11 @@ module.exports = {
         // do base install
         const installCmd = packager.installCmd({ packagerName })
         await system.run(installCmd, { onProgress: log })
+        // If they chose npm and also Expo Router, we need to run npm install ajv@^8 --legacy-peer-deps.
+        // see https://github.com/infinitered/ignite/issues/2840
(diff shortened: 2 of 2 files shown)


coolsoftwaretyler:
[16e30eb](https://github.com/infinitered/ignite/pull/2842/commits/16e30ebdc7333942102091e7b4c7565e9c366f77) is the more targeted approach, but we can easily drop that commit if simple is preferred.

coolsoftwaretyler:
Yeah, `eslint-plugin-n` needs to fix their peer dependencies, as in https://github.com/standard/eslint-config-standard/issues/412 (this is why we added `--legacy-peer-deps` for our own issue over in https://github.com/infinitered/ignite/issues/2818). Once `eslint-plugin-n` resolves their issue, we should be able to remove `--legacy-peer-deps` from our commands here and then remove this patch as well.

frankcalise:
@coolsoftwaretyler ah yeah I had a discussion about this in Slack. config-standard is abandoned I think for neostandard in that issue thread https://infinitered.slack.com/archives/C0X0P5LP7/p1730325079888599

should we look to swap it out for neo and see if n just works or cut back on all these extra config / plugins?

coolsoftwaretyler:
@frankcalise - if we swap in neostandard and it Just Works, I think that's ideal. I can take a look sometime this week to make sure it doesn't cause any changes while linting a freshly ignited project.

Want to merge this one for now to patch up the `ajv` issue in the short term and I'll get back to ya in a few days on neostandard?

PR #2843, opened 2024-11-14
feat: upgrade to Expo SDK 52

frankcalise:

## Describe your PR

- Upgrades to Expo SDK 52 and react-native 0.76.2
- General upgrade commands were:
```bash
yarn add@expo^52.0.0
npx expo install --fix
```

### CLI
- Removes New Architecture question from experimental flag, can use `--newArch`

### Boilerplate
- Updates Expo packages as per `npx expo install --fix`
- Reworks `expo-splash-screen` configuration via config plugin (as recommended) over app.json `splash` keys
- Remove `react-native-drawer-layout` patch now that v4 was released
- Update `react-native-mmkv` version for new architecture to be compatible with react-native 0.76.x
- Updates new arch flag at root expo key in app.json instead of build properties plugin as per changelog


src/commands/new.ts
```diff
@@ -158,6 +158,13 @@ export interface Options {
    * @default mst
    */
   state?: StateMgmt
+  /**
+   * Whether or not to enable the New Architecture
+   *
+   * Input Source: `prompt.ask` | `parameter.option`
+   * @default false
+   */
+  newArch?: boolean
```
jamonholmgren, line 167: Can you walk me through why you'd move this to its own flag rather than keeping it on the experimental side?

Is it because you want to keep the ability to drop back to old arch for a while?

Iannery: Since Expo/Meta is using the new arch by default, wouldn't it be better to have 
https://github.com/infinitered/ignite/blob/97ec737c6803f816474d67d6f72e88b90d351969/src/commands/new.ts#L165
and https://github.com/infinitered/ignite/blob/97ec737c6803f816474d67d6f72e88b90d351969/src/commands/new.ts#L521
as true instead, and keep the ability to drop back to old arch if needed?

frankcalise: @Iannery to align with them directly, technically yes. but we're discussing internally here whether or not we are ready to make that jump as we're still converting many of our client projects over, which is uncovering issues with several libraries we must use

but yeah that's why this is still a draft, that default is up for discussion

boilerplate/package.json
```diff
@@ -32,39 +30,39 @@
...
-    "react-native-gesture-handler": "~2.16.1",
+    "react-native": "0.76.1",
+    "react-native-drawer-layout": "^4.0.1",
+    "react-native-gesture-handler": "~2.20.2",
     "react-native-keyboard-controller": "^1.12.7",
-    "react-native-mmkv": "^2.12.2",
-    "react-native-reanimated": "~3.15.0",
-    "react-native-safe-area-context": "4.10.5",
-    "react-native-screens": "3.31.1",
+    "react-native-mmkv": "^3.1.0",
```
comertcimen, line 62: react-native-mmkv v3 requires newArchEnabled to be true

frankcalise: good catch thanks!

boilerplate/.eslintrc.js
```diff
@@ -26,6 +26,8 @@ module.exports = {
       },
     ],
     "@typescript-eslint/no-var-requires": 0,
+    "@typescript-eslint/no-require-imports": 0,
```
frankcalise, line 29: Seems like the new eslint rules from Expo turned this on, so I kept it off for consistency.

boilerplate/.eslintrc.js
```diff
@@ -26,6 +26,8 @@ module.exports = {
...
+    "@typescript-eslint/no-empty-object-type": 0,
```
frankcalise, line 30: Seems like the new eslint rules from Expo turned this on. I disabled it because we use this pattern quite a bit with some of the MST helpers.

RyanGst:
First of all, thank you so much for this @frankcalise

For anyone (just like me) wondering to get expo sdk 52 working:

```bash
npx ignite-cli@latest new MyApp
yarn add expo@^52.0.0
npx expo install --fix
```
> Enable new arch and run expo prebuild

https://github.com/infinitered/ignite/issues/2847#issuecomment-2486125532


add this to your `app.json`:

```json
{
  "expo": {
    "newArchEnabled": true
  }
}
```

and then run 

```bash
npx expo prebuild --clean && npx expo run:android
```

jamonholmgren (approved):
:goberserk:

PR #2883, opened 2025-01-11
fix: simplify ESLint config to improve dependency management

coolsoftwaretyler:

## Describe your PR

In https://github.com/infinitered/ignite/pull/2823, we added `--legacy-peer-deps` to the `npm` install script in order to try and resolve https://github.com/infinitered/ignite/issues/2818 (issues with conflicting `eslint-plugin-n` and `eslint-config-standard`).

I believe this led to some new bugs because of peer dependency resolution in npm, like https://github.com/infinitered/ignite/issues/2840. We have also had some reports in Slack where people are getting confused because they need to use the `--legacy-peer-deps` flag on subsequent dependency installs.

I was talking to @frankcalise yesterday about these plugins. They've been in Ignite for a while, and maybe it's simpler to eliminate eslint-plugin-n at this point. As far as I can tell, these are just [extra rules for Node.js](https://www.npmjs.com/package/eslint-plugin-n), which might not be very relevant in most React Native projects.

There was some support for slimming down linting configuration in [this Slack thread from October](https://infinitered.slack.com/archives/C0X0P5LP7/p1730325079888599). So I hope this PR aligns with that approach. I think it's reasonable to remove ESLint rules for Node.js in order to simplify a React Native project.

### Alternative approach

If we want to keep using the standard config and the `eslint-plugin-n`, we need to use [`neostandard`](https://github.com/neostandard/neostandard), which requires ESLint v9. But we are blocked on adopting ESLint v9 until React Native settles on support for ESLint v9: https://github.com/facebook/react-native/pull/43959, but that isn't ready yet.

No screenshots since there are no real UI changes, but here are two Ignite commands I ran from my local branch:

Standard settings, but use `npm` as the packager (just testing this PR as a replacement fix for https://github.com/infinitered/ignite/pull/2823)

```
   npx ignite-cli new npmfix \
        --bundle=com.npmfix \
        --git \
        --install-deps \
        --packager=npm \
        --target-path=/Users/tylerwilliams/npmfix \
        --remove-demo=false \
        --new-arch=false \
        --workflow=cng \
        --no-timeout=false \
        --state=mst
```

Standard settings, but use `npm` as the packager, and enable Expo Router (fixes https://github.com/infinitered/ignite/issues/2840, which is related to https://github.com/expo/expo/issues/29150)

```
      npx ignite-cli new npmfixexporouter \
        --bundle=com.npmfixexporouter \
        --git \
        --install-deps \
        --packager=npm \
        --target-path=/Users/tylerwilliams/npmfixexporouter \
        --remove-demo=false \
        --new-arch=false \
        --experimental=expo-router \
        --workflow=cng \
        --no-timeout=false \
        --state=mst
```

In both cases I ran `npm run android` and `npm run ios` from the project and Ignite started without issue.

### Failing with Yarn, now:

Now if I try to use yarn as my package manager, I end up with a new error after running `npx ignite-cli new yarncheck --yes`:

```
The following error occurred:

    yarn run v1.22.22
$ eslint . --fix

Oops! Something went wrong! :(

ESLint: 8.57.1

ESLint couldn't find the plugin "eslint-plugin-n".

(The package "eslint-plugin-n" was not found when loaded as a Node module from the directory "/Users/tylerwilliams/yarncheck".)

It's likely that the plugin isn't installed correctly. Try reinstalling by running the following:

    npm install eslint-plugin-n@latest --save-dev

The plugin "eslint-plugin-n" was referenced from the config file in ".eslintrc.js » eslint-config-standard".

If you still can't figure out the problem, please stop by https://eslint.org/chat/help to chat with the team.

error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```


coolsoftwaretyler:
The new `yarn` failure is what's failing CI.

frankcalise (approved):
This looks good to me. I know you didn't modify it but mind removing the `format` script from `package.json` in `boilerplate`? It seems to be a duplicate of `lint`. Since we have `lint:check` seems reasonable to keep `lint` over `format`

PR #2905, opened 2025-02-24
fix(generate): add support for Expo Router app directory prompts 

### LINKED ISSUE #2903: Generating a new screen when using Expo Router results in wrong placement
### Describe the bug

Reproduce:

```
npx ignite-cli@latest new PizzaApp
```

Keep the demo code in, use expo router.

```
npx ignite-cli@latest g screen About
```

Image

It adds it to `./app/screens`. Should be in `./src/app/screens`, I believe.

### Ignite version

10.1.6 (latest)

### Additional info

no

cdanwards:
## Description

- Issues: fixes #2903 

This PR addresses a user who creates screens within an ignited app using `expo-router`. It checks for the existence of `src/app` (the expo router configuration) and then determines where to put the screen in 4 ways: 
- An override provided in the CLI command
- `destinationDir` set from the screen generator template
- If neither of those exists, it prompts the user to provide a dir path, if none is provided or they decide to not create that directory, it defaults to `src/app/`

## Screenshots

| Scenario                           | Result                           |
| Using CLI Override for directory |  override |
| Using directory set in template front matter |  frontmatter |
| Setting via prompts - chooses to create dir |  created_dir |
| Setting via prompts - detects `src/app` and user just taps enter | CleanShot 2025-04-30 at 07 26 44@2x |
| Setting via prompts - goes to default `src/app` |  Screenshot 2025-02-27 at 10 57 11 AM |

## Checklist

- [X] `README.md` and other relevant documentation has been updated with my changes
- [X] I have manually tested this, including by generating a new app locally ([see docs](https://docs.infinite.red/ignite-cli/contributing/Contributing-To-Ignite/#testing-changes-from-your-local-copy-of-ignite)).


src/commands/generate.ts
```diff
@@ -52,6 +52,53 @@ async function generate(toolbox: GluegunToolbox) {
     pascalName = pascalName.slice(0, -1 * pascalGenerator.length)
     command(`npx ignite-cli generate ${generator} ${pascalName}`)
   }
+  // Check if src/app exists, denoting an Expo Router app
+  if (generator === "screen") {
+    const isExpoRouterApp = filesystem.exists("src/app") === "dir"
```
frankcalise, line 57: So this is the correct directory structure to determine expo-router in an Ignite application, however omitting the `src/` nesting is totally valid and it is possible a dir structure looks like this:

- app (the routes)
- components
- hooks
- etc...

This is in fact the default expo application structure and Ignite generators should continue to be compatible for any project, not just our opinionated structure.

So perhaps we should look into a different mechanism for determining this? Maybe we check `app` or `src/app` to cover both. Or look for `package.json` to have `expo-router`?

src/tools/generators.ts
```diff
@@ -354,6 +354,31 @@ export async function generateFromTemplate(
...
+export function frontMatterDirectoryDir(generator: string): "string" | undefined {
+  if (!validateGenerator(generator)) {
+    return undefined
+  }
+
+  const { path } = filesystem
+
+  // where are we copying from?
+  const templateDir = path(templatesDir(), generator)
+  console.log(templateDir)
```
frankcalise, line 369: ✂️

src/tools/generators.ts
```diff
@@ -354,6 +354,31 @@ export async function generateFromTemplate(
   return { written, exists, overwritten }
 }
 
+/**
+ * Checks a file for front matter.
```
frankcalise, line 358: We can make this more accurate, this function actually is looking for `destinationDir` to be present correct? Not just any front matter in the file

src/tools/generators.ts
```diff
@@ -354,6 +354,31 @@ export async function generateFromTemplate(
...
+  const templateDir = path(templatesDir(), generator)
+  console.log(templateDir)
+
+  const fileContents = filesystem.read(`${templateDir}/NAME.tsx.ejs`)
+  const { data: frontMatterData } = frontMatter(fileContents)
+
+  if (frontMatterData?.destinationDir === undefined) {
+    warning(`⚠️  No front matter found in ${generator} template.`)
+    return undefined
+  }
```
frankcalise, line 377: Do we always want to warn here? I maybe only want to see this if I haven't provided `--dir` because if that's my workflow and I never set it in the template, it's going to show each time.

frankcalise (commented):
Nice work! A few comments for discussion where I think we can shore things up. Also if you think we can add a test to double check the user gets prompted in the way you're expecting if both are omitted that's probably worth adding.

src/commands/generate.ts
cdanwards: Yeah, that definitely tracks. I think it would make sense to check for `expo-router`. Do you think it make more sense to check within the `package.json` dependencies or within the `app.json` expo plugins?

frankcalise: I'd say `package.json` is easier because you know where it will live, under `dependencies`, so you can search for the key there.

Issues with the `app.json` route:
- It could also be in `app.config.js` or `app.config.ts`
- Plugins is an array of a mix of strings and/or arrary, making searching through it a bit more difficult

cdanwards: Yeah that's fair, thanks!

src/commands/generate.ts
```diff
@@ -52,6 +52,54 @@ async function generate(toolbox: GluegunToolbox) {
...
+              format: prettyPrompt.format.boolean,
+            })
+
+            if (createDirResult.createDir) {
+              filesystem.dir(result.dir)
+              dir = result.dir
+            } else {
+              warning(`⚠️ Placing screen in src/app root.`)
+              p()
+              dir = "src/app"
```
frankcalise, line 96: I'm thinking this still needs to detect if `src/app` exists or `app` exists - as it could be either if not coming from an Igntie project

maybe we can set it to be the default of the prompt since it's the example in the message anyway

frankcalise (commented):
Looking good! Spotted a couple of things to look over

jamonholmgren:
Sorry about not chiming in on the original issue!

I have some thoughts.

1. I think the screen should be separated from the route.

```
npx ignite-cli g screen DetailScreen
```

The screen should always (IMO) be in `src/app/screens/detail-screen.tsx` (if it's an Expo Router app)

2. We need a route generator

```
npx ignite-cli g route listings/detail DetailScreen
```

This creates a route file at `app/listings/detail.tsx` which imports the specified screen.

This would allow us to have multiple routes that can import the same screen.

It also eliminates this specialized generator logic which I think is going away from the spirit of the simple generators we want for Ignite. Ignite's generators got too complicated in earlier versions, which is why I rewrote them to be super simple in Ignite 6. The whole CLI was just over 900 lines of code at that point (vs 3k+ for previous versions), and I really don't want to continue to bloat it with additional logic if we can avoid that.

frankcalise:
I made some updates to this to help progress it along. 

Now, for expo-router projects, we restored the screen generator to keep it being placed in `src/screens`, matching react navigation ignite projects (aside from the root dir)

In addition to that, upon a new project it will generate a `route` generator template which you can use like so:

`npx ignite-cli g route welcome --dir=src/app/some/route`

resulting in 

`src/app/some/route/welcome.tsx` and renders

```tsx
import { WelcomeScreen } from "@/screens/WelcomeScreen"

export default function Welcome() {
  return <WelcomeScreen />
}
```

frankcalise (approved):

boilerplate/src/app/index.tsx
```diff
@@ -1,84 +1,5 @@
...
-        />
-      </View>
-
-      <View style={[themed($bottomContainer), $bottomContainerInsets]}>
-        <Text tx="welcomeScreen:postscript" size="md" />
-      </View>
-    </Screen>
-  )
+export default function Index() {
+  return <WelcomeScreen />
```
markrickert, line 4: Praise: I love this! Less duplicated code in the boilerplate to maintain! 🎉

src/tools/react-native.ts
```diff
@@ -228,24 +228,15 @@ export async function replaceMaestroBundleIds(
...
+ * Defines an ejs template for a screen when using Expo Router.
+ */
+export const EXPO_ROUTER_SCREEN_TEMPLATE = `---
+destinationDir: src/screens
+---
 import { ViewStyle } from "react-native"
-import { Screen } from "@/components/Screen"
-import { Text } from "@/components/Text"
+
+import { Screen, Text } from "@/components"
```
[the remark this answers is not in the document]
frankcalise: Good catch, thanks!

boilerplate/src/app/index.tsx
frankcalise: Woot think I can reduce it further, changes incoming 🙏

PR #2765, opened 2024-09-06
 Allow passing ref to Text component in Ignite boilerplate #2758 

Aniganesh:
## Please verify the following:

- [x] `yarn test` **jest** tests pass with new tests, if relevant
- [x] `yarn lint` **eslint** checks pass with new code, if relevant
- [x] `yarn format:check` **prettier** checks pass with new code, if relevant
- [x] `README.md` (or relevant documentation) has been updated with your changes

## Describe your PR
This PR allows passing ref to the `Text` component in boilerplate. 
Have added the feature and a demo in Demo Community Screen. 
@frankcalise Let me know if I need to add something or if things need to be moved around.

PR #2909, opened 2025-03-06
Allow passing a ref prop to the Text component

### LINKED ISSUE #2758: Allow passing ref to `Text` component in Ignite boilerplate
Right now, the text component does not allow passing a ref. Would be nice if we could add that. Happy to open a PR as well. I did add the functionality on a prj I was working on anyway.

Aniganesh:
## Description

- Issues: fixes #2758 

## Screenshots

Demo page:

 

## Checklist

### DIFF
--- a/boilerplate/app/components/Text.tsx
+++ b/boilerplate/app/components/Text.tsx
@@ -53,10 +53,7 @@ export interface TextProps extends RNTextProps {
  * @param {TextProps} props - The props for the `Text` component.
  * @returns {JSX.Element} The rendered `Text` component.
  */
-export const Text = forwardRef(function Text(
-  props: TextProps,
-  ref: ForwardedRef<RNText>,
-) {
+export const Text = forwardRef(function Text(props: TextProps, ref: ForwardedRef<RNText>) {
   const { weight, size, tx, txOptions, text, children, style: $styleOverride, ...rest } = props
   const { themed } = useAppTheme()
 
--- a/boilerplate/app/components/Text.tsx
+++ b/boilerplate/app/components/Text.tsx
@@ -4,7 +4,7 @@ import { isRTL, translate, TxKeyPath } from "@/i18n"
 import type { ThemedStyle, ThemedStyleArray } from "@/theme"
 import { useAppTheme } from "@/utils/useAppTheme"
 import { typography } from "@/theme/typography"
-import { ReactNode, forwardRef, PropsWithChildren, ForwardedRef } from "react"
+import { ReactNode, forwardRef, ForwardedRef } from "react"
 
 type Sizes = keyof typeof $sizeStyles
 type Weights = keyof typeof typography.primary
@@ -54,8 +54,8 @@ export interface TextProps extends RNTextProps {
  * @returns {JSX.Element} The rendered `Text` component.
  */
 export const Text = forwardRef(function Text(
-  props: PropsWithChildren<TextProps>,
-  ref: ForwardedRef<RNText | null>,
+  props: TextProps,
+  ref: ForwardedRef<RNText>,
 ) {
   const { weight, size, tx, txOptions, text, children, style: $styleOverride, ...rest } = props
   const { themed } = useAppTheme()
--- a/boilerplate/app/components/Text.tsx
+++ b/boilerplate/app/components/Text.tsx
@@ -4,7 +4,7 @@ import { isRTL, translate, TxKeyPath } from "@/i18n"
 import type { ThemedStyle, ThemedStyleArray } from "@/theme"
 import { useAppTheme } from "@/utils/useAppTheme"
 import { typography } from "@/theme/typography"
-import { ReactNode } from "react"
+import { ReactNode, forwardRef, PropsWithChildren, ForwardedRef } from "react"
 
 type Sizes = keyof typeof $sizeStyles
 type Weights = keyof typeof typography.primary
@@ -53,7 +53,10 @@ export interface TextProps extends RNTextProps {
  * @param {TextProps} props - The props for the `Text` component.
  * @returns {JSX.Element} The rendered `Text` component.
  */
-export function Text(props: TextProps) {
+export const Text = forwardRef(function Text(
+  props: PropsWithChildren<TextProps>,
+  ref: ForwardedRef<RNText | null>,
+) {
   const { weight, size, tx, txOptions, text, children, style: $styleOverride, ...rest } = props
   const { themed } = useAppTheme()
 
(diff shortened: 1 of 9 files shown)


boilerplate/app/components/Text.tsx
```diff
@@ -53,7 +53,10 @@ export interface TextProps extends RNTextProps {
  * @param {TextProps} props - The props for the `Text` component.
  * @returns {JSX.Element} The rendered `Text` component.
  */
-export function Text(props: TextProps) {
+export const Text = forwardRef(function Text(
+  props: PropsWithChildren<TextProps>,
```
coolsoftwaretyler, line 57:
```suggestion
  props: TextProps,
```

**suggestion:** I don't think we need `PropsWithChildren` here since we already have it in the `TextProps` interface

boilerplate/app/components/Text.tsx
```diff
@@ -53,7 +53,10 @@ export interface TextProps extends RNTextProps {
...
+  ref: ForwardedRef<RNText | null>,
```
coolsoftwaretyler, line 58:
```suggestion
  ref: ForwardedRef<RNText>,
```

**suggestion:** I don't think we want `null` here. `RNText` should suffice.

boilerplate/app/screens/DemoShowroomScreen/demos/DemoText.tsx
```diff
@@ -1,9 +1,57 @@
...
 import { Demo } from "../DemoShowroomScreen"
 import { DemoDivider } from "../DemoDivider"
 import { DemoUseCase } from "../DemoUseCase"
 import { translate } from "@/i18n"
+import { Button } from "../../../components"
+import { TextInput, View } from "react-native"
+import { useAppTheme } from "@/utils/useAppTheme"
+
+// Define the TextRefDemo component outside of other function components
+const TextRefDemo = () => {
```
coolsoftwaretyler, line 13: **praise:** nice! Thanks for the demo.

coolsoftwaretyler, line 13: **suggestion:** maybe we should extract `TextRefDemo` to a separate file? I feel very neutral on this, but the other demo files seem to just be the demo objects. I'm not sure if there's a good example of something like this, though.

boilerplate/app/i18n/demo-fr.ts
```diff
@@ -3,7 +3,7 @@ import { DemoTranslations } from "./demo-en"
 export const demoFr: DemoTranslations = {
   demoIcon: {
     description:
-      "Un composant pour faire le rendu d’une icône enregistrée. Il est enveloppé dans un <TouchableOpacity /> si `onPress` est fourni, sinon dans une <View />.",
+      "Un composant pour faire le rendu d'une icône enregistrée. Il est enveloppé dans un <TouchableOpacity /> si `onPress` est fourni, sinon dans une <View />.",
```
coolsoftwaretyler, line 6:
```suggestion
      "Un composant pour faire le rendu d’une icône enregistrée. Il est enveloppé dans un <TouchableOpacity /> si `onPress` est fourni, sinon dans une <View />.",
```

**suggestion:** can we revert these changes? I'm not sure how much it matters, but it looks like maybe you had a format-on-save change a bunch of unrelated characters in the french localization file.

boilerplate/app/i18n/demo-ar.ts
```diff
@@ -393,6 +393,13 @@ export const demoAr: DemoTranslations = {
         text3:
           "Eiusmod occaecat laboris eu ex veniam ipsum adipisicing consectetur. Magna ullamco adipisicing tempor adipisicing.",
       },
+      refs: {
```
coolsoftwaretyler, line 396: **question:** @frankcalise - do we have a process for reviewing localization files across languages? I can run this through google translate but I'm curious if we've got a better mechanism for that across each locale.

boilerplate/app/i18n/demo-en.ts
```diff
@@ -389,6 +389,13 @@ export const demoEn = {
         text3:
           "Eiusmod occaecat laboris eu ex veniam ipsum adipisicing consectetur. Magna ullamco adipisicing tempor adipisicing.",
       },
+      refs: {
```
coolsoftwaretyler, line 392: **praise:** thank you for doing the localization!

boilerplate/app/screens/DemoShowroomScreen/demos/DemoText.tsx
```diff
@@ -1,9 +1,57 @@
...
+  const [highlighted, setHighlighted] = useState(false)
+  const { theme } = useAppTheme()
+
+  const handleFocusPress = () => {
+    if (textRef.current) {
+      // Simulate highlighting by changing state
+      setHighlighted(true)
+
+      // Reset highlight after a delay
+      setTimeout(() => {
```
coolsoftwaretyler, line 24: **suggestion:** rather than resetting the highlight, can we make the button toggle the focus? This is a personal preference - sometimes I look away and miss a time-based demo like this. Not a blocker.

coolsoftwaretyler (commented):
Hey @Aniganesh - this looks great. Thanks for writing it up against the latest Ignite. I see you had submitted https://github.com/infinitered/ignite/pull/2765 earlier. We appreciate your time!

I have a couple questions, thoughts, suggestions. I don't see anything that I consider a blocker, but I'll see if anyone else from IR wants to take a look before merge.

boilerplate/app/i18n/demo-ar.ts
frankcalise: Yeah it isn't as simple as just pushing strings through Google Translate, we've reached out to various parts of the community for official translations.

I'm going to say we don't need to have a use case for this demo, just accept the forwardRef on the component and that should tidy up this PR a bunch.

frankcalise (changes requested):
Thanks for updating your contribution!

If you wouldn't mind, can we revert the `i18n` and `DemoText.tsx` changes. It doesn't feel necessary to demo a ref since it's not really something Ignite is offering over how React works.

It'll also help keep the maintenance down for translating more strings and we'd also have to even verify that the ones here are valid.

Aniganesh:
Will update

boilerplate/app/i18n/demo-fr.ts
frankcalise, line 1: I think you made some unintended changes here to this file

frankcalise (changes requested):

boilerplate/app/i18n/demo-fr.ts
aniruddha-mithya: fixed

frankcalise (approved):
Thank you!

PR #2912, opened 2025-03-10
fix: add `android.allowBackup: false` to expo app config

markrickert:
## Description

Expo defines `android.allowBackup` as `true` by default. 😱  https://docs.expo.dev/versions/latest/config/app/#allowbackup

This should be opt-in behavior, imho. Setting this value to false will prevent apps from automatically backing up the app’s documents directory to google drive on the user’s behalf. 

Developers should have to opt into this behavior by manually changing this to true or removing the value from the `app.json` file.


markrickert:
Wondering if we should convert this `app.json` to `app.config.ts` so that we can annotate all the options. No comments in json files stinks.

frankcalise (approved):

PR #2917, opened 2025-03-24
fix(cli): reorder demo app question

### LINKED ISSUE #2916: Doc/CLI suggestion - Expo router removes demo code
Hey! First off I’m knee deep in making my first react native app with Ignite and it is a serious shortcut. Thank you for the amazing work.

I brought this up in the slack, and they requested I open an issue so here we are. As it stands, when you start a new Ignite application, if you start with an expo router application, it removes all the demo boilerplate, and the CLI doesn’t tell you this. In fact, the CLI allows you to explicitly include the demo code and then select expo only to remove it. Below is a copy paste of a test to show the current output:

✅ Remove demo code? We recommend leaving it in if it's your first time using Ignite (y/N) · No
✅ Which package manager do you want to use? · pnpm
✅ Do you want to install dependencies? (Y/n) · Yes
✅ [Experimental] Expo Router for navigation? (y/N) · Yes

I'm not here to say you need to support both demo apps, I totally get that is a huge huge ask. I'm more just wondering if there's a way to be more clear "If you opt into this, you opt out of that"

Maybe ask about expo in the CLI first, and have a small thing saying "incompatible with demo code" and only ask about demo code if they opt out of expo router? Perhaps something like:

✅ Which package manager do you want to use? · pnpm
✅ Do you want to install dependencies? (Y/n) · Yes
✅ [Experimental] Expo Router for navigation? (This removes demo code) (y/N) · no
✅ Remove demo code? We recommend leaving it in if it's your first time using Ignite (y/N) · No

And if Expo is chosen, just skip the 4th step.

It is worth mentioning that in the build process it does explicitly point out 

    🛠️ Removing fancy demo markup

But there doesn't appear to be any clear reason as to why.

frankcalise:
## Description

- Closes #2916 
- Moves the `Remove demo code` prompt after nav and state mgmt questions which can force it's removal
- Warns the user better about demo code being removed
- Updates docs to mention this as well

### Added notes in the prompts
image

### If expo-router and/or MST removal, prints a warning instead of prompting `Remove demo code`
image

## Checklist

### DIFF
--- a/docs/cli/Ignite-CLI.md
+++ b/docs/cli/Ignite-CLI.md
@@ -98,7 +98,7 @@ Starts the interactive prompt for generating a new Ignite project. Any options n
 - `--overwrite` overwrite the target directory if it exists
 - `--targetPath` string, specify a target directory where the project should be created
 - `--removeDemo` will remove the boilerplate demo code after project creation
-- `--state` string, one of `mst` or `none` to include MobX-State-Tree in project (can only be set to `none` if `--removeDemo=true`)
+- `--state` string, one of `mst` or `none` to include MobX-State-Tree in project (**note:** if opting out of MobX-State-Tree the demo application will be removed)
 - `--useCache` flag specifying to use dependency cache for quicker installs
 - `--no-timeout` flag to disable the timeout protection (useful for slow internet connections)
 - `--yes` accept all prompt defaults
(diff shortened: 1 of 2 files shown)


coolsoftwaretyler (approved):
LGTM!

jamonholmgren:
I like the improvements

Waffleophagus:
This looks fantastic! I think this would’ve completely cleared up my confusion when getting started and hopefully will save others! I definitely think this satisfies my issue I opened. https://github.com/infinitered/ignite/issues/2916

silasjmatson (approved):

PR #2924, opened 2025-04-05
fix(boilerplate): updated i18n translate

10xDevFound:
updated the doc for i18n translate to reflect proper usage of the translation keys

- Its a quick small change to rename the example key which is present in the translate.ts file for i18n translations.
- I am starting out the development of my own RN project and came across the ignite project and was exploring the project whenI found this tiny issue.


shishirchulliyil (approved):

markrickert:
Good catch! thanks!

PR #2915, opened 2025-03-19
fix(boilerplate) Fix Icon Type definitions

silasjmatson:
## Description

There were two issues at play here:

1. Icon's `style` prop overrides the Wrapped `TouchableOpacity`'s `style` prop with an incompatible `StyleProp<ImageStyle>`
2. Icon changed the wrapper used based on the presence of the `onPress` prop, which made accurate typing and linting difficult.

See
https://github.com/infinitered/ignite/issues/2888#issuecomment-2738096598 for more information.

This commit fixes #2888 by simply separating `Icon` into `PressableIcon` and `Icon` components, with simplified prop types for each, and updates the docs to reflect the change.

- Issues:  #2888 

## Checklist

- [x] `README.md` and other relevant documentation has been updated with my changes

PR #2930, opened 2025-04-15
improve logic for other `onPress` props

ChristopherGabba:
## Description

Ran into an issue where I wanted to use `onPressIn` and nothing was firing for the Icon button. I updated the Icon component to include all the pressability props to make sure the correct wrapper is used.

Should be a very minor change, so no real need for full on testing, but I'll let you guys double check my work!

Thanks again,
Chris


frankcalise:
Thanks for the contribution, however I believe this is covered by #2915

ChristopherGabba:
Fair enough, that's another solution! Thanks...

PR #2931, opened 2025-04-23
feat(boilerplate): e2e support for dev client or standalone builds

frankcalise:
## Description

Improves the maestro e2e workflow for the boilerplate by having flows for both development and preview/production applications.

### How it works

Each maestro test flow is meant to prepend the `onFlowStart` hook (see `FavoritePodcast.yaml`). This will help handle the Expo dev client launch screen when running in dev with metro or just boot the application normally if it was installed outside of dev. This is currently handled in the `package.json` script on the `test:maestro*` commands via their CLI `-e`, but could be extended to something else by the consumer, such as different bundle IDs, etc.

`Login.yaml` subflow has `clearState` and `launchApp` removed as it will be taken care of by the `onFlowStart` hook all the time.

### What else was added
- package.json scripts for
  - `test:maestro:ci` for when the consumer needs to specifiy a different env var
  - eas preview builds for sims
- `-e IS_DEV=true` is used for now because we don't provide a set up where we have multiple bundle IDs out of the box but this could be extended or documented further
- The podcast API url was added to `config.prod.ts` because when I built the preview build app, the e2e test failed and couldn't find any podcasts to favorite 😆 

## Screenshots

| Dev                           | Preview                           |
| image | image |

## Checklist


test/vanilla/__snapshots__/ignite-remove-demo.test.ts.snap
frankcalise, line 1: I want to say this is correct, because we'd remove Login/FavoritePodcast flows since they don't apply if you remove the demo app, but now keeping the _helper files so the directory is no longer removed

boilerplate/package.json
```diff
@@ -10,14 +10,17 @@
     "patch": "patch-package",
     "test": "jest",
     "test:watch": "jest --watch",
-    "test:maestro": "maestro test .maestro/FavoritePodcast.yaml",
+    "test:maestro": "maestro test -e MAESTRO_APP_ID=com.helloworld -e IS_DEV=true .maestro/FavoritePodcast.yaml",
+    "test:maestro:ci": "maestro test -e MAESTRO_APP_ID=com.helloworld -e IS_DEV=false .maestro/FavoritePodcast.yaml",
```
frankcalise, line 14: These should probably be more generic than running the exact file?

coolsoftwaretyler: We'll also want to set up a couple sub-folders I think. Let me test it out.

boilerplate/.maestro/_OnFlowStart.yaml
```diff
@@ -0,0 +1,36 @@
+# flow: run the login flow and then navigate to the demo podcast list screen, favorite a podcast, and then switch the list to only be favorites.
+
+appId: ${MAESTRO_APP_ID}
```
coolsoftwaretyler, line 3: **question:** did you copy over `FavoritePodcast` into `_OnFlowStart` by accident? This looks like a duplicate and not what I would expect here.

test/vanilla/__snapshots__/ignite-remove-demo.test.ts.snap
coolsoftwaretyler: Oh I think maybe I will break this with my latest commit - I put the demo removal lines in *all* the Maestro flows. Should we leave some?

frankcalise: They only need to go in demo app related flows, so anything login screen or fav podcast related

It doesn't need to be in the shared hook, dev client etc

coolsoftwaretyler (approved):
LGTM + pitched in a couple of commits. Thanks for doing this, man! I think it'll be very useful for the community.

Issue #2900, opened 2025-02-07
Back button on drawer hides automatically


wisammechano:
### Describe the bug

I am trying the demo app in Ignite version 10.1.6, and when I click the hamburger button, the drawer opens with the back button that disappears after the animation ends. This happened only on the simulator (iPhone 15 Pro Max, iOS 17.4), not on the web app.

command: `yarn ios`

screen record: 

https://github.com/user-attachments/assets/eb53e5ff-cd40-47f3-8de8-c11f8d5567c2

### Ignite version

10.1.6

### Additional info

`npx ignite-cli@latest doctor`

```
System
  platform           darwin                                                           
  arch               arm64                                                            
  cpu                8 cores             Apple M1 Pro                                 
  directory          axeer-mobile-demo   /Users/wisam/Projects/nova/axeer-mobile-demo 

JavaScript (and globally-installed packages)
  node                     20.18.0                                   /Users/wisam/Library/Application Support/Herd/config/nvm/versions/node/v20.18.0/bin/node 
  npm                      10.8.2                                    /Users/wisam/Library/Application Support/Herd/config/nvm/versions/node/v20.18.0/bin/npm  
    bcrypt-cli             1.1.0                                                                                                                              
    corepack               0.29.3                                                                                                                             
    npm                    10.8.2                                                                                                                             
  yarn                     1.22.19                                   /usr/local/bin/yarn                                                                      
    create-react-app       5.0.1                                                                                                                              
    create-react-library   3.1.1                                                                                                                              
  pnpm                     9.0.6                                     /usr/local/bin/pnpm                                                                      
                           <no pnpm global package info available>                                                                                            
  bun                      -                                         not installed                                                                            
  expo                     52.0.30                                   managed                                                                                  

Ignite
  ignite-cli         10.1.6       /Users/wisam/.npm/_npx/bec2839915ba8fc3/node_modules/.bin/ignite      
  ignite src         build        /Users/wisam/.npm/_npx/bec2839915ba8fc3/node_modules/ignite-cli/build 

Android
  java               23.0.1       /usr/bin/java 
  android home       -            undefined     

iOS
  xcode              16.2       
  cocoapods          1.15.2       /Users/wisam/.rvm/gems/ruby-3.3.1/bin/pod 

Tools
  git                git version 2.39.5 (Apple Git-154)   /usr/bin/git 
```

warmfire540:
[...]

PR #2938, opened 2025-05-03
[draft] feat: upgrade to Expo SDK 53

fpena:
## Description

- Issues: 

## Screenshots

| Before                           | After                           |
| [Insert before screenshot/video] | [Insert after screenshot/video] |

## Checklist

- [ ] `README.md` and other relevant documentation has been updated with my changes
- [ ] I have manually tested this, including by generating a new app locally ([see docs](https://docs.infinite.red/ignite-cli/contributing/Contributing-To-Ignite/#testing-changes-from-your-local-copy-of-ignite)).


src/commands/new.ts
```diff
@@ -472,7 +472,7 @@ module.exports = {
     }
 
     // New Architecture
-    const defaultNewArch = false
+    const defaultNewArch = true
```
fpena, line 475: Enabling new architecture by default with Expo 53.

markrickert:
Moved to ready for review.

markrickert:
I noticed that the Animated component we're using in the drawer to display the back arrow is animating in and then disappearing on the iOS ... anyone have ideas on what might be causing this?

What's SUUUPER weird is that the recording feature from the ios simulator shows that it's always there, but if i screencapture, there's clearly an issue:

| Sim Recorder | Screen Recorder |
|--------|--------|
| Simulator Screen Recording - iPhone 16 Pro - 2025-05-14 at 09 44 54 | F1A51929-93B2-4A09-9E34-5A1176CCFA8E-11904-00015ECEC3C101E1 |

frankcalise:
@markrickert that is a pre-existing issue, https://github.com/infinitered/ignite/issues/2900 - I did some digging on it but wasn't able to solve it. I'll see if I can find my notes on it. Ultimately, I know @lindboe was doing some exploration on replacing the drawer since it had accessibility problems plus some of these quirks

boilerplate/app/components/Button.tsx
```diff
@@ -239,7 +239,7 @@ const $pressedViewPresets: Record<Presets, ThemedStyle<ViewStyle>> = {
   reversed: ({ colors }) => ({ backgroundColor: colors.palette.neutral700 }),
 }
 
-const $pressedTextPresets: Record<Presets, ThemedStyle<ViewStyle>> = {
+const $pressedTextPresets: Record<Presets, ThemedStyle<TextStyle>> = {
```
frankcalise, line 242: Overlap with #2952

test/vanilla/__snapshots__/ignite-remove-demo.test.ts.snap
```diff
@@ -6,6 +6,7 @@ exports[`ignite-cli remove-demo should print the expected response 1`] = `
 removing file /user/home/ignite/.maestro/flows/FavoritePodcast.yaml
 removing file /user/home/ignite/.maestro/flows/Login.yaml
 removing file /user/home/ignite/.maestro/shared/_Login.yaml
+removing file /user/home/ignite/.maestro/shared/_OnFlowStart.yaml
```
frankcalise, line 9: Will need to update this snap if we decide to keep this file with my comment on it above

test/vanilla/ignite-generate.test.ts
```diff
@@ -91,15 +91,15 @@ test(\\"can be created\\", () => {
...
 
-        // @mst remove-file
-        export * from \\"./Pizza\\"
-        "
-      `)
+// @mst remove-file
+export * from "./Pizza"
+"
+`)
       expect(read(`${TEMP_DIR}/app/models/RootStore.ts`)).toEqual(
```
frankcalise, line 103: I'm also wondering if the eslint config change will have impact on these tests - but CI is green so seems fine

frankcalise (commented):
Looking good! Left a few comments. The one requiring most discussion is the removal of `_OnFlowStart.yaml` I think

boilerplate/.maestro/shared/_OnFlowStart.yaml
```diff
@@ -1,13 +1,59 @@
...
+    when:
+      platform: iOS
+      visible: 
+        id: "SBSwitcherWindow:Main" # id of the iOS bottom sheet we want to dismiss
+    commands:
+      - swipe:
+          direction: DOWN
+      - waitForAnimationToEnd
+
+# @demo remove-file
```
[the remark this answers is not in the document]
markrickert: Yeah, i agree that it would be useful. I'll get rid of that directive to remove the file

src/tools/__snapshots__/markup.test.ts.snap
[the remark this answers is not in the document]
markrickert: @frankcalise Yeah, i think this was fixed when i inherit the eslint config... helps make working on the ignite repo a little easier.

When i did a save-and-format on the file with all the double backslashes in it, it just took them all out. and running `yarn test -u` updates the snapshot properly and passes, so i think we're good here.

src/commands/new.ts
```diff
@@ -180,7 +180,7 @@ module.exports = {
 
     const yname = boolFlag(options.y) || boolFlag(options.yes)
     const noTimeout = options.noTimeout ?? false
-    const useDefault = (option: unknown) => yname && option === undefined
```
markrickert, line 183: The linter was complaining about the name of this function and that it should be a react hook because it begins with "use...". Renamed to `getDefault`.

markrickert (approved):

frankcalise (approved):

markrickert:
Rebased and fixed conflicts

markrickert:
Another rebase and i think this one is ready to roll. I've tested it pretty thoroughly and ran maestro tests in both new arch and non-new arch ignited apps and they are all passing.

Gonna go ahead and merge this one!

frankcalise:
@markrickert sounds good!

markrickert:
Added a few more commits. @frankcalise could you re-review please?

Additional changes since you last looked:

1. Followed the expo-fyi docs for moving the app.json `expo:{}` object to the root level. https://github.com/expo/fyi/blob/main/root-expo-object.md
2. Fixed the scripts to reference the root of the app.json file instead of the expo object and added the ignite version indicator to the [`"extra"` field](https://docs.expo.dev/versions/latest/config/app/#extra) for arbitrary values like this.
3. Fix the expo-router option that had a hard-coded `expo-localization` version. There was an update to this package so i thought it more future-proof to parse the json string and get the actual version we're already using.
4. I realized that some of the packages needed to be updated to their latest versions, so i spend a lot of time igniting apps and running `yarn upgrade-interactive --latest` and `npx expo install --fix` until i was able to get all the dependencies updated and working in the app.
5. During the process of 4 (above) i saw on the expo jest docs that we should remove the `react-test-renderer` library so i did. Tests all still pass in the main app and in an ignited app.

jamonholmgren:
Should we really even have the old arch option at all? Given the deprecation.

frankcalise:
> Should we really even have the old arch option at all? Given the deprecation.

@jamonholmgren true we could simplify the cli maintenance and remove it

frankcalise (approved):
Still looking good to me. If we want to opt out of letting users go to legacy arch, then that would help remove the prompt in `new.ts` and also the mmkv specific version code (v2 vs v3)

jamonholmgren:
> > Should we really even have the old arch option at all? Given the deprecation.
> 
> @jamonholmgren true we could simplify the cli maintenance and remove it

Yeah, and they can use the previous version of Ignite to get the old arch.

markrickert:
@jamonholmgren I'm going to merge this and then open a new PR to strip out the old arch option. I'll make a note in the readme and the docs about the command to use if you HAVE to have the legacy arch in an ignited app.

When we merge THAT one, i think we should bump the version with `BREAKING CHANGE`

boilerplate/app/screens/LoginScreen.tsx
```diff
@@ -11,23 +10,21 @@ import {
...
-import { useStores } from "@/models"
 import { AppStackScreenProps } from "@/navigators"
 import type { ThemedStyle } from "@/theme"
 import { useAppTheme } from "@/utils/useAppTheme"
+import { useAuth } from "@/context/AuthContext"
 
 interface LoginScreenProps extends AppStackScreenProps<"Login"> {}
 
-export const LoginScreen: FC<LoginScreenProps> = observer(function LoginScreen(_props) {
+export const LoginScreen: FC<LoginScreenProps> = (_props) => {
```
coolsoftwaretyler, line 20:
```suggestion
export const LoginScreen: FC<LoginScreenProps> = () => {
```

**suggestion:** if we aren't using these, should we just remove them?

boilerplate/package.json
```diff
@@ -29,70 +28,72 @@
...
   },
   "engines": {
     "node": "^18.18.0 || >=20.0.0"
+  },
+  "expo": {
+    "install": {
+      "exclude": [
+        "@shopify/flash-list"
+      ]
+    }
```
coolsoftwaretyler, line 97: **question:** why do we need this? And should we document it anywhere?

boilerplate/package.json
```diff
@@ -29,70 +28,72 @@
...
     "reactotron-react-js": "^3.3.11",
     "reactotron-react-native": "^5.0.5",
     "reactotron-react-native-mmkv": "^0.2.6",
     "ts-jest": "^29.1.1",
     "ts-node": "^10.9.2",
-    "typescript": "~5.3.3"
+    "typescript": "~5.8.3"
   },
   "engines": {
     "node": "^18.18.0 || >=20.0.0"
```
coolsoftwaretyler, line 90: **suggestion:** I think we should remove Node 18 since it's no longer [maintained](https://nodejs.org/en/about/previous-releases), but we can do that in a separate PR.

coolsoftwaretyler (approved):
@fpena @markrickert - this is great. I only had about 30-40 minutes today to take a look, so I left some non-blocking comments while I spun it up locally. I checked out newly Ignited app on Android and iOS with the default settings and tapped around for a smoke test. Looks great.

boilerplate/.maestro/flows/FavoritePodcast.yaml
```diff
@@ -2,7 +2,7 @@
 
 appId: ${MAESTRO_APP_ID}
 env:
-  FAVORITES_TEXT: "Switch on to only show favorites"
+  FAVORITES_TEXT: "Switch on to only show favorites" # en.demoPodcastListScreen.accessibility.switch
```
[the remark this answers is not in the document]
markrickert: Yeah... if someone runs the maestro tests in another language, the test will fail, so i put it here just as a reference

markrickert:
@no-1ne I've updated the main comment in this PR for better visibility (adding PR links and stuff) so that you can better track progress instead of a commend half way down this page.

I'm hoping that everything on that list will be done and tested in the next few weeks. React 19 ref passing as props might come shortly after the release of v11. It's not as high on the priority list as the theming updates.

Thank you for following along with us!

markrickert:
OK, i think this one is ready to go.

PR #2939, opened 2025-05-04
feat: better type suggestions in useNavigate hook

supriyoMondal:
## Description

 The useNavigate hook from react navigation by default not coming with any route name suggestions


frankcalise:
hi @supriyoMondal - you can type it based on the way your navigators are set up, for example:

```tsx
import { useNavigation } from "@react-navigation/native"
import { NativeStackNavigationProp } from "@react-navigation/native-stack"
import { AppStackParamList } from "../navigators"

export function useAppNavigation() {
  const navigation = useNavigation<NativeStackNavigationProp<AppStackParamList>>()

  return navigation
}
```

PR #2944, opened 2025-05-10
Fixed Broken Link in Documentation

Zammer8:
Edited the component library link in `/docs/README.md` to fix an issue where it linked to `/docs/docs`, an invalid URL. This results in `ignite/docs/README.md` needing to differ from `ignite/README.md`, however the link now works as expected. It could be changed to an absolute link to keep the two README files the same if that's needed.


markrickert:
Thanks for surfacing this and for the contribution! I think i found another broken link so i'll take care of that!

PR #2948, opened 2025-05-13
expo53: configuration, dependency, linting, and maestro fixes

markrickert:
## Description

This makes a few changes to @fpena's expo 53 branch in #2938:

1. Updates to the tsconfig. All these changes bring our config in line with [the expo sdk 53 config](https://github.com/expo/expo/blob/main/packages/expo/tsconfig.base.json).
  * `moduleResolution` changed from `'node` to `bundler`
  * `module` changed from `es2015` to `preserve`
  * Added `"customConditions": ["react-native"]` to the compiler options.
2. These updates to the tsconfig revealed some issues that weren't caught by the `yarn compile` command previously. Mostly just accidentally substituting `ViewStyle` for `TextStyle` in the type declarations.
3. Fixes the maestro tests so the same flow works for all permutations of `expo/non-expo + newArch/oldArch`. ... simplifies the number of files/scripts needed, and reduces the amount of logic needed to bypass both the dev server picker screen and the dev menu bottom sheet from expo that shows up in expo dev builds in the oldArch.
4. updates some packages to the correct versions that `npx expo install --fix` told me to update including the `@types/react` and `expo-router` deps.
5. During the course of updating packages, i moved the eslint config for the main ignite package to a javascript file and made it inherit from the boilerplate's config so that our linter is consistent throughout. 

## Checklist


src/commands/new.ts
```diff
@@ -180,7 +180,7 @@ module.exports = {
 
     const yname = boolFlag(options.y) || boolFlag(options.yes)
     const noTimeout = options.noTimeout ?? false
-    const useDefault = (option: unknown) => yname && option === undefined
```
markrickert, line 183: The new linter config was complaining about "use" being a react-only keyword so i changed it to "get"

boilerplate/app/screens/DemoShowroomScreen/DemoShowroomScreen.tsx
```diff
@@ -166,7 +168,11 @@ export const DemoShowroomScreen: FC<DemoTabScreenProps<"DemoShowroom">> =
     }
 
     useEffect(() => {
-      return () => timeout.current && clearTimeout(timeout.current)
```
markrickert, line 169: the linter was complaining about this possibly returning `null` 🤦

fpena (approved):
Looks good! Thanks for the changes.

PR #2961, opened 2025-05-30
feat: remove the ability to opt out of the new architecture

markrickert:
## Description

This PR removes the ability to opt OUT of the new architecture. Every ignited app will have `newArchEnabled: true` from now on. 

## Screenshots

The prompts no longer ask you if you want to enable the new arch or not, but if you pass `--new-arch=false` you'll get a warning and the process will exit:

Screenshot 2025-05-29 at 6 11 45 PM

## Checklist


frankcalise (approved):
Nice work! Less to maintain and pushes the community in the right direction.

Also nice idea to ditch some of those deprecated flags. 🎉

PR #2962, opened 2025-05-30
feat: remove all barrel files

markrickert:
## Description

**Ignite v11 Feature: no more barrel files**

This PR removes all `index.ts{x}` barrel export files from the application and reorganizes the imports to import directly from the source files instead of a barrel file.

After some internal discussion and reading up on the matter, we've come to the consensus that barrel files are generally bad, though I've left one in the demo code for all the demos to be imports and iterated through.

Articles that support this and indicate that while barrel files are *A* solution, they shouldn't be the *GO-TO* solution:

- https://marvinh.dev/blog/speeding-up-javascript-ecosystem-part-7/
- https://uglow.medium.com/burn-the-barrel-c282578f21b6
- https://tkdodo.eu/blog/please-stop-using-barrel-files

Barrel files made sense back in the day of manual imports, but almost everyone nowadays uses a code editor that can auto-import files as you need them and can automatically remove imports that aren't used anywhere in the file.

Along with removing the barrel files in the boilerplate, i had to modify the generators to no longer add generated files to their respective folder's barrel file.

> But  wait! There's more!

While i was mucking about with the imports sections of the boilerplate files, i implemented some more complex rules in the eslint for `import/order`. 

Import orders will now automatically be sorted by prettier in the following way:

1. react and react-native at the top followed by all expo related libraries
2. all other external dependencies
3. new line
4. all the `@/*` imports
5. new line
6. then all the relative imports first `../*` then `./*` with no newline between them.
7. require statements

Oh yeah, and it sorts them alphabetically now too.

I'm a *BIG FAN* 🪭  of strict rules for import sorting and auto-management of imports using eslint and prettier.

## Checklist


coolsoftwaretyler:
Haven't had a chance to look yet, but wanted to chime in that from an Ignite philosophy standpoint: the projects I've been on so far at Infinite Red usually include those import order rules. It's a good fit as an official IR recommendation.

jamonholmgren (approved):
I like this a lot. No changes needed that I can see.

PR #2968, opened 2025-06-14
Fix: useRef initial value

objectiveSee:
This fixes a small warning where the ref was not given an initial value.


markrickert:
Thanks for the PR @objectiveSee! I had noticed this issue and fixed it in the upcoming version 11 (#2938). We're going to be releasing that real soon, so I'm going to merge this with `[skip ci]` so you still get the credit for your contribution. 

Thanks again!

PR #2982, opened 2025-08-06
feat(boilerplate,docs): Remove FlashList

### LINKED ISSUE #2972: Remove FlashList
## Description

Version 1.x has given us some headaches. Maybe we'll re-evaluate v2 in the near future as we use it on more projects. But since we are having to reduce the number of uses on our client projects with FlashList one (in favor of FlatList or Legend List), let's remove the dependency from the boilerplate.

## How

1. Remove `@shopify/flash-list` from dependencies
2. Remove `expo.install.exclude` key from `package.json
3. Remove `app/components/ListView.tsx`
4. Replace any boilerplate `<ListView />` components with `<FlatList />` for now

SeanBarker182:
## Description

Removes FlashList dependency and replaces with FlatList

One thing worth mentioning is that DemoListItem is using FlatList from React Native Gesture Handler. This is to support nested scrollview since the FlatList in this file is nested inside the SectionListWithKeyboardAwareScrollView inside of DemoShowroomScreen. All other instances of FlatList import from React Native. 

- Issues: Resolves #2972 

## Checklist


frankcalise (approved):
Nice work! LGTM 👍 

Will let @markrickert look it over as well

Issue #2983, opened 2025-08-11
npm run android failed with gradlew.bat app:assembleDebug -x lint -x test --configure-on-demand --build-cache -PreactNativeDevServerPort=8081 -PreactNativeArchitectures=x86_64,arm64-v8a exited with non-zero code: 1


devimkarthika:
### Describe the bug

using 
java 17
node 20

### Ignite version

11.1.0

### Additional info

 platform           win32
  arch               x64
Node.js v20.19.4

chenp1204:
Seems this issue can resolve your problem...
https://github.com/expo/expo/issues/36274

joshuayoes:
So I ran into this recently as well :(

I had the long path enabled but I needed to update my ninja.exe like in this issue https://github.com/expo/expo/issues/36274#issuecomment-3242404346

PR #2985, opened 2025-08-14
fix(cli): run project path validation checks

SeanBarker182:
## Description

This update introduces a function to warn users about long project paths on Windows, which can lead to failures in Android CMake/Ninja builds due to MAX_PATH limitations.

- Issues: #2983 

## Checklist

### DIFF
--- a/src/tools/validations.ts
+++ b/src/tools/validations.ts
@@ -125,7 +125,7 @@ export function validateProjectPath(absPath: string, toolbox: GluegunToolbox) {
         `Windows project path is quite long (${len} chars). Android native builds can fail on very long paths.`,
       )
       print.warning(`Path: ${normalized}`)
-      print.warning(`Tip: move your project closer to the drive root, e.g. ${"C:\\\\src\\\\MyApp"}`)
+      print.warning(`Tip: move your project closer to the drive root, e.g. ${"C:\\src\\MyApp"}`)
       print.warning(
         `Why: CMake/Ninja can generate very long object file paths that hit Windows limits.`,
       )


src/commands/new.ts
```diff
@@ -1059,3 +1064,30 @@ export function findAndRemoveDependencies(
...
+
+// Threshold for warning about long paths on Windows
+const PATH_WARN_THRESHOLD = 120
+
+/**
+ * Warn if absolute project path is "deep" on Windows.
+ * Deep paths can cause Android CMake/Ninja builds to fail due to MAX_PATH issues.
+ */
+function checkWindowsPathLength(absPath: string, toolbox: GluegunToolbox): void {
+  if (process.platform !== "win32") return
```
frankcalise, line 1076: Is this just a safeguard against someone else calling it? You have the `if (isWindows)` check over in new, so maybe just leave this one and ditch the other?

frankcalise (commented):
I like it! Couple of thoughts/suggestions I can think of now that we went down this rabbit hole:

1. Move `checkWindowsPathLength` into `src/tools/validation.ts` since it seems to fit there
2. Consider renaming `checkWindowsPathLength` to `validateProjectPath` and we could have platform specific checks in there?
  - For example, Windows needs the length check
  - Honestly, macOS needs a check for spaces in the path because Xcode will die on build if spaces are in the directory I'm pretty sure

src/commands/new.ts
SeanBarker182: Yeah I figured it didn't hurt to double check just to be safe, but I think it makes sense to just make this a generic method and do the check in this function so we can add other platforms in the future if needed.

Issue #2987, opened 2025-08-25
Dependency conflict in the initial boilerplate


iceberg53:
### Describe the bug

I tried to spin up a new app with the framework and I got the following error suggesting that there might be an issue with the project's dependencies.

Here's the command I ran : 

    npx ignite-cli@latest new MyApp

And the error message I got : 

    The following error occurred:

    Error: Command failed: npm install
    npm error code ERESOLVE
    npm error ERESOLVE unable to resolve dependency tree
    npm error
    npm error While resolving: my-app@0.0.1
    npm error Found: i18next@23.16.8
    npm error node_modules/i18next
    npm error   i18next@"^23.14.0" from the root project
    npm error
    npm error Could not resolve dependency:
    npm error peer i18next@">= 25.4.1" from react-i18next@15.7.2
    npm error node_modules/react-i18next
    npm error   react-i18next@"^15.0.1" from the root project
    npm error
    npm error Fix the upstream dependency conflict, or retry
    npm error this command with --force or --legacy-peer-deps
    npm error to accept an incorrect (and potentially broken) dependency resolution.
    npm error
    npm error
    npm error For a full report see:
    npm error C:\Users\pseudo\AppData\Local\npm-cache\_logs\2025-08-25T10_01_34_527Z-eresolve-report.txt
    npm error A complete log of this run can be found in: C:\Users\pseudo\AppData\Local\npm-cache\_logs\2025-            08-25T10_01_34_527Z-debug-0.log

     Consider opening an issue with the following information at:
        https://github.com/infinitered/ignite/issues/new?template=bug_report.yml&labels=bug

        /  Gathering system and project details

    System
      platform           win32
      arch               x64
      cpu                8 cores          Intel(R) Core(TM) i5-8365U CPU @ 1.60GHz
      directory          MyApp   C:\Apps\MyApp
    JavaScript (and globally-installed packages)
      node               22.14.0      C:\laragon\bin\nodejs\node-v22.14.0-win-x64\node.EXE
      npm                10.9.2       C:\laragon\bin\nodejs\node-v22.14.0-win-x64\npm.CMD
        corepack         0.31.0
        lighthouse       12.8.0
        npm              10.9.2
        site-audit-seo   6.0.1
        ts-node          10.9.2
        tsx              4.20.3
      yarn               1.22.19      C:\laragon\bin\yarn\bin\yarn.CMD
      pnpm               -            not installed
      bun                -            not installed

    Ignite
[...]

PR #2988, opened 2025-08-26
fix(boilerplate): i18next dep versions

frankcalise:
## Description

Closes issue #2987 reported by the community. This occurred because of the caret range specifier on `reat-i18next`. They released a patch version which requires a major version bump in it's peer dependency `i18next`. 

The caret range won't jump that high for `i18next`, therefore installation would fail.

This fixes that issue by removing the caret range for these packages, but also upgrades them to the latest we were attempting to install.


coolsoftwaretyler (approved):
LGTM thanks!

PR #3002, opened 2025-10-06
Feat/add devclient cli command

coolsoftwaretyler:
## Description

This PR adds a new CLI command, `devclient`, to download pre-built development builds and launch from the command line.

### TODOs

- [ ] Update docs
- [ ] Figure out what we want to do for metrics

### Works when it finds a matching version

https://github.com/user-attachments/assets/476672bf-583c-43f2-82d5-9028e8e4ad9f

### Works when no matching version found

https://github.com/user-attachments/assets/3c97ffa7-cccf-4f4e-b85a-986d7453621d

## Checklist

- [ ] `README.md` and other relevant documentation has been updated with my changes

### DIFF
--- a/src/commands/devclient.ts
+++ b/src/commands/devclient.ts
@@ -1,0 +1,454 @@
+import { GluegunToolbox } from "gluegun"
+import * as os from "os"
+import * as path from "path"
+
+import { p, heading, warning, startSpinner, stopSpinner, clearSpinners } from "../tools/pretty"
+
+interface GitHubAsset {
+  name: string
+  browser_download_url: string
+}
+
+interface GitHubRelease {
+  tag_name: string
+  assets: GitHubAsset[]
+}
+
+type BuildType = "ios-simulator" | "android-emulator" | "android-device"
+
+const isMac = process.platform === "darwin"
+
+module.exports = {
+  description: "Download pre-built Expo development clients from GitHub releases",
+  run: async function (toolbox: GluegunToolbox) {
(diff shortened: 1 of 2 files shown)


src/commands/help.ts
```diff
@@ -62,6 +62,12 @@ module.exports = {
       ["npx ignite-cli remove-demo-markup", "npx ignite-cli remove-demo-markup --dry-run"],
     )
     p()
+    command(
+      "devclient        ",
```
coolsoftwaretyler, line 66: **note:** I think they're technically called "Development builds". The "devclient" is [actually this thing](https://docs.expo.dev/versions/latest/sdk/dev-client/). Maybe we should rename this before merging to get the terminology correct.

src/commands/devclient.ts
```diff
@@ -0,0 +1,454 @@
...
+  assets: GitHubAsset[]
+}
+
+type BuildType = "ios-simulator" | "android-emulator" | "android-device"
+
+const isMac = process.platform === "darwin"
+
+module.exports = {
+  description: "Download pre-built Expo development clients from GitHub releases",
+  run: async function (toolbox: GluegunToolbox) {
```
coolsoftwaretyler, line 23: **note:** we should figure out what we want to track for usage, and where we want to track it. there are a few paths here (did they find a matching download, if so, which one, if not which one did they pick, and then which of the three types of downloads did they choose).

I think we probably want to start tracking before we merge so we don't miss any data, but I'm open to merging and figuring that out later if we're OK with dropping some insights early on.

Issue #2995, opened 2025-09-18
Remove `useFonts` and embed them as part of the App Binary


ChristopherGabba:
Good morning Infinite Red,

I came across this X.com thread: https://x.com/aleqsio/status/1968273285239734405 and I read through the expo font docs: https://docs.expo.dev/versions/latest/sdk/font/

and they talk about the preference for fonts is to not use `useFonts` and load them asynchronously but to include them in the app binary using the `expo-config-plugin`. I just finished getting this working in my app, and here are all the diffs, which should be similar to ignite, just with different fonts:

```tsx
// App.tsx

```

```json

```

jamonholmgren:
Looks pretty sweet to me. What about Expo Web?

ChristopherGabba:
That's a solid question, didn't think about web. I'll look into that as well.

PR #3003, opened 2025-10-06
Feat/precompile fonts

ChristopherGabba:
## Description

I came across this X.com thread: https://x.com/aleqsio/status/1968273285239734405 and I read through the expo font docs: https://docs.expo.dev/versions/latest/sdk/font/

Based off everything I've read, there is no way to do this on web so I tried to account for that on the PR. Maybe someone from expo can shed some light on this, but the docs don't mention web on the config plugin. 

- Issues: #2995

## Screenshots

It appears to be working on all three platforms.

Android:
Screenshot 2025-10-06 at 2 04 28 PM

iOS:

IMG_0293

Web:

Screenshot 2025-10-06 at 2 47 02 PM

## Checklist


boilerplate/app/theme/typography.ts
```diff
@@ -2,30 +2,64 @@
...
+ * using the expo config plugin in `app.json`. See the project
+ * [`app.json`](../../app.json) for the expo-fonts configuration. The assets
+ * are added via the `app/assets/fonts` folder. This config plugin
+ * does NOT work for web, so we have to dynamically load the fonts via this hook.
+ *
+ * For more info: https://docs.expo.dev/versions/latest/sdk/font/
+ */
+export const useCustomFonts = (): [boolean, Error | null] => {
+  if (Platform.OS === "web") {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
```
coolsoftwaretyler: **suggestion:** this is technically safe since `Platform.OS` is never really going to change at runtime and trigger the problem this rule prevents, but I'd still prefer to comply to the rules of hooks. I think we can get around it by always calling `useFonts`, and then conditionally returning different values. Like this:

```patch
diff --git a/app/theme/typography.ts b/app/theme/typography.ts
index 65f7aff..a78dbb1 100644
--- a/app/theme/typography.ts
+++ b/app/theme/typography.ts
@@ -24,9 +24,10 @@ export const customFontsToLoad = FONT_FILES as Record<string, FontSource>
  * For more info: https://docs.expo.dev/versions/latest/sdk/font/
  */
 export const useCustomFonts = (): [boolean, Error | null] => {
+  const [areFontsLoaded, fontLoadError] = useFonts(customFontsToLoad)
+
   if (Platform.OS === "web") {
-    // eslint-disable-next-line react-hooks/rules-of-hooks
-    return useFonts(customFontsToLoad)
+    return [areFontsLoaded, fontLoadError]
   }
 
   // On native, fonts are precompiled and ready
```

coolsoftwaretyler:
Looks great, @ChristopherGabba - will you please rebase and fix the conflict? I think we're good after that.

ChristopherGabba:
Did I do it? haha never rebased before.

coolsoftwaretyler:
Looks like you merged instead of rebasing ([here's the difference](https://www.atlassian.com/git/tutorials/merging-vs-rebasing)). That's OK though, we don't have a super active main branch, and if you resolved conflicts through the GitHub UI, that's what they usually do by default anyway.

ChristopherGabba:
Awesome! I've been experimenting with different fonts on import and there's definitely some insistency between different fonts. There may be some fonts where we need to use Platform.select still. Montserrat is one of those where on iOS, the weight is not in the font name. 

Can you run an iOS build on your end and make sure it's correctly working?

coolsoftwaretyler:
Hey @ChristopherGabba - can you say more about the inconsistency you're seeing? Maybe post a couple screenshots? I don't think we use Montserrat here.

ChristopherGabba:
@coolsoftwaretyler No, I'm just using Monserrat in my personal app, but I have to differentiate between the iOS and android platforms for that particular font. 

Screenshot 2025-10-16 at 10 29 01 AM

Note how I'm importing them all with numbers, but the way to reference them still requires me to differentiate.

Screenshot 2025-10-16 at 10 29 32 AM

For some reason, SpaceGrotesk with numbers just works for both without `Platform.select`. I don't fully understand why because when I go to the info.plist for my app, they show up as numbers. But they literally do not render on iOS unless I use the `Platform.select`..

Screenshot 2025-10-16 at 10 31 11 AM

boilerplate/app.config.ts
```diff
@@ -16,9 +32,11 @@ import "tsx/cjs"
  */
 module.exports = ({ config }: ConfigContext): Partial<ExpoConfig> => {
   const existingPlugins = config.plugins ?? []
+  const fontPaths = getFontPathsFromAssetsFolder()
 
   return {
     ...config,
+
```
frankcalise: ✂️

frankcalise (changes requested):
Thanks for looking into this! I left some feedback - couple of things to clean up and one or two items to dig a little deeper on

frankcalise:
The updates look good! Thanks for doing this.

I'll try and get to pulling down the branch and testing the changes later today.

